PR #5, opened 2019-05-14
config: add TOML inputs

lucab:
This adds basic support for TOML configuration.


dist/examples/00-config-sample.toml
```diff
@@ -0,0 +1,19 @@
+[identity]
```
jlebon: If it's intended to serve as documentation, this sample TOML could probably use some comments for each field.

lucab: I plan to add a dedicated config doc-page once all the options are finalized, but yes more comments don't hurt. Added them.

PR #6, opened 2019-05-15
strategy: add immediate mode

lucab:
This adds update-strategy settings and "immediate" mode.
Immediate is the default update strategy. It fetches and
finalizes new updates immediately, regardless of maintenance
windows or cluster-wide orchestration.
It optionally allows to disable fetching or finalizing via
dedicated config options.

### DIFF
--- a/src/strategy/mod.rs
+++ b/src/strategy/mod.rs
@@ -1,0 +1,78 @@
+//! Update and reboot strategies.
+
+#![allow(unused)]
+
+use crate::config::inputs;
+use crate::identity::Identity;
+use failure::{bail, Fallible};
+use futures::prelude::*;
+use log::error;
+use serde::Serialize;
+
+mod immediate;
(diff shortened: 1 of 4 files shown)


src/strategy/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+use futures::prelude::*;
+use log::error;
+use serde::Serialize;
+
+mod immediate;
+pub(crate) use immediate::StratImmediate;
+
+#[derive(Clone, Debug, Serialize)]
+pub(crate) enum UpStrategy {
+    Immediate(StratImmediate),
```
jlebon: Minor/optional: I'd just spell those out to be clear (i.e. `UpdateStrategy` and `StrategyImmediate`).

lucab: Ack, fixed, thanks!
I'm unfortunately often too terse, so please **do** keep highlighting when things get cryptic.

PR #9, opened 2019-05-16
zincati: add agent state-machine

lucab:
This adds the main state-machine for the update agent.
The agent is modeled as an actix actor which periodically
tries to refresh its state-machine. It progresses through
individual states in order to check, fetch and finalize updates.

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -1,0 +1,111 @@
+//! Update agent actor.
+
+use super::{UpdateAgent, UpdateAgentState};
+use actix::prelude::*;
+use failure::Error;
+use log::trace;
+
+impl Actor for UpdateAgent {
+    type Context = Context<Self>;
+
+    fn started(&mut self, ctx: &mut Self::Context) {
+        trace!("update agent started");
+
+        // Kick-start the state machine.
+        Self::tick_now(ctx);
+    }
+}
+
+pub(crate) struct RefreshTick {}
+
+impl Message for RefreshTick {
+    type Result = Result<(), Error>;
+}
+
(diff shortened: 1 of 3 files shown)


src/update_agent/actor.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    fn handle(&mut self, _msg: RefreshTick, ctx: &mut Self::Context) -> Self::Result {
+        trace!("update agent tick, current state: {:?}", self.state);
+        let prev_state = self.state.clone();
+
+        let state_action = match self.state {
+            UpdateAgentState::StartState => self.initialize(),
+            UpdateAgentState::Initialized => self.try_steady(),
+            UpdateAgentState::Steady => self.try_check_updates(),
+            UpdateAgentState::_EndState => self.finalized(),
```
jlebon: Is the idea down the road that `EndState` happens if one of the state transition hits a failure? Or is it when checking for updates yields an update and it's ready to reboot?

lucab: If a state transition faults, it stays in the current state, logs an error and re-tries on the next tick.

The EndState is reached (not now, but once the whole state-machine is complete) when the agent manages to get a successful finalization from rpm-ostree. At that point, it will sit in the final state, waiting for reboot.

lucab:
I've slightly refactored the state machine to make transitions explicit and checked, and to make it easier to unit-test. I've also added basic unit tests in order to show what would be the normal happy-path of an update cycle. I will later tinker a bit more on how to mock the top-level actor, as its transitions will in general depend on external inputs (cincinnati response, airlock response, rpm-ostreed response).

lucab:
@jlebon mind another look and possibly stamping this?

PR #10, opened 2019-05-17
strategy/immediate: add unit tests

lucab:
This adds unit tests to cover the "immediate" strategy.
Async methods are tested via tokio current-thread runtime.
The full matrix of all configurable cases is covered via
proptest generation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,6 +137,19 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


lucab:
In case you want to read some, `proptest` has some quite lengthy docs at https://altsysrq.github.io/proptest-book/.

/cc @steveej @cgwalters as this may be generally interesting

PR #14, opened 2019-05-20
identity: expose agent parameters

lucab:
This adds some methods on Identity to export agent parameters, in order to be consumed by the Cincinnati client.

### DIFF
--- a/src/identity/mod.rs
+++ b/src/identity/mod.rs
@@ -2,6 +2,7 @@ use crate::config::inputs;
 use failure::{format_err, Fallible, ResultExt};
 use libsystemd::id128;
 use serde::Serialize;
+use std::collections::HashMap;
 
 /// Default group for reboot management.
 static DEFAULT_GROUP: &str = "default";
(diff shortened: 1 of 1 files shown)


src/identity/mod.rs
```diff
@@ -105,3 +133,39 @@ fn compute_node_uuid(app_id: &id128::Id128) -> Fallible<id128::Id128> {
...
+    }
+
+    #[test]
+    fn identity_url_variables() {
+        let id = mock_default(Some(500));
+        let vars = id.url_variables();
+
+        assert!(!vars.contains_key("node_uuid"));
+        assert!(!vars.contains_key("current_version"));
+        assert!(!vars.contains_key("throttle_permille"));
```
jlebon: Minor: maybe also a positive test, e.g. `assert!(vars.contains_key("basearch"))` ?

src/identity/mod.rs
```diff
@@ -74,6 +75,33 @@ impl Identity {
...
+        vars.insert("group".to_string(), self.group.clone());
+        vars.insert("platform".to_string(), self.platform.clone());
+        vars.insert("stream".to_string(), self.stream.clone());
+        vars
+    }
+
+    /// Return Cincinnati client parameters.
+    pub fn cincinnati_params(&self) -> HashMap<String, String> {
+        let mut vars = HashMap::new();
+        vars.insert("basearch".to_string(), self.basearch.clone());
```
jlebon: I guess we could just get the `HashMap` from `self.url_variables()` and enhance that one instead? Good with keeping it separate too though.

src/identity/mod.rs
lucab: Yeah, I have lazy here. I've added all cases now.

src/identity/mod.rs
lucab: I prefer to keep the two split, as they have different purposes and quick inspection is easier this way.

lucab:
@jlebon thanks for the review, I added some more test-logic to address your comment, PTAL.

PR #15, opened 2019-05-20
cincinnati: support fetching update-hints

lucab:
This adds a Cincinnati client, which is used by the agent in order
to fetch update hints and progress its state machine.

### DIFF
--- a/src/cincinnati/client.rs
+++ b/src/cincinnati/client.rs
@@ -1,0 +1,132 @@
+//! Asynchronous Cincinnati client.
+//!
+//! This client implements the [Cincinnati protocol] for update-hints.
+//!
+//! [Cincinnati protocol]: https://github.com/openshift/cincinnati/blob/master/docs/design/cincinnati.md#graph-api
+
+// TODO(lucab): eventually move to its own "cincinnati client library" crate
+
+#![allow(unused)]
+
+use failure::{Error, Fallible, ResultExt};
+use futures::future;
+use futures::prelude::*;
+use reqwest::r#async as asynchro;
+use reqwest::Method;
+use serde::Deserialize;
+use std::collections::HashMap;
+
+/// Cincinnati graph API path endpoint (v1).
+static V1_GRAPH_PATH: &str = "v1/graph";
+
+/// Cincinnati JSON protocol: node object.
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq)]
+pub struct Node {
+    pub version: String,
+    pub payload: String,
+    pub metadata: HashMap<String, String>,
+}
+
+/// Cincinnati JSON protocol: graph object.
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+pub struct Graph {
+    pub nodes: Vec<Node>,
+    pub edges: Vec<(u64, u64)>,
+}
+
+/// Client to make outgoing API requests.
+#[derive(Clone, Debug)]
+pub struct Client {
+    /// Base URL for API endpoint.
+    api_base: reqwest::Url,
+    /// Asynchronous reqwest client.
+    hclient: asynchro::Client,
+    /// Client parameters (query portion).
+    query_params: HashMap<String, String>,
+}
+
+impl Client {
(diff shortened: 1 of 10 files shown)


lucab:
/cc @jlebon for review

This is currently based on https://github.com/coreos/zincati/pull/14 to make review easier, only the last commit is relevant here.

src/cincinnati/client.rs
```diff
@@ -0,0 +1,124 @@
...
+
+use failure::{Error, Fallible, ResultExt};
+use futures::future;
+use futures::prelude::*;
+use reqwest::r#async as asynchro;
+use reqwest::Method;
+use serde::Deserialize;
+use std::collections::HashMap;
+
+/// Cincinnati graph API path endpoint (v1).
```
jlebon: Hmm, is there a link we can add around here to some upstream Cincinnati docs about the API? (Even if just some permalink to code).

src/cincinnati/client.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Asynchronous reqwest client.
+    hclient: asynchro::Client,
+    /// Client parameters (query portion).
+    query_params: HashMap<String, String>,
+}
+
+impl Client {
+    /// Fetch an update-graph from Cincinnati.
+    pub fn fetch_graph(&self) -> impl Future<Item = Graph, Error = Error> {
+        let req = self.new_request(Method::GET, V1_GRAPH_PATH);
```
jlebon: Not exactly the right place to discuss this, though I'm kinda surprised that we have to fetch the whole graph each time. Does the API not support passing the node we're currently on so it only returns a subset of the graph containing the newer node(s) that connect to it?

lucab: The protocol is agnostic to that. We do indeed pass the `current_version` as a client parameter here. The server can trim the graph to only contain "current-version node" plus all updates (current cincinnati backend doesn't). The tradeoff here is between a larger graph to be processed by each client, and graph-filtering to be performed on the fly server-side for each request. We are currently preferring the former, as it's a bit easier to overload the server with the latter, but it's a point that we'll re-evaluate once we scale up (both in number of clients, and in releases in the graph).

jlebon: Ahh gotcha. OK that makes sense.

> We do indeed pass the `current_version` as a client parameter here.

You mean, in the actual HTTP request? Hmm, missing where that's happening here.

lucab:
@jlebon I totally forgot to link it here, the protocol we are implementing is described at https://github.com/openshift/cincinnati/blob/master/docs/design/cincinnati.md#graph-api.

src/cincinnati/client.rs
lucab: Yes, as query parameter in the HTTP request. Logic is here: https://github.com/coreos/zincati/pull/15/files#diff-5f6838301c026c6c12974b331928bbb4R57.

lucab:
Followups from out-of-band discussion: current logic is not properly setting query parameter and content-type; it would be better to mock the call against a mockito stub.

lucab:
@jlebon thanks for catching the missing logic. I've added request parameters, protocol references, and a very basic mockito test.

src/cincinnati/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    for pos in targets {
+        match graph.nodes.get(pos) {
+            Some(n) => updates.push(n.clone()),
+            None => bail!("target node '{}' not present in graph"),
+        };
+    }
+
+    match updates.len() {
+        0 => Ok(None),
+        // TODO(lucab): stable pick next update
```
jlebon: Aside: so, since version strings can be (but usually aren't) pure free form text, probably what we should do eventually here is pick the node with the highest timestamp.

lucab: I'll followup on this separately. I'm not even 100% convinced we need stable-order here, it may depends on the rest of the logic (i.e. if we lock-in into the first update-hint we get). In the worst case, I was just thinking of lexicographic order, but any kind of stable-ordering is correct.

PR #16, opened 2019-05-27
cincinnati: fix media-type header

lucab:
This fixes the media-type header on client request, according to
https://github.com/openshift/cincinnati/blob/master/docs/design/cincinnati.md#request


lucab:
For reference, `accept` belongs to cliend-side request and `content-type` belongs to server-side response. An improper context-switch made my brain mixing the two.

PR #17, opened 2019-05-27
rpm_ostree: add basic client-actor for upgrades

lucab:
This adds an initial actor for rpm-ostree CLI client.
It also sketches the first step in downloading and applying updates.

### DIFF
--- a/src/rpm_ostree/cli_status.rs
+++ b/src/rpm_ostree/cli_status.rs
@@ -1,0 +1,68 @@
+//! Interface to `rpm-ostree status --json`.
+
+#![allow(unused)]
+
+use super::Release;
+use failure::{bail, ensure, format_err, Fallible, ResultExt};
+use serde::Deserialize;
+
+/// JSON output from `rpm-ostree status --json`
+#[derive(Debug, Deserialize)]
+pub struct StatusJSON {
+    deployments: Vec<DeploymentJSON>,
+}
+
+/// Partial deployment object (only fields relevant to zincati).
+#[derive(Debug, Deserialize)]
+pub struct DeploymentJSON {
+    booted: bool,
+    #[serde(rename = "base-checksum")]
+    base_checksum: Option<String>,
+    checksum: String,
+    version: String,
+}
+
+impl DeploymentJSON {
+    /// Convert into `Release`.
+    pub fn into_release(self) -> Release {
+        Release {
+            checksum: self.base_revision(),
+            version: self.version,
+        }
+    }
+
+    /// Return the deployment base revision.
+    pub fn base_revision(&self) -> String {
+        self.base_checksum
+            .clone()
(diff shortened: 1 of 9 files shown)


src/rpm_ostree/cli_upgrade.rs
```diff
@@ -0,0 +1,23 @@
...
+use failure::{bail, format_err, Fallible, ResultExt};
+
+/// Upgrade and leave the new deployment locked.
+pub fn locked_upgrade(release: Release) -> Fallible<Release> {
+    let cmd = std::process::Command::new("rpm-ostree")
+        .arg("upgrade")
+        .arg("--lock-finalization")
+        .arg(release.checksum.to_string())
+        .output()
+        .with_context(|e| format_err!("failed to run rpm-ostree: {}", e))?;
```
lucab: @jlebon my understanding is that this is a blocking command (i.e. its exit code is reliable to signal success). If not, I may need to tweak the state machine a bit.

src/rpm_ostree/cli_upgrade.rs
```diff
@@ -0,0 +1,23 @@
+//! Interface to `rpm-ostree upgrade --lock-finalization`.
+
+use super::Release;
+use failure::{bail, format_err, Fallible, ResultExt};
+
+/// Upgrade and leave the new deployment locked.
+pub fn locked_upgrade(release: Release) -> Fallible<Release> {
+    let cmd = std::process::Command::new("rpm-ostree")
+        .arg("upgrade")
```
jlebon: We just chatted about this OOB already, but just making a note here: this will have to change to `deploy` and the argument you give it would be `revision=$checksum`. Will fix up rpm-ostree.

jlebon: https://github.com/projectatomic/rpm-ostree/pull/1846

src/rpm_ostree/cli_status.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Debug, Deserialize)]
+pub struct StatusJSON {
+    deployments: Vec<DeploymentJSON>,
+}
+
+/// Partial deployment object (only fields relevant to zincati).
+#[derive(Debug, Deserialize)]
+pub struct DeploymentJSON {
+    booted: bool,
+    checksum: String,
```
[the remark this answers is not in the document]
lucab: Great catch, I was vaguely aware of this nuance but I didn't realized it would be relevant here. I'm fine with keeping the current behavior of rpm-ostree.

For reference, this specific codepath is not currently used. But I think future logic will depend on introspection, so I added this preemptively.

lucab:
@jlebon thanks a lot for the insightful review. I've opened a bunch of followup tickets for the things discussed offline, and addressed all code comments.

PR #20, opened 2019-05-28
identity: read platform-id from kernel cmdline

lucab:
This adds support for reading the Ignition platform ID from /proc/cmdline.

### DIFF
--- a/src/identity/platform.rs
+++ b/src/identity/platform.rs
@@ -1,0 +1,97 @@
+//! Kernel cmdline parsing - utility functions
+//!
+//! NOTE(lucab): this is not a complete/correct cmdline parser, as it implements
+//!  just enough logic to extract the platform ID value. In particular, it does not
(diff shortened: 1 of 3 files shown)


src/identity/platform.rs
```diff
@@ -0,0 +1,96 @@
+//! Kernel cmdline parsing - utility functions
+//!
+//! NOTE(lucab): this is not a complete/correct cmdline parser, as it implements
+//!  just enough logic to extract the platform ID value. In particular, it does not
+//!  handle separator quoting/escaping, list of values, and merging of repeated
+//!  flags.
```
jlebon: Probably worth mentioning that this code comes from Afterburn? Just so that any bugs found here should probably also be applied there.

lucab: Ack, I've added a note and direct link to Afterburn source.

PR #23, opened 2019-05-29
update_agent: add update finalization

lucab:
This adds support for finalizing an OS update.

### DIFF
--- a/src/rpm_ostree/cli_finalize.rs
+++ b/src/rpm_ostree/cli_finalize.rs
@@ -1,0 +1,22 @@
+//! Interface to `rpm-ostree finalize-deployment`.
+
+use super::Release;
+use failure::{bail, format_err, Fallible, ResultExt};
+
+/// Unlock and finalize the new deployment.
+pub fn finalize_deployment(release: Release) -> Fallible<Release> {
+    let cmd = std::process::Command::new("rpm-ostree")
+        .arg("finalize-deployment")
+        .arg(&release.checksum)
+        .output()
+        .with_context(|e| format_err!("failed to run rpm-ostree: {}", e))?;
+
+    if !cmd.status.success() {
+        bail!(
+            "rpm-ostree finalize-deployment failed:\n{}",
+            String::from_utf8_lossy(&cmd.stderr)
+        );
+    }
+
+    Ok(release)
+}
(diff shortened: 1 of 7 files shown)


src/rpm_ostree/cli_finalize.rs
```diff
@@ -0,0 +1,22 @@
+//! Interface to `rpm-ostree finalize-deployment`.
+
+use super::Release;
+use failure::{bail, format_err, Fallible, ResultExt};
+
+/// Unlock and finalize the new deployment.
+pub fn finalize_deployment(release: Release) -> Fallible<Release> {
+    let cmd = std::process::Command::new("rpm-ostree")
+        .arg("finalize-deployment")
+        .arg(release.reference_id())
```
jlebon: Though here, it expects the checksum, not a `revision=$checksum`. Offhand, I'd just drop the `reference_id()` thing and do the `format!()` directly in `locked_upgrade()`. Really, that special syntax is available just for `deploy` and `rebase`.

src/update_agent/actor.rs
```diff
@@ -148,15 +175,30 @@ impl UpdateAgent {
...
+        if !can_finalize {
+            return Box::new(actix::fut::err(()));
+        }
+
+        let msg = rpm_ostree::FinalizeDeployment { release };
+        let upgrade = self
+            .rpm_ostree_actor
+            .send(msg)
+            .flatten()
+            .map_err(|e| log::error!("failed to finalize update: {}", e))
```
jlebon: So let's say it fails (either because of [the reboot race](https://github.com/projectatomic/rpm-ostree/issues/1348), in which case it doesn't matter much, OR because it actually couldn't finalize), will the state machine stay in the `UpdateStaged` state and retry on the next tick?

src/rpm_ostree/cli_finalize.rs
lucab: Good catch. I guess I was assuming that the input parse was uniform across verbs. I'll undo the helper as you suggested.

src/update_agent/actor.rs
lucab: Yes, as we would still be returning an error (of type `()`). This is just doing error-type mapping, in order to log error messages without the caller caring about the details. The future chained after this (via a `map()`) will only run on success.

jlebon: Gotcha, thanks! Didn't mean to call out this specific line. Just confirming my understanding at a high level.

lucab:
@jlebon comments addressed, PTAL.

Issue #25, opened 2019-05-29
cincinnati: use base_checksum as lookup key


lucab:
Zincati currently assumes that OS version label is unique and that such label can be used as a lookup key in order to find the current OS node in the update-graph.
While this is likely true in most cases, it may be a bit fragile (across all the FCOS streams) and there is no real reason to do that. Instead, we can directly use the `base_checksum` for lookups in the graph received from the Cincinnati backend.

PR #28, opened 2019-05-31
cincinnati: lookup by OS checksum

lucab:
This switches the graph-lookup logic handling lookup in Cincinnati,
in order to use OS (base) checksum instead of version labels.

Closes: https://github.com/coreos/zincati/issues/25


lucab:
There was a minor conflict, fixed without semantic changes. Merging.

PR #29, opened 2019-06-05
dist: add systemd service unit

lucab:


dist/systemd/system/zincati.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Zincati update agent
```
jlebon: This is minor, though I think the convention leans a bit more on the description being Title Case?

jlebon: Maybe let's also add for now `Documentation=https://github.com/coreos/zincati` ?

dist/systemd/system/zincati.service
```diff
@@ -0,0 +1,15 @@
...
+Requires=rpm-ostreed.service
+After=rpm-ostreed.service
```
jlebon: I don't think we need these. The daemon isn't long-running; it auto-exits when there are no more clients connected.

lucab:
Amended.

Issue #18, opened 2019-05-27
cincinnati: define "payload" semantics


lucab:
Following up from @jlebon comment at https://github.com/coreos/zincati/pull/17#discussion_r287808035.

Right now we plan to use the sha-checksum as "payload" field in cincinnati graph. However, rpm-ostree can understand multiple kinds of identifiers (current: `revision` and `version`), so it would be nice to have some kind of explicit scheme. This would help in removing update-graph ambiguities and in future-proofing it.

Alternative approaches are:
 * require a mandatory type-prefix followed by a `=`, e.g. `revision=...`
 * add a key-value pair in metadata table to specify the scheme/prefix

lucab:
@bgilbert @arithx this is partially related to https://github.com/coreos/fedora-coreos-tracker/issues/98#issuecomment-497521551, so your feedback would be welcome.

The goal would be to have some room for future-proofing, in case we need to change what the client consider a "payload identifier". The feedback gathering is about 1) whether it's worth doing it 2) how that should be encoded in the Cincinnati graph.

jlebon:
> However, rpm-ostree can understand multiple kinds of identifiers

One note, this is just an `rpm-ostree deploy` and `rpm-ostree rebase` UX syntax thing. It's not really a deeper concept than that (e.g. we don't write down things in that syntax anywhere). In general the checksum is really the fundamental thing we care about.

That said, the idea of a scheme in the cincinnati data makes sense to me regardless. My strawman is just keeping the payload as the checksum, *but* in the metadata having:
- a `scheme` key set to e.g. "checksum"
- an `ostree-checksum` key
- an `ostree-version` key

?

The `ostree-` ones are intended to be stable keys from which you can always get that info (and having the version there is important for visualizing the graph using versions if possible).

bgilbert:
I don't think the type prefix would actually change much.  If we need to add a prefix later, we can do it without ambiguity; and doing so will break old clients whether we have a prefix or not.  But if folks would prefer a prefix, there's little harm to it.

I think the `ostree-*` metadata items make sense.  `scheme` is subject to the same argument as the above, though, so I'd favor leaving it out.

jlebon:
[...]

PR #32, opened 2019-06-11
cincinnati: define payload scheme

lucab:
This defines a well-known metadata key to detect payload scheme,
and reserve the initial "checksum" kind.

Closes: https://github.com/coreos/zincati/issues/18

### DIFF
--- a/src/rpm_ostree/mod.rs
+++ b/src/rpm_ostree/mod.rs
@@ -6,7 +6,8 @@ pub use cli_status::booted;
 mod actor;
 pub use actor::{FinalizeDeployment, RpmOstreeClient, StageDeployment};
 
-use crate::cincinnati::Node;
+use crate::cincinnati::{Node, CHECKSUM_SCHEME, SCHEME_KEY};
+use failure::{ensure, format_err, Fallible};
 use serde::Serialize;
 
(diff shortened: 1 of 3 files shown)


src/rpm_ostree/mod.rs
```diff
@@ -20,10 +21,18 @@ pub struct Release {
...
     /// Builds a `Release` object from a Cincinnati node.
-    pub fn from_cincinnati(node: Node) -> Self {
-        Self {
+    pub fn from_cincinnati(node: Node) -> Fallible<Self> {
+        let payload_is_checksum = node
+            .metadata
+            .get(SCHEME_KEY)
+            .map(|v| v == CHECKSUM_SCHEME)
+            .unwrap_or(false);
+        ensure!(payload_is_checksum, "unexpected non-checksum payload");
```
jlebon: Should we include the actual scheme we got instead here?

lucab: We can. There is the further case where there is no scheme at all, though.

jlebon: Consider it optional. Not blocking on this.

lucab:
Augmented with better error messages.

PR #33, opened 2019-06-12
strategy/immediate: remove config options

lucab:
This removes existing config options for `strategy.immediate`. It
was deemed that these options don't have a strong-enough usecase
at this point, and it is easy to add them later if proper
requirements arise.

### DIFF
--- a/src/strategy/immediate.rs
+++ b/src/strategy/immediate.rs
@@ -1,8 +1,5 @@
 //! Strategy for immediate updates.
 
-#![allow(unused)]
-
-use crate::config::inputs::StratImmediateInput;
 use failure::{Error, Fallible};
 use futures::future;
 use futures::prelude::*;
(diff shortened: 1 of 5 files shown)


tests/fixtures/00-config-sample.toml
```diff
@@ -12,13 +12,7 @@ group = "workers"
 [cincinnati]
 base_url= "http://example.com:80/"
 
-[updates]
 # Valid strategies: immediate / periodic / remote_http
```
jlebon: Aside: what's the difference between `immediate` and `periodic` again?

jlebon: Heh OK right, the next unread notification answers this question: https://github.com/coreos/zincati/issues/34 :)

src/strategy/immediate.rs
```diff
@@ -19,20 +16,6 @@ pub(crate) struct StrategyImmediate {
 }
```
jlebon: Do we still need the `check` and `finalize` booleans in the struct?

lucab: Probably not, unless somebody comes and ask back for this later on in the future. I was mostly concerned about not exposing the configuration knobs. Do you mind if I leave these for the time being, and revisit after a few releases?

jlebon: Yup, WFM!

lucab: Tracked a reminder ticket at https://github.com/coreos/zincati/issues/35.

PR #36, opened 2019-06-13
cargo: add metadata

lucab:
This adds Cargo manifest metadata, in preparation for the first release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,11 @@
 [package]
 name = "zincati"
 version = "0.0.0-dev.0"
+description = "Update agent for Fedora CoreOS"
+license = "Apache-2.0"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -29,3 +33,12 @@ tokio = "^0.1"
...
 [profile.release]
 lto = true
+
+[package.metadata.release]
+sign-commit = true
+upload-doc = false
+disable-publish = true
+disable-push = true
+pre-release-commit-message = "cargo: zincati release {{version}}"
+pro-release-commit-message = "cargo: development version bump"
```
jlebon: Aside: I'm confused why this is `pro` instead of `post` if they were going for the opposite of `pre`... :smile:

lucab: On spot. I do believe the author is a native speaker of a non-Latin-rooted language. That field has always been called that way.

PR #43, opened 2019-06-24
config: use liboverdrop scanner 

lucab:
This updates configuration logic, dropping our custom file scanner and using [liboverdrop](https://docs.rs/liboverdrop/0.0.1/liboverdrop/struct.FragmentScanner.html#method.scan) instead.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -34,8 +34,14 @@ pub(crate) struct Settings {
 impl Settings {
     /// Assemble runtime settings.
     pub(crate) fn assemble() -> Fallible<Self> {
-        let dirs = vec!["/usr/lib", "/run", "/etc"];
-        let cfg = inputs::ConfigInput::read_configs(&dirs, crate_name!())?;
+        let prefixes = vec![
+            "/usr/lib/".to_string(),
+            "/run/".to_string(),
+            "/etc/".to_string(),
+        ];
+        let common_path = format!("{}/config.d/", crate_name!());
+        let extensions = vec!["toml".to_string()];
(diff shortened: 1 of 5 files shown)


src/config/mod.rs
```diff
@@ -34,8 +34,14 @@ pub(crate) struct Settings {
...
-        let dirs = vec!["/usr/lib", "/run", "/etc"];
-        let cfg = inputs::ConfigInput::read_configs(&dirs, crate_name!())?;
+        let prefixes = vec![
+            "/usr/lib/".to_string(),
+            "/run/".to_string(),
+            "/etc/".to_string(),
+        ];
+        let common_path = format!("{}/config.d/", crate_name!());
+        let extensions = vec!["toml".to_string()];
+        let cfg = inputs::ConfigInput::read_configs(prefixes, &common_path, extensions)?;
```
rfairley: Ahh - having to later borrow the string returned by `format!` makes me wonder if `liboverdrop::FragmentScanner::new()` should accept an owned `String` for the `shared_path` parameter instead of `&str`. This would be consistent with the `prefixes` and `extensions` holding `String` as well.

Having `shared_path: &str` is easier for directly passing in a slice e.g. `"my-app/config.d"` but the more common pattern would probably be using `format!` and passing in the crate name.

rfairley: Opened an issue for this: https://github.com/overdrop/liboverdrop-rs/issues/7

lucab:
@jlebon unless objections, I'd like to get this merged then proceed to tag version 0.0.2.

Issue #34, opened 2019-06-12
strategy: add weekly maintenance windows mode, configuration and logic


lucab:
Zincati needs to have an additional update strategy, called "periodic". This is used to schedule update finalization (i.e. host reboot) locally on each node, based on current datetime and configuration.

This is used to configure what's normally called a "maintenance window" for OS reboot. Prior art for this is https://github.com/coreos/locksmith#reboot-windows.

A `StrategyPeriodic` will be in charge of gating the "finalize-deployment" call, only allowing it when local time is within the configured window.

Configuration options to add:
 * a `updates.strategy = periodic` variant
 * a `strategy.periodic` section
 * a list of finalization `[[window]]`, each one containing:
    - a list of short-day-names (like `date %a` in C local, i.e. `[ "Mon", "Wed" ]`)
    - a window start time, in 24h format (i.e. `23:45`)
    - window length, in minutes
    - ~~UTC (bool, default: `true`). Still unsure about this, see https://github.com/coreos/airlock/issues/13#issuecomment-501349851.~~

Note: this relies on local node clock and time configuration. Fedora CoreOS docs should recommend running an NTP client, and setting host timezone to UTC.

Limero:
This is the missing piece for me migrating my home server to Fedora CoreOS. I can't have it randomly restart while I use it, so scheduling updates during the night would be perfect.

lucab:
Made some progress on this, I've currently wired up the following configuration in #47:

```toml
[updates]
strategy = "weekly"

[[updates.weekly.window]]
days = [ "Sat", "Sun" ]
start_time = "23:30"
length_minutes = 60

[[updates.weekly.window]]
days = [ "Wed" ]
start_time = "01:00"
length_minutes = 30
```

Punting on the "local timezone" setting for the moment, as Daylight Saving Time (DST) switches and timezone-law changes have many nasty side-effects on maintenance windows.

PR #47, opened 2019-06-26
strategy: add logic for weekly maintenance windows

lucab:
This implements a new update strategy, with support for weekly maintenance windows.

Closes: https://github.com/coreos/zincati/issues/34

### DIFF
--- a/docs/usage/updates-strategy.md
+++ b/docs/usage/updates-strategy.md
@@ -56,3 +56,40 @@ Such an approach is only recommended where nodes are already grouped into an orc
 
 [fleet_lock]: https://github.com/coreos/airlock/pull/1 
 [etcd3]: https://etcd.io/
+
+# Periodic strategy
+
+The `periodic` strategy allows Zincati to only reboot for updates during certain timeframes, also known as "maintenance windows" or "reboot windows".
+Outside of those maintenance windows, reboots are not automatically performed and auto-updates are staged and held until the next available window.
+
+Reboot windows recur on a weekly basis, and can be defined in any arbitrary order and length. Their individual length must be greater than zero.
+To avoid timezone-related skews in a fleet of machines, all maintenance windows are defined in UTC dates and times.
+
+Periodic reboot windows can be configured and enabled in the following way:
+
+```toml
+[updates]
+strategy = "periodic"
+
+[[updates.periodic.window]]
+days = [ "Sat", "Sun" ]
+start_time = "23:30"
+length_minutes = 60
+
+[[updates.periodic.window]]
+days = [ "Wed" ]
+start_time = "01:00"
+length_minutes = 30
+```
+
+The above configuration would result in three maintenance windows during which Zincati is allowed to reboot the machine for updates:
+ * 60 minutes starting at 23:30 UTC on Saturday night, and ending at 00:30 UTC on Sunday morning
+ * 60 minutes starting at 23:30 UTC on Sunday night, and ending at 00:30 UTC on Monday morning
+ * 30 minutes starting at 01:00 UTC on Wednesday morning, and ending at 01:30 UTC on Wednesday morning
+
+Reboot windows can be separately configured in multiple snippets, as long as each `updates.periodic.window` entry contains all the required properties:
+ * `days`: an array of weekdays (C locale), either in full or abbreviated (first three letters) form
+ * `start_time`: window starting time, in `hh:mm` ISO 8601 format
+ * `length_minutes`: non-zero window duration, in minutes
+
+For convenience, multiple entries can be defined with overlapping times, and each window definition is allowed to cross day and week boundaries (wrapping to the next day).
--- a/src/config/inputs.rs
+++ b/src/config/inputs.rs
@@ -48,7 +48,7 @@ impl ConfigInput {
     }
 
     /// Merge multiple fragments into a single configuration.
-    fn merge_fragments(fragments: Vec<fragments::ConfigFragment>) -> Self {
+    pub(crate) fn merge_fragments(fragments: Vec<fragments::ConfigFragment>) -> Self {
         let mut agents = vec![];
         let mut cincinnatis = vec![];
         let mut updates = vec![];
@@ -167,6 +167,8 @@ pub(crate) struct UpdateInput {
     pub(crate) strategy: String,
     /// `fleet_lock` strategy config.
     pub(crate) fleet_lock: FleetLockInput,
+    /// `periodic` strategy config.
+    pub(crate) periodic: PeriodicInput,
 }
 
 /// Config for "fleet_lock" strategy.
@@ -174,6 +176,21 @@ pub(crate) struct UpdateInput {
 pub(crate) struct FleetLockInput {
     /// Base URL (template) for the FleetLock service.
     pub(crate) base_url: String,
+}
+
+/// Config for "periodic" strategy.
+#[derive(Clone, Debug, Serialize)]
+pub(crate) struct PeriodicInput {
+    /// Set of updates windows.
+    pub(crate) intervals: Vec<PeriodicIntervalInput>,
+}
+
+/// Update window for a "periodic" interval.
+#[derive(Clone, Debug, Serialize)]
+pub(crate) struct PeriodicIntervalInput {
+    pub(crate) start_day: String,
+    pub(crate) start_time: String,
+    pub(crate) length_minutes: u32,
 }
 
 impl UpdateInput {
@@ -184,6 +201,7 @@ impl UpdateInput {
         let mut fleet_lock = FleetLockInput {
             base_url: String::new(),
         };
+        let mut periodic = PeriodicInput { intervals: vec![] };
 
         for snip in fragments {
             if let Some(a) = snip.allow_downgrade {
@@ -200,6 +218,20 @@ impl UpdateInput {
                     fleet_lock.base_url = b;
                 }
             }
+            if let Some(w) = snip.periodic {
+                if let Some(win) = w.window {
+                    for entry in win {
+                        for day in entry.days {
+                            let interval = PeriodicIntervalInput {
+                                start_day: day,
+                                start_time: entry.start_time.clone(),
+                                length_minutes: entry.length_minutes,
+                            };
+                            periodic.intervals.push(interval);
+                        }
+                    }
+                }
+            }
         }
 
         Self {
@@ -207,6 +239,7 @@ impl UpdateInput {
             enabled,
             strategy,
             fleet_lock,
+            periodic,
         }
     }
 }
--- a/src/weekly/mod.rs
+++ b/src/weekly/mod.rs
@@ -1,0 +1,411 @@
+//! Calendar-windows for events recurring on weekly basis.
+//!
+//! This contains helper logic to handle intervals of time which recur every week:
+//!  * `WeeklyWindow`: a continuous interval over a single week.
+//!  * `WeeklyCalendar`: a set of intervals on a weekly calendar.
+
+// TODO(lucab): stabilize and split this to its own `weekly` crate.
+
+pub(crate) mod utils;
+
+use chrono::{DateTime, Utc};
+use failure::Fallible;
+use intervaltree::{Element, IntervalTree};
+use serde::{Serialize, Serializer};
+use std::cmp::Ordering;
+use std::iter::FromIterator;
+use std::ops::Range;
+use std::time::Duration;
+
+/// Whole week duration, in minutes.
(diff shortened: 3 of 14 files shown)


lucab:
Status update: rebased to master, lower level logic for weekly intervals should be complete now.

lucab:
@jlebon this should be ready for an initial review pass, possibly starting top-down from the usage docs. I still plan to perform a couple more manual tests to make sure I didn't miss some corner cases.

src/config/inputs.rs
```diff
@@ -176,6 +178,21 @@ pub(crate) struct FleetLockInput {
     pub(crate) base_url: String,
 }
 
+/// Config for "weekly" strategy.
```
jlebon, line 181: Sidenote: I find the naming for this strategy a little confusing. "weekly" makes it seem like all it can do is do updates every week. When really, it also implements daily or every X hours, etc, right? I don't really have a better suggestion though... maybe "windowed"?

src/strategy/weekly.rs
```diff
@@ -0,0 +1,122 @@
...
+            let start = utils::time_from_string(&entry.start_time)?;
+            let length = Duration::from_secs(u64::from(entry.length_minutes).saturating_mul(60));
+            let windows = DailyWindow::parse_timespan(weekday, start.0, start.1, length)?;
+            intervals.extend(windows);
+        }
+
+        let calendar = WeeklyCalendar::new(intervals);
+        match calendar.total_length_minutes() {
+            0 => failure::bail!("zero-length weekly updates window"),
+            n => log::trace!("weekly updates, total length: {} minutes", n),
```
jlebon, line 36: Hmm, wouldn't it be more useful to report the non-overlapping total number of minutes here?

docs/usage/updates-strategy.md
```diff
@@ -56,3 +56,40 @@ Such an approach is only recommended where nodes are already grouped into an orc
...
+start_time = "01:00"
+length_minutes = 30
+```
+
+The above configuration would result in three maintenance windows during which Zincati is allowed to reboot the machine for updates:
+ * 60 minutes starting at 23:30 UTC on Saturday night, and ending at 00:30 UTC on Sunday morning
+ * 60 minutes starting at 23:30 UTC on Sunday night, and ending at 00:30 UTC on Monday morning
+ * 30 minutes starting at 01:00 UTC on Wednesday morning, and ending at 01:30 UTC on Wednesday morning
+
+Reboot windows can be separetely configured in multiple snippets, as long as each `updates.weekly.window` entry contains all the required properties:
```
jlebon, line 90: separately

src/strategy/weekly.rs
lucab: Absolutely yes, but I'd need to implement that in the underlying library first (I have it on my radar). This is a trace-log not seen by default and mostly meant for debugging.
Once we have the correct number, it should be better exposed via metrics or such.

src/config/inputs.rs
lucab: I do not disagree :)

In the first iteration here I used "periodic", but I realized it was too generic (and I'm still considering if I can have another strategy based on https://www.freedesktop.org/software/systemd/man/systemd.time.html# later on).

Locksmith used to just call this "reboot_window" (but it allowed a single one).

If we can find a single word which conveys both the "timespans on a calendar" and "recurring every week" I'd be very happy.

src/weekly/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+impl DailyWindow {
+    /// Parse a timespan into daily windows.
+    pub fn parse_timespan(
+        start_day: chrono::Weekday,
+        start_hour: u8,
+        start_minute: u8,
+        length: Duration,
+    ) -> Fallible<Vec<Self>> {
```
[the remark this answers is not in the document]
lucab: That was in my initial implementation, but it was a bit too naïf thus failing on corner-cases.

In particular, a weekly calendar is a [finite field of order 7](https://en.wikipedia.org/wiki/Finite_field). Perhaps not very intuitive, but all the logic here has to obey [modular arithmetic](https://en.wikipedia.org/wiki/Modular_arithmetic) in that field. (same goes for hours and minutes too, really)

In other words, it has to account for the fact that a timespan can start on Sunday night and end on Monday morning, without jumping back in time :)

Splitting into multiple daily windows here makes the rest of the intervals logic easier to deal with, as it can simply ignore all modular details and just use integer arithmetic when comparing `MinuteInWeek`.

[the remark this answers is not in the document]
jlebon: Right, that makes sense. I guess we could split just the intervals which cross week boundaries into two. But meh... I see the appeal in being consistent.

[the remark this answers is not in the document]
lucab: Uh, good point, I somehow assumed it was easier to handle interval windows on daily boundaries but that's not a given. I'll try to quickly sketch your alternative next and check if it is easier to follow.

src/strategy/weekly.rs
lucab: For reference, https://github.com/main--/rust-intervaltree/issues/7 is in 0.2.6. I've left a TODO to compute the proper length measure here later on.

src/weekly/mod.rs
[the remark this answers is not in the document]
lucab: I reworked the logic here to use week boundaries, and the final result is indeed easier to read. I've stick to that and pushed this new version, thanks for the suggestion!

lucab:
@jlebon rebased with your suggestion, PTAL. I'm leaving the "compute length with overlaps" as a TODO as I wanted to re-check the algorithm for that first.

src/config/inputs.rs
arithx: `interval`?

dustymabe: what was wrong with `periodic`? One way to handle adding another one later on is to use a version maybe

lucab: @dustymabe nothing per se, it was just feeling more generic/broad than `weekly`. But if the latter is deceiving and the former seems ok, I can revert back.

(no need to worry for future strategy names here, I just wanted to note that this may not be the only time-based one)

docs/usage/updates-strategy.md
```diff
@@ -56,3 +56,40 @@ Such an approach is only recommended where nodes are already grouped into an orc
...
+length_minutes = 30
+```
+
+The above configuration would result in three maintenance windows during which Zincati is allowed to reboot the machine for updates:
+ * 60 minutes starting at 23:30 UTC on Saturday night, and ending at 00:30 UTC on Sunday morning
+ * 60 minutes starting at 23:30 UTC on Sunday night, and ending at 00:30 UTC on Monday morning
+ * 30 minutes starting at 01:00 UTC on Wednesday morning, and ending at 01:30 UTC on Wednesday morning
+
+Reboot windows can be separately configured in multiple snippets, as long as each `updates.weekly.window` entry contains all the required properties:
+ * `days`: an array of weekdays (C locale), either in full or abbreviated form
```
zonggen, line 91: super minor: could mention abbreviated = first three letters of full form

src/config/inputs.rs
jlebon: Will throw another entry in the hat: `weekly_schedule`. It's more wordy, but it makes it more clear that it's about setting up a schedule tied to a week rather than a single window which happens every week.

src/strategy/weekly.rs
```diff
@@ -0,0 +1,122 @@
+//! Strategy for weekly weekly updates.
```
jlebon, line 1: repeated word

src/weekly/utils.rs
```diff
@@ -0,0 +1,181 @@
...
+        .saturating_add(u32::from(minute.min(59)));
+
+    assert!(weekly_minute < MAX_WEEKLY_MINS);
+    weekly_minute
+}
+
+/// Check duration for a sane lower and upper bound (whole week).
+pub(crate) fn check_duration(length: Duration) -> Fallible<Duration> {
+    // Whole week duration, in seconds.
+    const MAX_WEEKLY_SECS: u64 = MAX_WEEKLY_MINS as u64 * 60;
```
jlebon, line 39: This is already defined in the `weekly` module.

src/weekly/utils.rs
```diff
@@ -0,0 +1,181 @@
...
+        "fri" | "friday" => Weekday::Fri,
+        "sat" | "saturday" => Weekday::Sat,
+        "sun" | "sunday" => Weekday::Sun,
+        _ => bail!("unrecognized week day: {}", input),
+    };
+
+    Ok(day)
+}
+
+/// Parse a time string (in 24h format).
```
jlebon, line 67: Would help to have a sample time string here.

src/strategy/weekly.rs
```diff
@@ -0,0 +1,122 @@
...
+use log::trace;
+use serde::Serialize;
+use std::pin::Pin;
+use std::time::Duration;
+
+/// Strategy for weekly weekly updates.
+#[derive(Clone, Debug, Default, Serialize)]
+pub(crate) struct StrategyWeekly {
+    /// Whitelisted time windows during which updates are allowed.
+    allowed_intervals: WeeklyCalendar,
```
jlebon, line 17: How about calling this e.g. `schedule`? It feels funny to name it something that refers to the underlying implementation when the struct name is meant to abstract it.

src/weekly/mod.rs
```diff
@@ -0,0 +1,411 @@
...
+    pub fn containing_windows(&self, datetime: &DateTime<Utc>) -> Vec<&WeeklyWindow> {
+        let timepoint = utils::datetime_as_weekly_minute(datetime);
+        self.windows
+            .query_point(timepoint)
+            .map(|elem| &elem.value)
+            .collect()
+    }
+}
+
+impl Serialize for WeeklyCalendar {
```
jlebon, line 106: Oh interesting. Why do we need a serialization support for this?

src/weekly/mod.rs
```diff
@@ -0,0 +1,411 @@
...
+    ) -> Fallible<Vec<Self>> {
+        failure::ensure!(
+            start_hour <= 24 && start_minute <= 59,
+            "invalid start time: {}:{}",
+            start_hour,
+            start_minute
+        );
+
+        let mut windows = vec![];
+        let mut remaining_length = utils::check_duration(length)?;
```
jlebon, line 155: Hmm, this makes it look like it's transforming its input. I'd personally just make the input parameter a `mut` and just call `utils::check_duration(length)?;` on its own.

src/weekly/mod.rs
```diff
@@ -0,0 +1,411 @@
...
+        let mut start_weekly_minute =
+            utils::time_as_weekly_minute(start_day, start_hour, start_minute);
+        while remaining_length.as_secs() > 0 {
```
jlebon, line 158: This is an interesting construction. Really this can iterate at most twice, right? It might be clearer to just change `chop_timestan` to return a Vec of either one or two windows.

lucab: Good catch! Both the while-loop and most of `chop_timespan` logic are not actually useful anymore after moving away from daily windows. I moved some code around and dropped both.

src/weekly/mod.rs
lucab: Fair point, it's a bit deceiving. I changed the helper signature to immutably borrow the argument, as it doesn't really need ownership there.

src/weekly/mod.rs
lucab: We do embed this in the final runtime configuration, whose top-level container has a `Serialize` annotation.

I don't think it's used right now. I used to have some private logic for runtime config introspection, but I didn't turn into a public API as I wasn't unsure about the details. I can try resurrecting that if it seems useful.
Keeping the datastructure serializable is a low incremental effort and unlocks such future usage.

src/weekly/utils.rs
lucab: Compiler checked examples in docstrings FTW!

src/config/inputs.rs
lucab: Ok, it looks like we have full circle without a clear winner. It seems that much confusion came from my late decision to step away from the initial pick. I have reverted everything back to use `periodic` as in the original form. How does it look like?

lucab:
Rebased again, with a revert to the original `periodic` name, PTAL (and thanks all for the feedback so far).

lucab:
Thanks everybody for the reviews/feedbacks! I'm holding this for a last a round of time-based manual testing, will merge afterwards unless anything is found to be majestically broken.

lucab:
Manually tested on top of a `31.20200420.2.0` -> `31.20200505.2.0` update on `testing`, all good. Merging.

PR #48, opened 2019-07-01
update-agent: observe state changes

lucab:
This adds some instrumentation to update-agent state machine, so that
its internal state and last transition can be observed externally.

### DIFF
--- a/src/update_agent/mod.rs
+++ b/src/update_agent/mod.rs
@@ -9,10 +9,18 @@ use crate::rpm_ostree::{Release, RpmOstreeClient};
 use crate::strategy::UpdateStrategy;
 use actix::Addr;
 use chrono::prelude::*;
+use prometheus::IntGauge;
 use std::time::Duration;
 
 /// Default tick/refresh period for the state machine (in seconds).
 const DEFAULT_REFRESH_PERIOD_SECS: u64 = 5 * 60;
+
+lazy_static::lazy_static! {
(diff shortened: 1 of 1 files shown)


src/update_agent/mod.rs
```diff
@@ -35,17 +51,45 @@ enum UpdateAgentState {
...
+
+        match self {
+            UpdateAgentState::StartState => 0,
+            UpdateAgentState::Initialized => 1,
+            UpdateAgentState::Steady => 2,
+            UpdateAgentState::UpdateAvailable(_) => 3,
+            UpdateAgentState::UpdateStaged(_) => 4,
+            UpdateAgentState::UpdateFinalized(_) => 5,
+            UpdateAgentState::EndState => 6,
+        }
```
jlebon: Hmm, does reporting these back to Prometheus imply we're stabilizing the various states we have? Or are we free to change these later on?

lucab: Good question. The socket exposing this is still under `/private` so it's fine for now, but yes it will become part of the public interface and somehow stable. Maybe I should avoid exposing this.

jlebon: Probably not, though could we report strings instead? E.g. "initialized", "staged", "finalized", etc...

lucab: We cannot **as values** (only f64 allowed). We can as part of **metrics name**, but that would crate a separate metric for each state, and monitoring folks strongly recommended against that. I'll just drop the state-id for now, I'll re-add later when the core logic is more stable.

Issue #54, opened 2019-07-03
identity: drop harcoded basearch


lucab:
Zincati needs to introspect host architecture at runtime, in order to request the relevant update-graph to the Cincinnati backend. In particular, it needs to detect `basearch` (from RPM vocabulary) in order to match rpm-ostree expectations. Such architecture label is in several cases not aligned with the kernel one (i.e. `uname -m` or similar), thus we can't simply ask the kernel and we need to rely on the userland to expose this information.

We are currently using an hardcoded translation table at https://github.com/coreos/zincati/blob/025453b4637052281feb50ed00b7ce66cbe31b2f/src/identity/basearch.rs#L6

We should replace it with proper `basearch` sourcing. This is currently blocked on https://github.com/coreos/fedora-coreos-tracker/issues/154.

jlebon:
Here's an example `rpm-ostree status --json` output with the `basearch` injected as per https://github.com/coreos/coreos-assembler/pull/590:

```json
{
  "deployments" : [
    {
      "unlocked" : "none",
      "requested-local-packages" : [
      ],
      "base-commit-meta" : {
        "coreos-assembler.config-dirty" : "true",
        "coreos-assembler.config-gitrev" : "e5ffda727170e124c7f7d8782e256b0462bf9869",
        "coreos-assembler.basearch" : "x86_64",
        "rpmostree.inputhash" : "0e2ba98bac847c8a8cc5ffd42822a41d53ef6719ccaf6a0094559768aa24da33",
        "version" : "30.1",
        "rpmostree.rpmmd-repos" : [
          {
            "id" : "fedora-coreos-pool",
            "timestamp" : 7435743550895030272
          },
          {
            "id" : "fedora",
            "timestamp" : -7689964138319052800
          },
          {
            "id" : "fedora-updates",
            "timestamp" : -1292510703390818304
          },
          {
            "id" : "coreos-assembler-local-overrides",
            "timestamp" : 1402332922660257792
          }
        ]
      },
      "base-removals" : [
      ],
      "gpg-enabled" : false,
      "origin" : "fedora/x86_64/coreos/testing-devel",
      "osname" : "fedora-coreos",
      "pinned" : false,
      "requested-base-local-replacements" : [
      ],
      "checksum" : "ce718b90ce19c2cac07fa45efdce91aad208a508c98e9f7f572f5aaba76f569a",
      "regenerate-initramfs" : false,
[...]

PR #58, opened 2019-07-04
identity: introspect basearch

lucab:
This introspects node basearch from rpm-ostree, instead of using
build-time configuration and the internal mapping table.

Ref: https://github.com/coreos/coreos-assembler/pull/590
Closes: https://github.com/coreos/zincati/issues/54


src/rpm_ostree/cli_status.rs
```diff
@@ -18,6 +18,8 @@ pub struct DeploymentJSON {
     booted: bool,
     #[serde(rename = "base-checksum")]
     base_checksum: Option<String>,
+    #[serde(rename = "coreos-assembler.basearch")]
```
jlebon: Check out the JSON at https://github.com/coreos/zincati/issues/54#issuecomment-508087569. This is actually nested under `base-commit-meta`.

lucab:
Ouch, I totally missed one level of nesting. To compensate for that, I've know added fixtures and mock to cover the deserialization part, PTAL

tests/fixtures/rpm-ostree-status.json
```diff
@@ -0,0 +1,60 @@
...
+        "rpmostree.inputhash" : "0e2ba98bac847c8a8cc5ffd42822a41d53ef6719ccaf6a0094559768aa24da33",
+        "version" : "30.1",
+        "rpmostree.rpmmd-repos" : [
+          {
+            "id" : "fedora-coreos-pool",
+            "timestamp" : 7435743550895030272
+          },
+          {
+            "id" : "fedora",
+            "timestamp" : -7689964138319052800
```
lucab: @jlebon not strictly related to this PR, but I fear there something going wrong in the compose if this timestamp ends up negative.

jlebon: Yeah, that's tracked in https://github.com/projectatomic/rpm-ostree/issues/1734. Need to look at that again soon.

PR #59, opened 2019-07-04
dist: add polkit rule for rpm-ostree

rfairley:
Add a polkit rule to allow the `zincati` user to make `deploy` and
`finalize-deployment` calls to the D-Bus API exposed by rpm-ostree.

### DIFF
--- a/dist/polkit-1/rules.d/zincati.rules
+++ b/dist/polkit-1/rules.d/zincati.rules
@@ -1,0 +1,8 @@
+// Allow Zincati to deploy, and finalize a staged deployment through rpm-ostree.
+polkit.addRule(function(action, subject) {
+    if ((action.id == "org.projectatomic.rpmostree1.deploy" ||
+         action.id == "org.projectatomic.rpmostree1.finalize-deployment") &&
+        subject.user == "zincati") {
+        return polkit.Result.YES;
+    }
+})


dist/polkit-1/rules.d/zincati.rules
```diff
@@ -0,0 +1,8 @@
+// Allow Zincati to deploy, and finalize a staged deployment through rpm-ostree.
+polkit.addRule(function(action, subject) {
+    if ((action.id == "org.projectatomic.rpmostree1.deploy" ||
+         action.id == "org.projectatomic.rpmostree1.finalize-deployment") &&
+         subject.user == "zincati") {
```
jlebon: Minor/optional: the indentation here makes it look like that final conditional is part of the same nested parentheses. Maybe just de-indent by one space?

rfairley: Right, thanks for catching - fixed!

rfairley:
Note https://www.freedesktop.org/software/polkit/docs/latest/polkit.8.html states `"applications, mechanisms and general-purpose operating systems must never include any authorization rules"`. However, Zincati is targeted for FCOS, and I would think that an exception is acceptable (similar to `/usr/share/polkit-1/rules.d/systemd-networkd.rules` for the `systemd-network` sysuser).

Alternatively the rule could be installed by the FCOS-specific overlay in https://github.com/coreos/fedora-coreos-config/tree/testing-devel/overlay, rather than the Zincati RPM, but ideally the rule is kept in the same codebase as the rest of `zincati`.

lucab:
I don't like the option where this rules comes from the overlay, so let's stick it in the RPM instead.

rfairley:
:+1: from me, on having the rule in the RPM.

jlebon:
> However, Zincati is targeted for FCOS, and I would think that an exception is acceptable

Agreed. I think we can consider FCOS to be a special-purpose operating system. :)

rfairley:
Going to merge this! Then, I'll merge https://src.fedoraproject.org/rpms/rust-zincati/pull-request/1 to get a build up.

Issue #62, opened 2019-07-05
Disabling Zincati


jlebon:
AFAICT, the Zincati config doesn't seem to have a master enable/disable switch, correct? Is the workflow simply to disable the systemd service if you don't want updates?

In some contexts, it's cleaner to do this via drop-in vs symlinking the systemd unit. Would it be possible to have e.g. a trivial `strategy = "disabled"` setting which just makes Zincati exit on startup?

lucab:
I initially had a strategy "[never](https://github.com/lucab/exp-zincati/blob/afa12fca405db09f62de44defd2d30931c3e03a7/src/strategy/never.rs)", I can re-introduce something to that effect.

The workflow via systemd-unit disabling works too, but it's indeed a different beast. 

Strategies are now controlling when to apply/finalize updates, so this may or may not fit in there as it has a slightly more powerful semantics, I need to explore this a bit more. Worst case, it gets its own top level boolean flag.
Also, I was thinking of jumping straight to the end-state, but without early exiting.

bgilbert:
For context: https://github.com/coreos/fedora-coreos-tracker/issues/163#issuecomment-508847356

The master disable switch probably shouldn't reuse an existing setting.  Otherwise, if an Ignition config for a prod stream sets a non-default strategy, and that config is used to test a non-prod image, it'll inadvertently re-enable Zincati.

lucab:
I do agree with the last comment. It's better to have an explicit `updates.enabled` boolean flag to disable auto-updates logic. That is also aligned with fcos-pinger configuration, which exposes a `reporting.enabled` entry.

PR #64, opened 2019-07-08
update_agent: add a knob to disable auto-updates logic

lucab:
This adds an `updates.enabled = <bool>` configuration entry to disable
auto-updates logic.

Closes: https://github.com/coreos/zincati/issues/62


lucab:
@rfairley in here there is also an additional `10-auto-updates.toml` configuration snippet to install via RPM.

rfairley:
@lucab ack, thanks - will update the RPM with the snippet with the next release.

PR #65, opened 2019-07-08
rpm-ostree: improve input validation

lucab:
This improves input validation on input releases/nodes coming
from the Cincinnati server, with relevant tests too.


src/rpm_ostree/mod.rs
```diff
@@ -40,3 +42,49 @@ impl Release {
         Ok(rel)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use maplit::hashmap;
```
jlebon: Ouhhh, I've definitely needed this in the past. Nice!

PR #66, opened 2019-07-08
rpm_ostree/deploy: record metrics

lucab:
This records some metrics on 'rpm-ostree deploy' runs, including total
attempts and failures.

### DIFF
--- a/src/rpm_ostree/cli_deploy.rs
+++ b/src/rpm_ostree/cli_deploy.rs
@@ -2,9 +2,22 @@
 
 use super::Release;
 use failure::{bail, format_err, Fallible, ResultExt};
+use prometheus::IntCounter;
+
+lazy_static::lazy_static! {
+    static ref DEPLOY_ATTEMPTS: IntCounter = register_int_counter!(opts!(
+        "zincati_rpm_ostree_deploy_attempts_total",
+        "Total number of 'rpm-ostree deploy' attempts."
+    )).unwrap();
+    static ref DEPLOY_FAILURES: IntCounter = register_int_counter!(opts!(
+        "zincati_rpm_ostree_deploy_failures_total",
+        "Total number of 'rpm-ostree deploy' failures."
+    )).unwrap();
+}
 
 /// Deploy an upgrade (by checksum) and leave the new deployment locked.
 pub fn deploy_locked(release: Release) -> Fallible<Release> {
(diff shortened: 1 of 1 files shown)


src/rpm_ostree/cli_deploy.rs
```diff
@@ -2,19 +2,36 @@
...
+use prometheus::{IntCounter, IntCounterVec};
+
+lazy_static::lazy_static! {
+    static ref DEPLOY_ATTEMPTS: IntCounter = register_int_counter!(opts!(
+        "zincati_rpm_ostree_deploy_attempts_total",
+        "Total number of 'rpm-ostree deploy' attempts."
+    )).unwrap();
+    static ref DEPLOY_RUNS: IntCounterVec = register_int_counter_vec!(opts!(
+        "zincati_rpm_ostree_deploy_runs_total",
+        "Total number of 'rpm-ostree deploy' runs."),
```
jlebon: Maybe `s/runs/results/`? Otherwise, it's not obvious what the difference is between attempt vs run.

src/rpm_ostree/cli_deploy.rs
```diff
@@ -2,19 +2,36 @@
...
 pub fn deploy_locked(release: Release) -> Fallible<Release> {
+    DEPLOY_ATTEMPTS.inc();
     let cmd = std::process::Command::new("rpm-ostree")
         .arg("deploy")
         .arg("--lock-finalization")
         .arg(format!("revision={}", release.checksum))
         .output()
         .with_context(|e| format_err!("failed to run rpm-ostree: {}", e))?;
 
+    let exit_code = cmd.status.code().unwrap_or(-1).to_string();
```
jlebon: Sure, though I don't think we really use exit codes much to express various failure modes. The only special one we have is 77 for the no-op case.

Also, remember because rpm-ostree is client/daemon structured, most failures are likely to happen in the daemon, and those just translate to exit code 1 at the CLI level.

lucab: Good point. I was mostly interested in counting failures and successes. Given that the exit code value per-se is uninteresting, I will just track those directly.

lucab:
@jlebon I got rid of exit-code tracking. This only tracks `deploy_attempts_total` and `deploy_failures_total` now.

Issue #55, opened 2019-07-03
identity: drop hardcoded stream label


lucab:
Zincati needs to introspect its own Fedora CoreOS stream label at runtime, in order to request the relevant update-graph to the Cincinnati backend. This information is a read-only property of the booted deployment, and it should be exposed by rpm-ostree.

We are currently using an hardcoded value at https://github.com/coreos/zincati/blob/025453b4637052281feb50ed00b7ce66cbe31b2f/src/identity/mod.rs#L129

We should replace it with proper stream label sourcing. This is currently blocked on https://github.com/coreos/fedora-coreos-tracker/issues/163.

PR #67, opened 2019-07-08
identity: introspect updates stream

lucab:
This removes the hardcoded stream value, introspecting it from booted
deployment metadata instead.

Closes: https://github.com/coreos/zincati/issues/55


lucab:
https://github.com/coreos/fedora-coreos-config/pull/110 got merged. Thus, this is ready to land.

PR #75, opened 2019-07-17
zincati: tweak logging

lucab:
This adjusts few minor log messages and levels, uniforming them.

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -96,9 +96,10 @@ impl UpdateAgent {
         let initialization = self.nop().map(|_r, actor, _ctx| {
             if actor.enabled {
                 actor.state.initialized();
+                log::info!("initialization complete, auto-updates logic enabled");
             } else {
-                log::info!("auto-updates logic disabled by configuration");
                 actor.state.end();
+                log::warn!("initialization complete, auto-updates logic disabled by configuration");
             }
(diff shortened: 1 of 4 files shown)


src/update_agent/actor.rs
```diff
@@ -112,7 +113,10 @@ impl UpdateAgent {
         let report_steady = self.strategy.report_steady(&self.identity);
         let state_change =
             actix::fut::wrap_future::<_, Self>(report_steady).map(|is_steady, actor, _ctx| {
-                actor.state.steady(is_steady);
+                if is_steady {
+                    actor.state.steady();
+                    log::debug!("reached steady state, periodically polling for updates");
```
jlebon: Should this just be `"reached steady state"`? How updates are actually pulled will depend on the chosen strategy, right?

lucab: In a recent cabal we decided that strategy only affects when the node **applies** update (i.e. reboots).
Thus with that logic, the check + deploy steps keep happening in background (periodically), till we are finally allowed to finalize a staged deployment.

As most of this logic does not exist yet (except for skeleton interfaces), we are still in time to make adjustments if the above sounds wrong.

jlebon: Gotcha, yup. That makes sense to me.

PR #77, opened 2019-07-17
dist/systemd: delay zincati boot order

lucab:
This tweaks Zincati service unit in order to start a bit later in
the boot process.

In particular, we want to be scheduled after two other targets
have been reached:
 * `network.target`: network is required to reach Cincinnati, so it
   should start setting up before zincati starts.
 * `boot-complete.target`: for remote coordination, releasing the
   locked update-slot should happen only if the system is fully
   operational after reboot.

### DIFF
--- a/dist/systemd/system/zincati.service
+++ b/dist/systemd/system/zincati.service
@@ -1,6 +1,11 @@
 [Unit]
 Description=Zincati Update Agent
 Documentation=https://github.com/coreos/zincati
+After=network.target
+# Wait for the boot to be marked as successful. In cluster contexts,
+# this prevents rolling out broken updates to all nodes in the fleet.
+Requires=boot-complete.target
+After=boot-complete.target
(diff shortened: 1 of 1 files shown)


dist/systemd/system/zincati.service
```diff
@@ -1,6 +1,9 @@
 [Unit]
 Description=Zincati Update Agent
 Documentation=https://github.com/coreos/zincati
+After=network.target
+Requires=boot-complete.target
```
jlebon: I feel like the `boot-complete.target` ones probably warrant a comment like:

```
# Make sure to start *after* a successful boot, so broken updates don't keep rolling out in cluster contexts
```

?

lucab: Indeed. Added a few explanatory words.

lucab:
Amended.

PR #86, opened 2019-07-26
update_agent: log relevant transaction events

lucab:
This augments logging at the default level in order to show when
(possibly long-running) rpm-ostree transactions start.

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -95,11 +95,11 @@ impl UpdateAgent {
 
         let initialization = self.nop().map(|_r, actor, _ctx| {
             if actor.enabled {
-                actor.state.initialized();
                 log::info!("initialization complete, auto-updates logic enabled");
+                actor.state.initialized();
             } else {
-                actor.state.end();
                 log::warn!("initialization complete, auto-updates logic disabled by configuration");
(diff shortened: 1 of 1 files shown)


src/update_agent/actor.rs
```diff
@@ -190,12 +190,16 @@ impl UpdateAgent {
             return Box::new(actix::fut::err(()));
         }
 
+        log::info!(
+            "new deployment '{}' selected, proceeding to stage it",
```
jlebon: Super minor: so, "deployment" has a very particular meaning in rpm-ostree. It more or less means "checked out OSTree commit". But we haven't deployed anything yet. Maybe e.g. `release` or `version` would fit better here?

src/update_agent/actor.rs
```diff
@@ -211,12 +215,16 @@ impl UpdateAgent {
             return Box::new(actix::fut::err(()));
         }
 
+        log::info!(
+            "staged deployment '{}' available, proceeding to finalize it",
```
jlebon: Whereas here `staged deployment` == :ok_hand:   :)

lucab:
@jlebon thanks for the suggestion, updated.

PR #92, opened 2019-08-01
rpm_ostree/deploy: add faulting tests for CLI calls 

lucab:
This adds some testing mocks around `rpm-ostree deploy`, starting
with simple metrics checks. Failpoints in here can be later re-used
in order to test consumers.

### DIFF
--- a/src/rpm_ostree/cli_deploy.rs
+++ b/src/rpm_ostree/cli_deploy.rs
@@ -18,6 +18,20 @@ lazy_static::lazy_static! {
 /// Deploy an upgrade (by checksum) and leave the new deployment locked.
 pub fn deploy_locked(release: Release) -> Fallible<Release> {
     DEPLOY_ATTEMPTS.inc();
+
+    let result = invoke_cli(release);
+    if result.is_err() {
+        DEPLOY_FAILURES.inc();
+    }
+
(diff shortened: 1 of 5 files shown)


src/rpm_ostree/cli_deploy.rs
```diff
@@ -26,12 +40,47 @@ pub fn deploy_locked(release: Release) -> Fallible<Release> {
...
+        fail::cfg("deploy_locked_err", "return").unwrap();
+
+        let release = Release {
+            version: "foo".to_string(),
+            checksum: "bar".to_string(),
+        };
+        let result = deploy_locked(release);
+        assert!(result.is_err());
+        assert!(DEPLOY_ATTEMPTS.get() >= 1);
+        assert!(DEPLOY_FAILURES.get() >= 1);
```
jlebon: Minor: could also check that `DEPLOY_ATTEMPTS == DEPLOY_FAILURES`.

lucab: I can't right now as the counters are shared between tests.
`FailScenario` is already guarding against races, but I need this in the library first for resetting the counters: https://github.com/pingcap/rust-prometheus/pull/261.

Issue #82, opened 2019-07-22
cincinnati: configure rollout wariness


lucab:
Cincinnati backend will add support for throttling new release rollouts over a period of time.

By default, the whole process is performed by the backend without any client explicit client involvement:
 1. at any point in time, a rolling-out release is assigned a throttling value `X`, sliding from 0.0 (start) to 1.0 (full rollout)
 1. each client request is arbitrarily assigned a rollout wariness score `Y`, between 0.00001 (bold) and 1.0 (careful)
 1. if `X < Y`, the update-graph (for this request) will be pruned of all incoming edges to the throttled release

However, it would be nice for the server to allow clients to pick their own wariness score.
On the client-side, this means exposing a `rollout_wariness` in configuration, and forwarding it to Cincinnati.

lucab:
This requires some throttling coordination with the server, which is ongoing at https://github.com/openshift/cincinnati/pull/130#issuecomment-507718938.

lucab:
For reference, current OpenShift draft has a "sucess rate" for releases and a "risk appetite" for clients, in the 0.0 to 1.0 range.
The pruning-decision logic is the same as the one described above.

lucab:
@arithx I have a PR up for this at https://github.com/coreos/zincati/pull/97. However I'd like some UX double-check, bikeshedding encouraged.

The proposed usage is a `identity.rollout_wariness = <float>` configuration entry, with allowed values in the `[0.0, 1.0]` range.

arithx:
I think that at a glance `identity` didn't seem like the most obvious grouping until I looked into the context of what else was contained in it.

As for `rollout_wariness` I think that does a good job at conveying the intent. If you're searching for other ideas in general maybe something like `update_tolerance` or `update_confidence`?

lucab:
> I think that at a glance `identity` didn't seem like the most obvious grouping until I looked into the context of what else was contained in it.

A few alternative options we could check:
 * rename the `identity` section. It currently contains `node_uuid` and `group`, which are sent to both Cincinnati and Airlock. I don't have better ideas for naming, though.
 * move to the `cincinnati` section. Though I was planning for this section to only container HTTP-releated parameters (address, auth, tls CA, etc.)
 * move to the `updates` section. At the moment it contains only "strategy" and enablement bool. I am not sure it is a good fit here, as we are talking about a parameter that is going to be sent to a remote server, while update knobs so far have only local effects.

[...]

PR #97, opened 2019-08-06
identity: allow setting rollout wariness

lucab:
This exposes an `identity.rollout_wariness = <float>` configuration
knob. It allows clients to pick their own constant wariness for
throttled rollout. Allowed values are within the range from `0.0`
(bold clients) to `1.0` (cautious clients), limits included.

Closes: https://github.com/coreos/zincati/issues/82

### DIFF
--- a/dist/config.d/10-identity.toml
+++ b/dist/config.d/10-identity.toml
@@ -1,0 +1,15 @@
+# Configure agent identity.
+[identity]
+
+# Node group, used for cluster-wide reboot orchestration (e.g. Airlock).
+group = "default"
+
+# Node ID, in sd-id128(3) format.
+# By default it is automatically computed as
+# `systemd-id128 machine-id -a de35106b6ec24688b63afddaa156679b`
+#node_uuid = "<ID128>"
+
+# Client wariness for throttled rollouts, as a floating point number.
+# Allowed values are within the range from `0.0` (very bold clients) to `1.0` (cautious clients), limits included.
+# By default, clients are arbitrarily throttled by the Cincinnati server.
+#rollout_wariness = 0.5
--- a/tests/fixtures/00-config-sample.toml
+++ b/tests/fixtures/00-config-sample.toml
@@ -1,17 +1,11 @@
 [identity]
-# Node group: used for cluster-wide reboot orchestration (via remote strategy)
 group = "workers"
+node_uuid = "27e3ac02af3946af995c9940e18b0cce"
+rollout_wariness = 0.5
 
-# Node UUID: by default derived from "/etc/machine-id"
-#node_uuid = "27e3ac02-af39-46af-995c-9940e18b0cce"
-
-# Base URL to the Cincinnati service
 [cincinnati]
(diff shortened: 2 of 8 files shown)


lucab:
@rfairley this depends on the `ordered-float`, which doesn't seem to be in fedora yet.

tests/fixtures/00-config-sample.toml
```diff
@@ -1,9 +1,8 @@
 [identity]
 # Node group: used for cluster-wide reboot orchestration (via remote strategy)
 group = "workers"
-
-# Node UUID: by default derived from "/etc/machine-id"
```
arithx: Was dropping this comment intentional?

tests/fixtures/00-config-sample.toml
```diff
@@ -1,9 +1,8 @@
...
-#node_uuid = "27e3ac02-af39-46af-995c-9940e18b0cce"
+node_uuid = "27e3ac02af3946af995c9940e18b0cce"
+rollout_wariness = 0.5
```
arithx: Worth adding a comment containing the range of values?

ignatenkobrain:
`ordered-float` is now in Fedora.

tests/fixtures/00-config-sample.toml
lucab: Yes, this is only a fixture file used for testing. I took a note to add a sample with comments under `dist/`, will add on next rebase.

lucab:
This should be now ready for review.

dist/config.d/10-identity.toml
```diff
@@ -0,0 +1,15 @@
+# Configure agent identity.
+[identity]
+
+# Node group, used for cluster-wide reboot orchestration (e.g. Airlock).
+group = "default"
+
+# Node ID, in sd-id128(3) format.
+# By default it is automatically computed as
+# `systemd-id128 machine-id -a de35106b6ec24688b63afddaa156679b`
```
jlebon: Not new here, though what's the rationale for doing `-a <app-id>` vs. just using the machine-id directly?

lucab: Because we need to send the uuid over the network (to Cincinnati backend), and the machine-id is not supposed to be leaked that way. See https://www.freedesktop.org/software/systemd/man/machine-id.html:

> This ID uniquely identifies the host. It should be considered "confidential", and must not be exposed in untrusted environments, in particular on the network. If a stable unique identifier that is tied to the machine is needed for some application, the machine ID or any part of it must not be used directly. Instead the machine ID should be hashed with a cryptographic, keyed hash function, using a fixed, application-specific key. That way the ID will be properly unique, and derived in a constant way from the machine ID but there will be no way to retrieve the original machine ID from the application-specific one. The sd_id128_get_machine_app_specific(3) API provides an implementation of such an algorithm.

lucab:
@arithx anything else left here or can I merge this one?

arithx:
@lucab I'm fine with merging

Issue #87, opened 2019-07-26
rpm-ostree: block downgrades, expose opt-in config boolean


lucab:
Zincati and rpm-ostree currently allow downgrades, if the Cincinnati server instruct them to do so.

We actively plan to use this feature for CI testing the update logic, but it is somehow a regression compared to other ostree-based systems that by default do not allow downgrades.
Thus we should also start denying downgrades by default, but allowing specific cases where deploying downgrades is fine.

In order to do that, we have to:
 * get the commit-timestamp for each release from Cincinnati
 * introduce a `updates.allow_downgrade` config boolean (default: false)
 * implement downgrade detection logic, based on timestamp comparison
 * filter out downgrade from viable updates
 * pass the relevant flag to `rpm-ostree deploy` (see https://github.com/projectatomic/rpm-ostree/pull/1876)

lucab:
Followup from https://github.com/coreos/zincati/pull/98: Zincati best-effort filtering logic will be based on the age-index from the Cincinnati server, while the actual no-downgrade-logic enforcement will be performed by rpm-ostree based on real timestamps.

PR #98, opened 2019-08-12
rpm_ostree: add release ordering based on age-index

lucab:
This implements total ordering on releases, based on metadata
provided by the Cincinnati server.
In particular, from now on we require a metadata entry named
`org.fedoraproject.coreos.releases.age_index` which allows us
to compare the age (i.e. positional index) of releases as they
originally appear in the stream release index.

### DIFF
--- a/src/rpm_ostree/mock_tests.rs
+++ b/src/rpm_ostree/mock_tests.rs
@@ -1,0 +1,55 @@
+use crate::cincinnati::Cincinnati;
+use crate::rpm_ostree::Release;
+use crate::identity::Identity;
+use mockito::{self, Matcher};
+use tokio::runtime::current_thread as rt;
+
+#[test]
+fn test_simple_graph() {
+    let simple_graph = r#"
+{
+  "nodes": [
+    {
+      "version": "0.0.0-mock",
+      "metadata": {
+        "org.fedoraproject.coreos.scheme": "checksum",
+        "org.fedoraproject.coreos.releases.age_index": "0"
+      },
+      "payload": "sha-mock"
+    },
+    {
+      "version": "30.20190725.0",
+      "metadata": {
+        "org.fedoraproject.coreos.scheme": "checksum",
+        "org.fedoraproject.coreos.releases.age_index": "1"
+      },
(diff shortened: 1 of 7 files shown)


lucab:
For context, I plan to use this total ordering to solve at least two currently open points.

First is to filter out downgrades from the list of possible targets in the update-graph received from Cincinnati, i.e. https://github.com/coreos/zincati/issues/87. Note that Zincati only needs to performs some kind of best-effort filtering, while rpm-ostree will enforce the final timestamp-based invariant via https://github.com/projectatomic/rpm-ostree/pull/1876.

Second is to improve our target selection logic to make it stable regardless of graph serialization, i.e. https://github.com/coreos/zincati/blob/384026c1588c2ddfbf8d28394b651387acb8d4b3/src/cincinnati/mod.rs#L134-L138

src/rpm_ostree/mock_tests.rs
```diff
@@ -0,0 +1,53 @@
...
+    let id = Identity::mock_default();
+    let client = Cincinnati {
+        base_url: mockito::server_url(),
+    };
+    let update = rt::block_on_all(client.fetch_update_hint(&id, true)).unwrap();
+    m_graph.assert();
+
+    let next = update.unwrap();
+
+    assert_eq!(next.version, "30.20190725.0")
```
jlebon: This is a cool test!

src/rpm_ostree/mod.rs
```diff
@@ -17,6 +21,35 @@ pub struct Release {
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        let self_age = self.age_index.clone().unwrap_or(0);
+        let other_age = other.age_index.clone().unwrap_or(0);
+
+        if self_age != other_age {
+            return self_age.cmp(&other_age);
+        }
+
+        if self.version != other.version {
+            return self.version.cmp(&other.version);
```
jlebon: OK, just to confirm: in theory, we shouldn't even get past this because the age index of all the nodes in the graph should be different, right?

This is probably worth a comment here that comparing version strings and checksums are just to get a stable order in that event, even though it's not necessarily a meaningful order at the high-level.

lucab: Correct!
I've added a comment to make that explicit.

PR #99, opened 2019-08-13
docs: update LICENSE content from apache.org

kshithijiyer:


lucab:
@kshithijiyer thanks for the PR! I'm not exactly sure how I managed to screw the first line of this file, but thanks for noticing.
Can you please just use the the file at https://www.apache.org/licenses/LICENSE-2.0.txt, keeping the appendix? In the future It would be useful to have the pristine copy, so that we can do a direct digest comparison to ensure the content is the same as the original one.

kshithijiyer:
> @kshithijiyer thanks for the PR! I'm not exactly sure how I managed to screw the first line of this file, but thanks for noticing.
> Can you please just use the the file at https://www.apache.org/licenses/LICENSE-2.0.txt, keeping the appendix? In the future It would be useful to have the pristine copy, so that we can do a direct digest comparison to ensure the content is the same as the original one.

Done

PR #102, opened 2019-08-15
updates: block downgrades, expose opt-in config

lucab:
This denies downgrades by default, but introduces a configuration
boolean `updates.allow_downgrade` (default: `false`) to disable
such logic for setups where deploying downgrades is fine.

Zincati only performs best-effort filtering based on the age-index
from the Cincinnati server, while the final enforcement is performed
by rpm-ostree based on real timestamps.

Closes: https://github.com/coreos/zincati/issues/87

### DIFF
--- a/docs/usage/auto-updates.md
+++ b/docs/usage/auto-updates.md
@@ -55,6 +55,24 @@ For further documentation on configurations, check the [updates strategy][update
 [strategy-fleet_lock]: updates-strategy.md#lock-based-strategy
 [updates-strategy]: updates-strategy.md
 
+## Updates ordering and downgrades
+
+OS updates have a strict ascending ordering called "age index", which is based on the date and time of release.
+Versions that have been released earlier in time have a lower index than recent ones.
+
+Zincati uses this absolute ordering to prefer newer releases (i.e. with higher age index) when multiple updates are available at the same time.
+By default, this ordering is also used to prevent automatic downgrades.
+
+For custom environments where automatic downgrades have to be supported, the following configuration snippet can be used to enable them:
+
+```toml
+[updates]
+allow_downgrade = true
+```
+
+Enabling such logic removes an additional safety check, and may allow rogue Cincinnati servers to induce downgrades to old releases with known security vulnerabilities.
+It is generally not recommended to allow and perform automatic downgrades via Zincati.
+
 ## Disabling auto-updates
 
 To disable auto-updates, a configuration snippet containing the following has to be installed on the system:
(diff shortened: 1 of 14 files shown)


lucab:
This is currently based on https://github.com/coreos/zincati/pull/112 and will need to be rebased once that is merged into master.

lucab:
Unblocked, rebased on latest master. This is now ready for review.

/cc @jlebon @ashcrow

docs/usage/auto-updates.md
```diff
@@ -55,6 +55,24 @@ For further documentation on configurations, check the [updates strategy][update
...
+By default, this ordering is also used to prevent automatic downgrades.
+
+For custom environments where automatic downgrades have to be supported, the following configuration snippet can be used to enable them:
+
+```toml
+[updates]
+allow_downgrade = true
+```
+
+Enabling such logic removes an additional safety check which may be valuable in case of rogue Cincinnati servers.
```
ashcrow: This line reads a bit off to me. I believe it's meant to note that removing the logic would allow a rogue/malicious Cincinnati server to provide compromised or broken updates. What about:

```
[...] may allow rogue Cincinnati servers to provide broken or malicious updates.
```

or something like that?

src/cincinnati/mod.rs
```diff
@@ -200,6 +207,15 @@ fn find_update(
         None => return Ok(None),
     };
 
+    // Check for downgrades.
+    if next <= cur_release {
+        log::warn!("downgrade hint towards target release '{}'", next.version);
```
ashcrow: :+1:

src/rpm_ostree/actor.rs
```diff
@@ -24,6 +24,8 @@ impl RpmOstreeClient {
 /// Request: stage a deployment (in finalization-locked mode).
 #[derive(Debug, Clone)]
 pub struct StageDeployment {
+    /// Whether to allow downgrades.
+    pub allow_downgrade: bool,
```
jlebon: Feels more like it'd be a property of the actor (e.g. a field of `RpmOstreeClient`) than a dynamic field from the message, though no strong opinion! I guess if Zincati learns to support `systemctl reload` in the future, it's easier this way.

lucab: The remark about `reload` is interesting (I still don't know how to fit it in Zincati). The boolean setting itself has effects on both rpm-ostree and cincinnati logic, so I'd prefer to keep it in the main actor.

jlebon: Yup, SGTM.

lucab:
@ashcrow good point.

I've added a warning log and a metrics (so that it can be monitored+aggregated externally), how does it look now?

PR #103, opened 2019-08-15
dist/systemd: change binary path to /usr/libexec

rfairley:
Change the path to the installed binary from `/usr/bin/zincati` to
`/usr/libexec/zincati`. This is done because `zincati` is not
currently intended to be run directly by users.

When installing Zincati and running Zincati through this unit, the
binary placed by `cargo install` should be moved to this location
under `/usr/libexec`.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/244


rfairley:
Required by: https://src.fedoraproject.org/rpms/rust-zincati/pull-request/2

PR #104, opened 2019-08-19
identity: send OS checksum to Cincinnati

lucab:
This adds back the OS checksum in Cincinnati client parameters,
which was dropped by mistake in https://github.com/coreos/zincati/pull/101.


lucab:
/cc @jlebon for review.

For context, I realized we still need this parameter in order to reflect back the checksum for things like https://github.com/coreos/fedora-coreos-tracker/issues/228#issuecomment-522559397.

PR #106, opened 2019-08-20
identity: track rollout wariness in metrics

lucab:
This adds a metrics gauge to track and expose user-configured
rollout wariness.


jlebon:
Do you have more context on this? It's static (i.e. won't change during the lifetime of the app), and it's from a config (which means in a cluster context we would've configured it ourselves... e.g. through Ignition).

lucab:
@jlebon correct. But this is meant for an external observation system to track this parameter fleet-wide. For example, you can observe updates+reboots going through the cluster and make sure they rollout according to the configured wariness, without having to locally inspect every single node.

Issue #37, opened 2019-06-13
strategy: add configuration and logic for remote coordination


lucab:
In https://github.com/coreos/fedora-coreos-tracker/issues/3 we decided to move the remote coordination logic into a separate out-of-host process.

https://github.com/coreos/airlock implements a containerizable version of that. The current proposed version of the remote protocol is https://github.com/coreos/airlock/pull/1.

On zincati side, we need to:
 * implement an airlock client library
 * reserve a "strategy" kind for this mode
 * define all configuration options we requires (bare minimum: "base_url")

PR #121, opened 2019-09-03
strategy: add FleetLock cluster-wide reboot coordination

lucab:
This adds configuration and logic for a `fleet_lock` update strategy,
which allows cluster-wide reboot coordination.

Protocol: https://github.com/coreos/airlock/pull/1
Docs: https://github.com/coreos/zincati/pull/134
Closes: https://github.com/coreos/zincati/issues/37

### DIFF
--- a/src/fleet_lock/mod.rs
+++ b/src/fleet_lock/mod.rs
@@ -1,0 +1,261 @@
+//! Asynchronous FleetLock client, remote lock management.
+//!
+//! This module implements a client for FleetLock, a bare HTTP
+//! protocol for managing cluster-wide reboot via a remote
+//! lock manager. Protocol specification is currently in progress at
+//! https://github.com/coreos/airlock/pull/1.
+
+use crate::identity::Identity;
+use failure::{Error, Fail, Fallible, ResultExt};
+use futures::future;
+use futures::prelude::*;
+use reqwest::r#async as asynchro;
+use reqwest::Method;
+use serde::{Deserialize, Serialize};
+
+#[cfg(test)]
+mod mock_tests;
+
+/// FleetLock pre-reboot API path endpoint (v1).
+static V1_PRE_REBOOT: &str = "v1/pre-reboot";
+
+/// FleetLock steady-state API path endpoint (v1).
+static V1_STEADY_STATE: &str = "v1/steady-state";
+
+/// Error from lock manager.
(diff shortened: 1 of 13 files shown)


lucab:
Work in progress, currently missing:
 - [ ] unit tests and mocks
 - [ ] docs on configuration and semantics
 - [ ] field testing

lucab:
Field testing is :heavy_check_mark:.

I set up an [airlock](https://github.com/coreos/airlock) deployment on Fedora communishift.

Zincati configuration to use that is:
```toml
[updates]
strategy = "fleet_lock"

[updates.fleet_lock]
base_url = "https://airlock-lucab.apps.os.fedorainfracloud.org"
```

Played with manually locking/unlocking, in the middle of a `30.20190801.0` -> `30.20190905.0` upgrade. Client successfully locks and unlocks the semaphore, and retries if there is no semaphore slot available at the moment.

When all semaphore slots are taken, Zincati will output this to the journal logs (message is coming from the server and can be improved):
```
[ERROR] lock manager pre-reboot rejection, code 500: all 1 semaphore slots currently locked
```

lucab:
I've split the documentation part to https://github.com/coreos/zincati/pull/134 so that it should be slightly easier to review in smaller pieces.

src/fleet_lock/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+            self.endpoint,
+            self.status.as_u16(),
+            self.value
+        )
+    }
+}
+
+impl LockRejection {
+    /// Convert a response into a lock result.
+    fn error_for_response(
```
jlebon: This name honestly confused me for a bit because I didn't realized it also handled the happy case. I think it'd be easier to understand as a stand-alone function named e.g. `map_response()` or something?

lucab: You are right, this is overly confusing to follow.
Let me see if I can actually manage to scope this back to the error-path only, which I think would make this even easier to read. If that is not feasible for some reason, I'll rename as suggested,

lucab:
@jlebon I've piled a fixup commit on top and added some [guide-comments](http://antirez.com/news/124). Does it look better now?

lucab:
Did the final rebase to master (without code changes) and manual tests.

This is how a cluster-wide auto-update looks like, while rolling through a cluster of five nodes via a lock-manager with one reboot slot:
grafana

PR #123, opened 2019-09-04
zincati: update trait object syntax

lucab:
This fixes several old-style trait object signatures, to avoid all
warnings like "trait objects without an explicit `dyn` are deprecated".


lucab:
@jlebon this should be another easy-peasy review if you ever get bored of your backlog ;)

PR #126, opened 2019-09-05
dist/systemd: don't start Zincati if booted as a live system

bgilbert:
...such as on a live system.  We can't perform updates in that case.

### DIFF
--- a/dist/systemd/system/zincati.service
+++ b/dist/systemd/system/zincati.service
@@ -1,6 +1,8 @@
 [Unit]
 Description=Zincati Update Agent
 Documentation=https://github.com/coreos/zincati
+# Skip live systems not meant to be auto-updated (e.g. live PXE, live ISO)
+ConditionPathExists=!/run/ostree-live
 After=network.target
 # Wait for the boot to be marked as successful. In cluster contexts,
 # this prevents rolling out broken updates to all nodes in the fleet.


lucab:
Approach SGTM.
Let's maybe add an explicit comment for context? Something like `# Skip read-only live systems, not meant to be auto-updated (e.g. live-PXE, live-ISO)`.

@jlebon @cgwalters is this a good invariant we can guarantee on rpm-ostree side to uphold?

cgwalters:
> @jlebon @cgwalters is this a good invariant we can guarantee on rpm-ostree side to uphold?

Depends on the implementation of `ConditionPathIsReadWrite=`...technically, the `/` mount point is writable, but it also has the immutable bit there.

In the end we need to drop a stamp file I think of some form to tell libostree what's going on, and I think we could also have Zincati key off that?

Not having a `/boot` directory may also be a potential key?

lucab:
Per systemd docs, that conditions checks "whether the underlying file system is readable and writable (i.e. not mounted read-only)".

bgilbert:
Updated.

lucab:
Self note: `/run/ostree-live` has been agreed upon in https://github.com/coreos/fedora-coreos-config/pull/155#issuecomment-528487271.

Issue #139, opened 2019-09-19
update_agent: introduce some jitter in the refresh loop


lucab:
The refresh loop in the update-agent currently has a fixed duration, equal for all clients. In the steady state, the agent queries the Cincinnati service at the beginning of each loop.

This means that under specific adverse circumstances (e.g. a smart circuit breaker in a reverse-proxy between the client and the service), the backend could involuntarily phase-lock all the clients into the exact same synchronized loop. This can in turn exacerbate any existing backend problem.

I *think* I've actually observed this behavior on the staging cluster, due to network hiccups. See the oscillation in the rate of requests, with peak-to-dip amplitudes that increase over time:
![p1](https://user-images.githubusercontent.com/98086/65234484-35a59500-dac4-11e9-9c68-b4dd2921d932.png)

Suggestion is to introduce some random jitter in the loop duration, e.g. `P + x*P/100` with `P` being the existing fixed duration and `x` a random integer in `[0, 10]`.

PR #140, opened 2019-09-19
update_agent: introduce some jitter in the refresh loop

lucab:
This adds a small random amount of jitter (0% to 10%) to the refresh
loop period. It is meant to prevent multiple clients from converging
into a problematic stable state where all the (distributed) refresh
events are isochronous.

Closes: https://github.com/coreos/zincati/issues/139

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -67,11 +67,12 @@ impl Handler<RefreshTick> for UpdateAgent {
                 actor.state_changed = update_timestamp;
                 Self::tick_now(ctx);
             } else {
+                let pause = Self::add_jitter(actor.refresh_period);
                 log::trace!(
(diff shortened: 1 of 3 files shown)


src/update_agent/actor.rs
```diff
@@ -94,6 +95,19 @@ impl UpdateAgent {
         ctx.notify_later(RefreshTick {}, after)
     }
 
+    /// Apply a small, random amount (0% to 10%) of jitter to a given period.
```
ashcrow: This is creating and returning the duration for jittering, not actually applying it, correct?

lucab: This is doing the sum internally and returning the total, but not mutating the fixed period in input.
I've slightly renamed this, it should be a bit clearer now.

PR #147, opened 2019-09-24
identity: rename OS version metrics

lucab:
This renames the OS version label to be less generic, making its
usage more meaningful at query-time.


lucab:
This follows https://www.robustperception.io/exposing-the-software-version-to-prometheus and a few more similar suggestions I got out-of-band.

PR #149, opened 2019-09-25
fleet_lock: update protocol parameters

lucab:
This updates the FleetLock client for a last-minute change in protocol
parameters naming.


lucab:
This goes together with https://github.com/coreos/airlock/pull/27.

PR #152, opened 2019-09-27
cincinnati: uniform and instrument error kinds

lucab:
This uniforms all Cincinnati-related logic to use a module-specific
error type. The intent is to make sure each failure can be recorded
with machine-friendly and human-friendly identifiers.
For server errors, it optionally tracks the status code too.

It also augments error metrics with the machine-friendly
kind. The resulting metrics now looks like this:
```
zincati_cincinnati_update_checks_errors_total{kind="generic_http_418"} 42
```

Which in turn can be aggregated and visualized like this:
errors-rate

### DIFF
--- a/src/cincinnati/client.rs
+++ b/src/cincinnati/client.rs
@@ -8,12 +8,12 @@
 
 #![allow(unused)]
 
-use failure::{format_err, Error, Fallible, ResultExt};
+use failure::{format_err, Error, Fail, Fallible, ResultExt};
 use futures::future;
 use futures::prelude::*;
 use reqwest::r#async as asynchro;
 use reqwest::Method;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 /// Cincinnati graph API path endpoint (v1).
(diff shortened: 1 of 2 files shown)


src/cincinnati/client.rs
```diff
@@ -34,6 +34,33 @@ pub struct Graph {
     pub edges: Vec<(u64, u64)>,
 }
 
+/// Error related to the graph service.
+#[derive(Clone, Debug, Fail, Deserialize, Serialize, PartialEq, Eq)]
+pub struct GraphError {
+    /// HTTP status code returned by the server.
+    #[serde(skip)]
+    pub(crate) status: Option<reqwest::StatusCode>,
```
jlebon: Minor: I'd call this `status_code` to make it more obvious why client-side failures have `None`.

src/cincinnati/client.rs
```diff
@@ -34,6 +34,33 @@ pub struct Graph {
...
+    /// Machine-friendly brief error kind.
+    pub(crate) kind: String,
+    /// Human-friendly detailed error explanation.
+    pub(crate) value: String,
```
jlebon: Optional: hmm, sounds like we could group them into an enum? E.g. each member being a different error and have `Display` be the human-friendly one and an impl function for the machine one.

src/cincinnati/mod.rs
```diff
@@ -39,10 +39,11 @@ lazy_static::lazy_static! {
         "zincati_cincinnati_update_checks_total",
         "Total number of checks for updates to the upstream Cincinnati server."
     )).unwrap();
-    static ref UPDATE_CHECKS_ERRORS: IntCounter = register_int_counter!(opts!(
+    static ref UPDATE_CHECKS_ERRORS: IntCounterVec = register_int_counter_vec!(
         "zincati_cincinnati_update_checks_errors_total",
-        "Total number of errors on checks for updates."
-    )).unwrap();
+        "Total number of errors while checking for updates.",
+        &["kind"]
```
jlebon: That's cool!

src/cincinnati/client.rs
lucab: I'm not sure I'm following you on this one. Mind writing down the enum variants you are thinking about? Also, do note that in some cases `kind` is dynamic (e.g. the ones derived from a raw status code).

jlebon: Something like:

```rust
enum GraphError {
    HTTP(u32),
    JsonDecode(String),
    ...
}

impl GraphError {
    fn machine_error(&self) -> String { ... }
}

impl std::fmt::Display for GraphError {
    ...
}
```

?

Feel free to disregard if it's not worth the overhead. :)

lucab:
@jlebon I've reworked this to use an explicit enum with error variants. Does it look better now? Mind another review pass?

src/cincinnati/client.rs
```diff
@@ -34,6 +34,81 @@ pub struct Graph {
     pub edges: Vec<(u64, u64)>,
 }
 
+/// Cincinnati JSON protocol: service error.
```
jlebon: Ahh OK, I missed at first that this was actually part of the protocol itself until I saw the code deserializing straight from the response. Maybe link to the relevant part of the protocol here?

lucab: You missed it because I really blurred it in the previous iteration. The whole protocol is linked in the top (module-level) docstring, is that enough?

jlebon: Ahh gotcha, yup that's fine!

Issue #90, opened 2019-07-29
cincinnati: expose dead-end release information


lucab:
There is a vague idea for the Cincinnati server to encode the "version is a dead-end" information in the graph served to clients.

If/when that happens, Zincati could take a proactive approach and notify that somehow to a monitoring system or to a human (e.g. via metrics/env-file/motd/etc).

lucab:
@bgilbert behavioral doubt: what should the client do when it detects that it is running on a dead-end release? Should it kept polling for updates (in case the server changes its graph) or can it assume it will never get out of it and park itself?

bgilbert:
Doesn't Cincinnati inherently encode that a version is a dead-end, by not having any out-edges?

IMO it should keep polling.  Otherwise, if we ever published a graph with incorrect dead-end nodes (or even just changed our minds), there'd be no way to fix it without user intervention.

PR #156, opened 2019-10-07
cincinnati: evaluate dead-end status for booted OS

lucab:
This adds logic to evaluate and record whether the booted deployment
is a dead-end release from the point of view of auto-updates graph
(i.e. it cannot further auto-update).
As a dead-end release requires manual intervention to get out of its
stuck state, this piece of information has to be bubbled-up to
monitoring systems (e.g. via metrics).

Ref: https://github.com/coreos/zincati/issues/90

### DIFF
--- a/src/cincinnati/mod.rs
+++ b/src/cincinnati/mod.rs
@@ -23,6 +23,12 @@ pub static AGE_INDEX_KEY: &str = "org.fedoraproject.coreos.releases.age_index";
 /// Metadata key for payload scheme.
 pub static SCHEME_KEY: &str = "org.fedoraproject.coreos.scheme";
 
+/// Metadata key for dead-end sentinel.
+pub static DEADEND_KEY: &str = "org.fedoraproject.coreos.updates.deadend";
+
+/// Metadata key for dead-end reason.
+pub static DEADEND_REASON_KEY: &str = "org.fedoraproject.coreos.updates.deadend_reason";
+
 /// Metadata value for "checksum" payload scheme.
 pub static CHECKSUM_SCHEME: &str = "checksum";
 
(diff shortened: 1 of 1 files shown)


lucab:
I have done some manual testing and this seems to work as expected. When booted OS is detected as a dead-end, the following content is recorded.

Metrics:
```
zincati_cincinnati_release_is_deadend{version="1.2.3"} 1
```

Logs:
```
[WARN ] booted OS detected as a dead-end release, reason: manual test
```

src/cincinnati/mod.rs
```diff
@@ -35,6 +41,11 @@ lazy_static::lazy_static! {
         "zincati_cincinnati_graph_edges_count",
         "Number of edges in Cincinnati update graph."
     )).unwrap();
+    static ref RELEASE_IS_DEADEND: IntGaugeVec = register_int_gauge_vec!(
+        "zincati_cincinnati_release_is_deadend",
+        "Whether current release is a dead-end.",
+        &["version"]
```
jlebon: Don't we already expose the OS version via `zincati_identity_os_info` ? This could just be a single `IntGauge` no?

src/cincinnati/mod.rs
```diff
@@ -130,16 +143,25 @@ fn find_update(
...
     {
-        Some(pos) => pos,
+        Some(current) => current,
         None => return Ok(None),
     };
     drop(booted_depl);
 
+    // Evaluate and record whether booted OS is a dead-end release.
+    if let Some(reason) = evaluate_deadend(&cur_release) {
+        log::warn!(
```
jlebon: Will we emit this warning on each tick? I wonder here if it'd be cleaner to just warn once instead.

lucab: Yes, I had the same doubt. The actual behavior we want is a bit more complex and consists of:
 1) a flip-flop logic which triggers only at the edge of state changes
 2) one (or more) well-known place that sysadmin should setup some automated monitoring on

src/cincinnati/mod.rs
lucab: Ack, done.

lucab:
@jlebon I've pushed a fixup commit, which drops logging and only acts on metrics for the moment. If you don't mind, I'd like to buy myself a bit of time to explore how the final picture on this would look like. PTAL.

PR #167, opened 2019-10-29
ci: bump minimum supported Rust version to 1.38

lucab:
Bumping to 1.38 in order to cope with dependency updates.


lucab:
@jlebon yup, Zincati is FCOS-only pretty much by design. We are keeping the rest (Afterburn, and soon coreos-installer) at 1.37.0.

Issue #203, opened 2020-01-20
Allow customizing check frequency


jlebon:
Writing an update test, and I'd like Zincati to be able to check for updates much faster for the purpose of the test. I see there's a related TODO to this:

https://github.com/coreos/zincati/blob/eb55d9364ab5afbc441488c8fede9f2a9279c898/src/update_agent/mod.rs#L157-L160

Config knob works, or if we're not ready to commit to an interface yet, an env var also would be OK.

lucab:
Yes, I agree it is a good thing to fix that TODO and make this configurable.
The first step in that direction is to split this field, so that each state has its own pause duration.

Once there, what should the configuration looks like? It sounds like you need an aggressive refresh loop while the agent is in the steady state. A non-zero number of seconds may suit?

How does the below strawman sound?
```toml

# Configure update-agent pace.
[agent.timing]
# Pausing interval between updates checks in steady mode, in seconds.
steady_interval_secs = N

## Further settings, maybe later.
# Pausing interval between finalization attempts, in seconds.
# finalization_interval_secs = N
```

jlebon:
> How does the below strawman sound?

Yup, that strawman works for me! My suggestion is to only add knobs as the need comes up so that we aren't unnecessarily tying ourselves down. E.g. a future `finalization_interval_secs` feels like overkill to me since it's really something that should rarely fail.

lucab:
Fair, I don't plan to have all of them upfront. Just future-proofing the format so that we can fit more knobs without pain when/if we'll need them in the future.

PR #205, opened 2020-01-21
update_agent: make intervals state specific

lucab:
This tweaks refresh logic in order to make the refresh interval
specific to each state. It will allow exposing knobs to tweak
refresh delays for some states, later on.

Ref: https://github.com/coreos/zincati/issues/203

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -7,6 +7,7 @@ use failure::Error;
 use log::trace;
 use prometheus::IntGauge;
 use std::collections::BTreeSet;
+use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


lucab:
> Right, though concretely do you see duration knobs being useful for other states apart from the steady state? All the other other ones are mostly transitional, right?

I have another state-transition as a target in my mind, which is the one between `staged` -> `finalized`.
That edge is gated by a `strategy.can_finalize() -> bool` which can be true/false at different times, based for example on cluster state (fleet-lock) or time of the week (periodic). For this case it may be worth tweaking the interval between checks.

PR #218, opened 2020-02-10
zincati: use system proxy in HTTP clients

lucab:
This adds support to detect system proxy from environment variables,
using it for all HTTP clients (i.e. cincinnati and fleet_lock).

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/379


zonggen:
:+1:

jamescassell:
Makes sense, but the underlying library is broken in that it isn't enabled by default.  Also, the lack of no_proxy support will break anyone running the service inside their firewall.

PR #219, opened 2020-02-10
config: add agent timing knobs

lucab:
This adds support for a new config section `agent.timing` to configure
timing-related settings on the agent.
It currently has a single knob `steady_interval_secs` to tweak the refresh
period while in steady state.

Closes: https://github.com/coreos/zincati/issues/203

### DIFF
--- a/dist/config.d/10-agent.toml
+++ b/dist/config.d/10-agent.toml
@@ -1,0 +1,5 @@
+# Configure agent timing.
+[agent.timing]
+
+# Pausing interval between updates checks in steady mode, in seconds.
+steady_interval_secs = 300
--- a/src/config/fragments.rs
+++ b/src/config/fragments.rs
@@ -2,10 +2,13 @@
 
 use ordered_float::NotNan;
 use serde::Deserialize;
+use std::num::NonZeroU64;
 
 /// Top-level configuration stanza.
 #[derive(Debug, Deserialize, PartialEq, Eq)]
 pub(crate) struct ConfigFragment {
+    /// Agent configuration.
+    pub(crate) agent: Option<AgentFragment>,
     /// Cincinnati client configuration.
     pub(crate) cincinnati: Option<CincinnatiFragment>,
     /// Agent identity.
@@ -14,6 +17,21 @@ pub(crate) struct ConfigFragment {
     pub(crate) updates: Option<UpdateFragment>,
 }
 
+/// Config fragment for agent settings.
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+pub(crate) struct AgentFragment {
+    /// Timing settings for the agent.
+    pub(crate) timing: Option<AgentTiming>,
+}
+
+/// Config fragment for agent timing.
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+pub(crate) struct AgentTiming {
+    /// Pausing interval between updates checks in steady mode, in seconds (default: 300).
+    pub(crate) steady_interval_secs: Option<NonZeroU64>,
+}
+
+// Config fragment for agent identity.
 #[derive(Debug, Deserialize, PartialEq, Eq)]
 pub(crate) struct IdentityFragment {
     /// Update group for this agent (default: 'default')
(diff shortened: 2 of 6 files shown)


src/config/fragments.rs
```diff
@@ -14,6 +17,21 @@ pub(crate) struct ConfigFragment {
     pub(crate) updates: Option<UpdateFragment>,
 }
 
+// Config fragment for agent settings.
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+pub(crate) struct AgentFragment {
+    /// Update group for this agent (default: 'default')
```
jlebon: Wrong comment

dist/config.d/10-agent.toml
```diff
@@ -0,0 +1,5 @@
+# Configure agent timing.
+[agent.timing]
+
+# Pausing interval between updates checks in steady mode, in seconds.
+steady_interval_secs = 300
```
jlebon: Aside: hmm, any reason why we're shipping defaults in separate fragments vs. just a single one called e.g. `01-defaults.toml`? Is there a method to the priorities those separate fragments have (I see we're using 10, 30, and 50)?

Related to this, I wonder if we should do something similar to systemd where we just have the defaults commented out, so that it also serves as documentation on all the settings that exist.

Anyway, not blocking this of course!

src/update_agent/mod.rs
```diff
@@ -13,7 +13,9 @@ use prometheus::IntGauge;
 use std::time::Duration;
 
 /// Default tick/refresh period for the state machine (in seconds).
-const DEFAULT_REFRESH_PERIOD_SECS: u64 = 5 * 60;
+const DEFAULT_REFRESH_PERIOD_SECS: u64 = 300; // 5 minutes.
```
jlebon: Is this used anywhere now?

dist/config.d/10-agent.toml
lucab: I think you are basically asking for a single file with no real content, but only docstring and commented default values.

The approach we took so far is to have explicit defaults set via config. Those are split in smaller files so that it's easier to completely discard specific sections by placing a null-symlink with the same filename.

Priorities are somewhat arbitraries, but I'm using the `[10, 50]` range to make it easier to override by a user.

src/update_agent/mod.rs
lucab: Yes, for non-steady states, here: https://github.com/coreos/zincati/blob/70fe933c63dac5492fb18bd2cf6679c3a0074c4d/src/update_agent/actor.rs#L115

jlebon: Ahh heh, I had grepped my local copy of Zincati, but hadn't `git pull`ed the latest tree. :)

lucab: This reminds me I should enable `clippy` here, which would block the PR on unused items.

lucab: `clippy` followup at https://github.com/coreos/zincati/pull/220.

dist/config.d/10-agent.toml
jlebon: > I think you are basically asking for a single file with no real content, but only docstring and commented default values.

Yes, exactly. See e.g. https://github.com/systemd/systemd/blob/d900701eeab4a43b6626256226c68ed00a63962d/src/journal/journald.conf.

> The approach we took so far is to have explicit defaults set via config. Those are split in smaller files so that it's easier to completely discard specific sections by placing a null-symlink with the same filename.

I think the thing that seems confusing to me is that we're essentially shipping two sets of defaults: compiled-in options, and these fragments. (Though I think only the non-string ones actually have compiled-in defaults and the rest Zincati would error out on if the config is missing?). What does it mean to "mask" a default fragment? We'd just fall back to the compiled default, no? I can see the value of masking though if there are multiple layered fragments injected by other apps/layers that you want to cancel out.

Re. priorities, I think it'd be clearer if they were all e.g. `01`. Otherwise, we're just making the range under the priority we choose useless for that specific knob, no?

(Sorry, probably should split this into a separate issue for discussion!)

lucab: Yes, we can split to another ticket.

I'd say the approach so far has been more similar to https://github.com/systemd/systemd/tree/d900701eeab4a43b6626256226c68ed00a63962d/sysctl.d instead.

We leave a bit of space before (<10) for out-of-repo defaults, and most of the range (>50) for overrides.

As you noted, some in-code defaults are not the same as the config, and just fail validation if the fragment is masked.

PR #220, opened 2020-02-10
ci: add a clippy run to travis

lucab:
This fixes all existing warnings and adds a `clippy` job to travis.


lucab:
> I think Travis is missing clippy

Indeed, shame on me. CI should be green now.

Issue #223, opened 2020-02-11
zincati: port to futures-0.3


lucab:
There is an API bump for futures-0.3 pending at https://github.com/coreos/zincati/pull/216.
This will in turn unblock all the other queued dependency bumps (tokio+reqwest+actix).

Unfortunately this requires some invasive changes to port the codebase to stdlib futures and to new API for all dependencies. Additionally, it will need to go in a single shot due to dependency coupling.

PR #228, opened 2020-02-20
zincati: rework to use std::future

lucab:
This reworks the whole codebase to use new-style `Futures` from
Rust stdlib, and futures-0.3 combinators.
In turn, this allows updating all other queued dependency versions
(tokio, reqwest, and actix).

Closes: https://github.com/coreos/zincati/issues/223

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.38.0  # minimum supported toolchain
+  - 1.39.0  # minimum supported toolchain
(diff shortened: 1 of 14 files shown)


lucab:
@jlebon there is probably more room to simplify signatures/combinators even further, but I'm still in the process of porting my previous asyncio knowledge to the new stdlib model. 

This is almost purely a compiler-driver rework. I manually tested the resulting binary on `testing`-1, and it properly updated to lastest release.

Issue #229, opened 2020-02-21
update-agent: rework steady/polling state


lucab:
There is currently some friction in the update-state state machine design, in particular around the steady state and the polling logic.

Right now there are three relevant states:
 * `Initialized`: configuration has been successfully parsed, updates are not disabled, polling is about to start
 * `Steady`: one of
    - initialization is done, but haven't yet checked Cincinnati (there may or may not be updates available)
   - last Cincinnati check didn't return any target (there are no updates available)
 * `UpdateAvailable`: last Cincinnati check returned at least an available update, which we are not targeting

I'm feeling that the steady state is overloaded, as the first sub-bullet is an intermediate and specific situation. In particular, at point the agent doesn't actually know if there is an update available.
It may be better to split it into its own state-node (e.g. `UpdateCheckKickoff`), so that we can:
 1. know when we got the first result from Cincinnati
 2. better report when there are no updates available.

PR #233, opened 2020-02-27
update-agent: rework steady/polling state

lucab:
This splits the "reported steady" and "checked, but no updates
available" states, making it easier to track and report upgrade
progress.

Closes: https://github.com/coreos/zincati/issues/229

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -61,20 +61,21 @@ impl Handler<RefreshTick> for UpdateAgent {
         let prev_state = self.state.clone();
 
         let state_action = match &self.state {
-            UpdateAgentState::StartState => self.initialize(),
-            UpdateAgentState::Initialized => self.try_steady(),
-            UpdateAgentState::Steady => self.try_check_updates(),
+            UpdateAgentState::StartState => self.tick_initialize(),
+            UpdateAgentState::Initialized => self.tick_report_steady(),
+            UpdateAgentState::ReportedSteady => self.tick_check_updates(),
+            UpdateAgentState::NoNewUpdate => self.tick_check_updates(),
             UpdateAgentState::UpdateAvailable(release) => {
                 let update = release.clone();
-                self.try_stage_update(update)
+                self.tick_stage_update(update)
             }
             UpdateAgentState::UpdateStaged(release) => {
                 let update = release.clone();
-                self.try_finalize_update(update)
+                self.tick_finalize_update(update)
             }
             UpdateAgentState::UpdateFinalized(release) => {
                 let update = release.clone();
-                self.end(update)
+                self.tick_end(update)
             }
             UpdateAgentState::EndState => self.nop(),
         };
@@ -123,7 +124,9 @@ impl UpdateAgent {
         let default_delay = Duration::from_secs(super::DEFAULT_REFRESH_PERIOD_SECS);
 
         match self.state {
-            UpdateAgentState::Steady => self.steady_interval,
+            UpdateAgentState::ReportedSteady | UpdateAgentState::NoNewUpdate => {
+                self.steady_interval
+            }
             _ => default_delay,
         }
     }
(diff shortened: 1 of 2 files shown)


src/update_agent/actor.rs
```diff
@@ -135,7 +145,7 @@ impl UpdateAgent {
     }
 
     /// Initialize the update agent.
-    fn initialize(&mut self) -> ResponseActFuture<Self, Result<(), ()>> {
+    fn tick_initialize(&mut self) -> ResponseActFuture<Self, Result<(), ()>> {
```
jlebon: Why the naming changes? Is it to represent that these functions are directly called by the tick handler?

src/update_agent/mod.rs
```diff
@@ -35,8 +31,10 @@ enum UpdateAgentState {
     StartState,
     /// Agent initialized.
     Initialized,
-    /// Agent ready to check for updates.
-    Steady,
+    /// Node steady, agent allowed to check for updates.
+    ReportedSteady,
+    /// No further updates available yet.
+    NoNewUpdate,
```
jlebon: Bikeshed: WDYT about naming the `NoNewUpdate` one with the steady keyword since it's likely going to be the stable/steady state in which the agent will mostly sit in? And maybe instead of `ReportedSteady`, something like `ReadyToCheck`? Makes it a little more obvious too that it hasn't checked for updates yet.

src/update_agent/mod.rs
```diff
@@ -57,18 +55,20 @@ impl Default for UpdateAgentState {
 
 impl UpdateAgentState {
     /// Return the discriminant for current state
+    #[allow(dead_code)]
     fn discriminant(&self) -> u8 {
```
jlebon: Do we actually need this function? The only place that was using it is now gone.

jlebon:
Rationale overall makes sense to me otherwise!

src/update_agent/actor.rs
lucab: Yes (but mostly: a drive-by change to make code easier to skim/grep).

src/update_agent/mod.rs
lucab: Unfortunately I'm a dummy so I have already coupled the meaning of "steady" with https://github.com/coreos/zincati/blob/6bad305124522efb494144d107f2b8c18a39bc79/src/fleet_lock/mod.rs#L25

So, to avoid confusion I'd rather stick with the definition of "steady" as "we released the lock, marking the current release green".

src/update_agent/mod.rs
lucab: Yes, dropped. I was planning to use it for checking the "from" state on transition, but I realized I cannot do it that way.

lucab:
Rebased.

src/update_agent/mod.rs
jlebon: Hmm, is https://github.com/coreos/airlock/pull/1 still open for change? Would it make sense to change the specs here so that we only actually release the lock once we've also been able to check for an update? It intuitively makes sense to me to do that, but I might be missing something.

lucab: > Hmm, is coreos/airlock#1 still open for change?

Not really anymore.

> Would it make sense to change the specs here so that we only actually release the lock once we've also been able to check for an update?

I don't think so. We don't want to couple the "did the node reboot" question with the "is Cincinnati-service ok" question inside Zincati. Plus, users can enforce stronger pre-conditions via https://github.com/coreos/zincati/blob/master/dist/systemd/system/zincati.service#L9-L10

jlebon: Ahh OK right. I'd like to chat about this some more, but I don't think it blocks this PR as is.

lucab: Forwarded to https://github.com/coreos/zincati/issues/239.

src/update_agent/mod.rs
lucab: Self-note: one approach to explore here is to statically enforce allowed transitions via generics, like https://hoverbear.org/blog/rust-state-machine-pattern/#generically-sophistication.

Issue #74, opened 2019-07-17
metrics: stabilize metrics endpoint


lucab:
Zincati currently tracks metrics and exposes them over a Unix-domain socket at `/run/zincati/private/metrics.promsock`.
The endpoint is currently private and the protocol is a provisional one, as expected by [Prometheus local_exporter](https://github.com/lucab/local_exporter).

Once the protocol is stabilized, the socket should be moved to `public/` and permissiond ajusted.

Blockers:
 - [x] https://github.com/lucab/local_exporter/issues/15

PR #243, opened 2020-03-06
metrics: stabilize metrics endpoint

lucab:
This stabilize the metrics endpoint under its new public path,
`/run/zincati/public/metrics.promsock`.
Previous private path is kept as a symlink for compatibility.

Closes: https://github.com/coreos/zincati/issues/74


lucab:
@jlebon my plan is to wait for this (in the next release) to reach `stable`, then drop it.

PR #247, opened 2020-03-10
Release 0.0.9

lucab:
Changes:
 * zincati: rework to use std::future
 * zincati: add completion timeout to HTTP requests
 * metrics: stabilize metrics endpoint
 * update-agent: rework steady/polling state
 * docs: add state-machine diagram
 * cargo: relax dependencies micro versions
 * cargo: update all dependencies

Tracker: https://github.com/coreos/zincati/milestone/6

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2481,7 +2481,7 @@ dependencies = [
 
 [[package]]
 name = "zincati"
-version = "0.0.9"
+version = "0.0.10-alpha.0"
 dependencies = [
  "actix",
  "cfg-if",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2481,7 +2481,7 @@ dependencies = [
 
 [[package]]
 name = "zincati"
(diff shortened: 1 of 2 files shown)


jlebon:
Related, did you see https://bugzilla.redhat.com/show_bug.cgi?id=1800069 BTW? I think we'll need to sort that out before we can get a release into FCOS.

lucab:
I didn't see that BZ, but Igor was asking for a new release in order to unblock dependencies in Fedora. My understanding is that this should help lowering the pressure, but I don't have a good view/understanding of the whole picture.

/cc @rfairley @ignatenkobrain

ignatenkobrain:
Does this already include pre-release version of actix?

lucab:
@ignatenkobrain not yet in here, as 1) they didn't move out of alpha yet 2) I quickly tried a manual bump, but the API break affects this and code won't build without some further massaging.

PR #248, opened 2020-03-12
ci: hook up to CoreOS CI

jlebon:
Add a build & kola test to the repo. Notably, we also run the upgrade
tests since zincati is a big part of this.


jlebon:
Hmm actually, we should probably just enable upgrade testing as part of `fcosBuild` so it applies across all the repos.

jlebon:
OK, this now requires: https://github.com/coreos/coreos-ci-lib/pull/14

jlebon:
Gonna hold this one until https://github.com/coreos/coreos-ci-lib/pull/14 is in so we can make sure it works here!

PR #251, opened 2020-03-24
systemd: Activative via zincati.timer, not by default

cgwalters:
Fedora CoreOS bootup is pretty slow relative to other
distributions out there.  I use `cosa run` a *lot*.  Part
of starting faster is simply doing less.

Starting rpm-ostree today unfortunately is a bit heavyweight.
And zincati is its own service with a binary of nontrivial size.

Since auto-updates are going to be slowly rolled out anyways on
the server side, let's hook zincati startup to an asynchronous
timer run after normal bootup.

In particular, starting zincati/rpm-ostree is no longer *before*
the login prompt with `cosa run`.

But by using a timer with an accuracy slack, we should also
avoid "thundering herd" problems when a lot of machines
boot up at once after e.g. a power failure and go to check
in with the server.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ all:
 install:
 	install -D -t ${DESTDIR}/usr/libexec target/${PROFILE}/zincati
 	install -D -m 644 -t ${DESTDIR}/usr/lib/zincati/config.d dist/config.d/*.toml
-	install -D -m 644 -t ${DESTDIR}/usr/lib/systemd/system dist/systemd/system/*.service
+	install -D -m 644 -t ${DESTDIR}/usr/lib/systemd/system dist/systemd/system/*.{service,timer}
 	install -D -m 644 -t ${DESTDIR}/usr/lib/sysusers.d dist/sysusers.d/*.conf
 	install -D -m 644 -t ${DESTDIR}/usr/lib/tmpfiles.d dist/tmpfiles.d/*.conf
 	install -D -m 644 -t ${DESTDIR}/usr/share/polkit-1/rules.d dist/polkit-1/rules.d/*.rules
--- a/dist/systemd/system/zincati.service
+++ b/dist/systemd/system/zincati.service
@@ -16,6 +16,3 @@ Environment=ZINCATI_VERBOSITY="-v"
 ExecStart=/usr/libexec/zincati agent ${ZINCATI_VERBOSITY}
 Restart=on-failure
 RestartSec=10s
-
-[Install]
-WantedBy=multi-user.target
--- a/dist/systemd/system/zincati.timer
+++ b/dist/systemd/system/zincati.timer
@@ -1,0 +1,8 @@
+[Timer]
+# Spreading things across 30 minutes should help
+# to avoid thundering herds by default.
+OnBootSec=1h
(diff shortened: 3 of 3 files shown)


cgwalters:
This also needs this patch to fedora-coreos-config:
```
diff --git a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
index 89a4bc8..af80635 100644
--- a/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
+++ b/overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset
@@ -13,3 +13,4 @@ enable afterburn-firstboot-checkin.service
 enable afterburn-sshkeys@.service
 # Update agent
 enable zincati.service
+enable zincati.timer
```

dustymabe:
I'm not opposed to starting this as a timer though I have a few comments:

- If we start on boot we should be able to catch an issue in CI if the service doesn't start for whatever reason
    - of course we can add a test that explicitly starts it
- IIUC zincati today already has a level of random time checking so maybe the thundering herd problem is already handled.

cc @lucab - he should be back next week to be able to take a look

lucab:
(I think the "Activative" in the commit title is a typo)

@dustymabe yes, your points are all correct.

My feedback here is more or less of a NACK for this specific path, and the rationale is mirrored at https://github.com/coreos/fedora-coreos-tracker/issues/147#issuecomment-606526951. On top of that, zincati also reports back to the lock-managing service when an update completed successfully, so we are actually interested in **not** having arbitrary large delays there.

However, I'm interested to get more context here to see if/what can be improved.
@cgwalters if I got your case right, you are trying to micro-optimize "RAM/CPU consumption at startup for devel images". Systemd should be already starting things in parallel, so I don't think that Zincati startup directly delays getty console, right?

For devel images we already disabled most of Zincati logic with https://github.com/coreos/fedora-coreos-config/pull/112, but we keep it around for metrics and overall sanity check.

If we still have a resource optimization needs at this point, we could consider:
 * adding a generic delaying unit on devel images to push `boot-complete.target` a bit forward
 * completely disabling zincati on devel images

cgwalters:
> completely disabling zincati on devel images

We effectively do because there aren't any offered updates for testing-devel (AIUI).  But yeah, I guess the alternative is to change e.g. `cosa run` to disable it via Ignition.

dustymabe:
> I guess the alternative is to change e.g. `cosa run` to disable it via Ignition.

this might be relevant: https://github.com/coreos/fedora-coreos-tracker/issues/392

cgwalters:
A core tension I struggle with is that for FCOS "automatic updates" are such a pillar of what we do, but doing development I just so often `cosa run` on devel images that won't have auto-updates.  This also relates to how our tests (except the upgrade tests) disable it.

Probably the right thing is to change `cosa run` to disable zincati by default.  I'll do that instead.

lucab:
FWIW, I also use `cosa run` to manually check Zincati PRs by injecting newer logic into existing releases (e.g. https://github.com/coreos/zincati/pull/286#issuecomment-629094623).

PR #259, opened 2020-04-03
config: update FCOS Cincinnati endpoint (prod)

lucab:
This updates the default FCOS Cincinnati endpoint, switching from
staging infrastructure to production one.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/347


dustymabe:
I probably should have asked [this](https://github.com/coreos/fedora-coreos-streams/pull/78#issuecomment-608408302) question here in this issue.

lucab:
@dustymabe yes, they are serving the same content with the same software+config (later on `stg` may diverge a bit for). Nodes will automatically jump to the new endpoint as soon as this PR hits FCOS. Yay, config fragments!

PR #276, opened 2020-04-24
Expose dead-end release information to users

### LINKED ISSUE #90: cincinnati: expose dead-end release information

sohankunkerkar:
Fixes #90

### DIFF
--- a/dist/tmpfiles.d/zincati.conf
+++ b/dist/tmpfiles.d/zincati.conf
@@ -12,3 +12,10 @@ d     /run/zincati/public    0775 zincati zincati  -   -
 
 # Legacy symlink to metrics socket
 L+    /run/zincati/private/metrics.promsock  - - - - ../public/metrics.promsock
+
+# Runtime MOTD messages which Zincati can write to
+d     /run/zincati/public/motd.d  0775 zincati zincati  -   -
+
+# Symlink to deadend MOTD message, and stub at destination
+f+    /run/zincati/public/motd.d/deadend.motd   0664 zincati zincati -
+L+    /run/motd.d/85-zincati-deadend.motd   - - - - ../zincati/public/motd.d/deadend.motd
--- a/src/cincinnati/mod.rs
+++ b/src/cincinnati/mod.rs
@@ -10,12 +10,16 @@ mod mock_tests;
 use crate::config::inputs;
 use crate::identity::Identity;
 use crate::rpm_ostree::Release;
-use failure::{bail, Fallible};
+use failure::{bail, Fallible, ResultExt};
 use futures::prelude::*;
 use futures::TryFutureExt;
 use prometheus::{IntCounter, IntCounterVec, IntGauge};
 use serde::Serialize;
 use std::collections::BTreeSet;
+use std::fs;
+use std::fs::Permissions;
+use std::io::Write;
+use std::os::unix::fs::PermissionsExt;
 use std::pin::Pin;
 
 /// Metadata key for payload scheme.
@@ -130,6 +134,56 @@ impl Cincinnati {
     }
 }
 
+/// Write MOTD indicating a dead-end release, with the passed `reason`.
+/// If `reason` is `None`, the MOTD written is left as an empty file.
+fn refresh_deadend_motd(reason: Option<String>) -> Fallible<()> {
+    // Avoid showing partially-written messages using tempfile and
+    // persist (rename).
+    let mut f = tempfile::Builder::new()
+        .prefix(".deadend.")
+        .suffix(".motd.partial")
+        // Create the tempfile in the same directory as the final MOTD,
+        // to ensure proper SELinux labels are applied to the tempfile
+        // before renaming.
+        .tempfile_in("/run/zincati/public/motd.d")
+        .with_context(|e| format!("failed to create temporary MOTD file: {}", e))?;
+    // Set correct permissions of the temporary file, before moving to
+    // the destination (`tempfile` creates files with mode 0600).
+    fs::set_permissions(f.path(), Permissions::from_mode(0o664))
+        .with_context(|e| format!("failed to set permissions of temporary MOTD file: {}", e))?;
+
+    if let Some(reason) = reason {
+        writeln!(
+            f,
+            "This release is a dead-end and won't auto-update: {}",
+            reason
+        )
+        .with_context(|e| format!("failed to write MOTD: {}", e))?;
+    }
+
+    f.persist("/run/zincati/public/motd.d/deadend.motd")
+        .with_context(|e| format!("failed to persist temporary MOTD file: {}", e))?;
+    Ok(())
+}
+
+/// Evaluate and record whether booted OS is a dead-end release, and
+/// log that information in a MOTD file.
+fn refresh_deadend_status(node: &Node) {
+    let deadend_reason = evaluate_deadend(node);
+    match &deadend_reason {
+        Some(reason) => {
+            log::info!("dead-end release detected: {}", reason);
+            BOOTED_DEADEND.set(1);
+        }
+        None => {
+            BOOTED_DEADEND.set(0);
+        }
+    };
+    if let Err(e) = refresh_deadend_motd(deadend_reason) {
+        log::warn!("failed to update dead-end release MOTD: {}", e);
+    }
+}
+
 /// Walk the graph, looking for an update reachable from the given digest.
 fn find_update(
     graph: client::Graph,
@@ -159,14 +213,7 @@ fn find_update(
     let cur_release = Release::from_cincinnati(cur_node.clone())
         .map_err(|e| CincinnatiError::FailedNodeParsing(e.to_string()))?;
 
-    // Evaluate and record whether booted OS is a dead-end release.
-    // TODO(lucab): consider exposing this information in more places
-    // (e.g. logs, motd, env/json file in a well-known location).
-    let is_deadend = match evaluate_deadend(&cur_node) {
-        Some(_) => 1,
-        None => 0,
-    };
-    BOOTED_DEADEND.set(is_deadend);
+    refresh_deadend_status(&cur_node);
 
     // Try to find all local deployments in the graph too.
     let local_releases = find_local_releases(&graph, local_depls);
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ reqwest = { version = "^0.10", features = ["json"] }
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = "^1.0"
 structopt = "^0.3"
+tempfile = "^3.1"
 tokio = "^0.2"
 toml = "^0.5"
 url_serde = "^0.2"


src/cincinnati/mod.rs
```diff
@@ -277,7 +277,13 @@ fn evaluate_deadend(node: &Node) -> Option<String> {
     if deadend_reason.is_empty() {
         deadend_reason = "(unknown reason)".to_string();
     }
-
+    if !deadend_reason.is_empty() {
```
jlebon, line 280: This can never be true, right? We set it to a default value just before this.

src/cincinnati/mod.rs
```diff
@@ -277,7 +277,13 @@ fn evaluate_deadend(node: &Node) -> Option<String> {
...
+        let _ = std::fs::remove_file("/etc/motd");
```
jlebon, line 281: I think we want to write to e.g. `motd.d/zincati.motd` instead, no?

lucab: Indeed, this should plug into https://github.com/coreos/console-login-helper-messages.

src/cincinnati/mod.rs
lucab: Almost, this is a tautological check (condition is always true).
The root problem here is that this logic is in the wrong place; it should be pushed one level up, handling the `Option<String>`.

src/cincinnati/mod.rs
```diff
@@ -277,7 +277,13 @@ fn evaluate_deadend(node: &Node) -> Option<String> {
...
+        let data = format!("version: {} is a dead-end release", node.version);
```
lucab, line 282: The version here is not very interesting. The important bits are:
 * is this node running on a deadend release (boolean, required)
 * if so, what's the reason/detail for this deadend release (string, optional)

lucab:
@sohankunkerkar did you try to manually test this change? I have a feeling that security policy/permissions won't allow the `zincati` user to write the file as your PR is doing.

@rfairley this writes/updates a MOTD fragment every once in a while. What is the intended mechanism to refresh the overall banner?

@jlebon in order to make this behavior a bit more atomic, should we split this logic into a file-writing step followed by a symlink/move operations?

jlebon:
> @jlebon in order to make this behavior a bit more atomic, should we split this logic into a file-writing step followed by a symlink/move operations?

+1 It's just good practice in general. We use `tempfile` in coreos-installer to make that easier. Though how essential it is I guess depends on how the MOTD banner refresh actually happens.

rfairley:
> @rfairley this writes/updates a MOTD fragment every once in a while. What is the intended mechanism to refresh the overall banner?

The latest updated fragments will be read upon SSH login, if writing under `/run/motd.d` (upon login, `sshd` in Fedora by default will invoke `pam_motd`, which will read and display fragments from `/run/motd.d` with overrides/lexicographic order). For more, see https://www.mankier.com/8/pam_motd.

Note that `console-login-helper-messages` is not involved when just writing to `/run/motd.d/`, it would only be involved if writing to `/run/console-login-helper-messages/motd.d`. In the latter case, `console-login-helper-messages` will end up writing a consolidated snippet to display which is the result of appending the the content of all snippets that were found under the `console-login-helper-messages/motd.d` directories (this snippet is currently only updated during boot - it will soon be updated upon a file being dropped into `/run/console-login-helper-messages/motd.d` by use of `PathChanged=` with https://github.com/coreos/console-login-helper-messages/pull/43). The reason for checking `console-login-helper-messages/motd.d` is to support systems where the PAM version does not yet have `/run/motd.d` support. Fedora 31+ has a new enough version of PAM - so `/run/motd.d/` may be used here.

If we want to display this same information after logging in by other means than SSH, e.g. login through serial console with `agetty --autologin`, `/etc/motd` would need to be used for the time being **or** we could configure `/run/motd.d/zincati.motd` to also be displayed through `/etc/login.defs` (see https://github.com/coreos/fedora-coreos-config/pull/371#issuecomment-622037955).

rfairley:
:+1: to `/run/motd.d/zincati.motd` or so here. We may also want a numeric prefix in the filename (e.g. `90-zincati.motd`) to explicitly suggest ordering w.r.t. fragments from elsewhere that may be placed in `/run/motd.d`.

rfairley:
> @sohankunkerkar did you try to manually test this change? I have a feeling that security policy/permissions won't allow the zincati user to write the file as your PR is doing.

:+1: - worth checking if there is an issue with the `zincati` user writing out the fragment (as a reference, `console-login-helper-messages` runs as `init_t` in SELinux and as root user, and writes in `/run/motd.d` without issue - have not tried writing snippets from units run as a sysuser).

> > @jlebon in order to make this behavior a bit more atomic, should we split this logic into a file-writing step followed by a symlink/move operations?
> 
> +1 It's just good practice in general. We use `tempfile` in coreos-installer to make that easier. Though how essential it is I guess depends on how the MOTD banner refresh actually happens.

:+1: - given the refresh will happen upon new SSH login, this shouldn't be too much of a concern - but +1 for the tmpfile/symlink pattern if it can be added in here.

lucab:
@sohankunkerkar bump, do you have this PR still on your radar?

sohankunkerkar:
> @sohankunkerkar bump, do you have this PR still on your radar?

I will be working on this PR in the upcoming sprint. Sorry for the latency here.

rfairley:
> > @sohankunkerkar did you try to manually test this change? I have a feeling that security policy/permissions won't allow the zincati user to write the file as your PR is doing.
> 
> +1 - worth checking if there is an issue with the `zincati` user writing out the fragment (as a reference, `console-login-helper-messages` runs as `init_t` in SELinux and as root user, and writes in `/run/motd.d` without issue - have not tried writing snippets from units run as a sysuser).
> 

Briefly verified now that with a test script that writes a MOTD to `/run/motd.d/zincati.motd`, running the script from `zincati.service` as the `zincati` sysuser gives `Permission denied`, but running `zincati.service` as root, the MOTD file gets written. This makes sense as `/run/motd.d` is owned by root.

Discussed with Luca - and a sane approach to allow the `zincati` sysuser to display a MOTD would probably be writing to somewhere under the zincati-owned directory `/run/zincati`, and installing a symlink `/run/motd.d/zincati.motd -> /run/zincati/public/zincati.motd` or so when packaging the RPM.

Trouble with this is `sshd` is not reading the `/run/zincati/public/zincati.motd` file at the symlink target, due to SELinux denying a read outside of the `/run/motd.d` directory:

```
[35953.812783] audit: type=1400 audit(1593535777.088:985): avc:  denied  { read } for  pid=3299 comm="sshd" name="zincati.motd" dev="tmpfs" ino=69361 scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:var_run_t:s0 tclass=file permissive=1
[35953.816450] audit: type=1400 audit(1593535777.088:985): avc:  denied  { open } for  pid=3299 comm="sshd" path="/run/zincati/zincati.motd" dev="tmpfs" ino=69361 scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:var_run_t:s0 tclass=file permissive=1
[35953.820596] audit: type=1400 audit(1593535777.088:986): avc:  denied  { getattr } for  pid=3299 comm="sshd" path="/run/zincati/zincati.motd" dev="tmpfs" ino=69361 scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:var_run_t:s0 tclass=file permissive=1
```

I'm not sure of a way around this other than 1) providing some means of allowing certain sysusers to write directly into `/run/motd.d` (i.e. a group for writers to motd.d) or 2) adding an exception for `/run/zincati/public/zincati.motd` in the Fedora SELinux policy.

Otherwise, a quick way forward I see is installing a symlink `/run/console-login-helper-messages/motd.d/zincati.motd -> /run/zincati/public/zincati.motd` (which would be done in the Zincati Makefile/package). Then, `console-login-helper-messages-motdgen.service` should be able to follow symlinks to other directories and display the message from `/run/zincati/public/zincati.motd`.

@sohankunkerkar - could you update this to write to /run/zincati/public/zincati.motd (or another filename if @lucab has a preference) instead of /run/motd.d/zincati.motd?

I'll then test to see whether the `console-login-helper-messages` symlink method above works. Happy to hear other thoughts too!

sohankunkerkar:
@rfairley Thanks for the detailed analysis.

> Briefly verified now that with a test script that writes a MOTD to `/run/motd.d/zincati.motd`, running the script from `zincati.service` as the `zincati` sysuser gives `Permission denied`, but running `zincati.service` as root, the MOTD file gets written. This makes sense as `/run/motd.d` is owned by root.

Looks like I was hitting this issue but not sure about the exact resolution there. Thanks for this information. TIL

> Discussed with Luca - and a sane approach to allow the `zincati` sysuser to display a MOTD would probably be writing to somewhere under the zincati-owned directory `/run/zincati`, and installing a symlink `/run/motd.d/zincati.motd -> /run/zincati/public/zincati.motd` or so when packaging the RPM.

I don't have any preference here as such. I'm fine with `/run/zincati/public/zincati.motd`

> Trouble with this (after trying just now) is `sshd` is not reading the `/run/zincati/public/zincati.motd` file at the symlink target, due to SELinux denying a read outside of the `/run/motd.d` directory:
> 
> ```
> [35953.812783] audit: type=1400 audit(1593535777.088:985): avc:  denied  { read } for  pid=3299 comm="sshd" name="zincati.motd" dev="tmpfs" ino=69361 scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:var_run_t:s0 tclass=file permissive=1
> [35953.816450] audit: type=1400 audit(1593535777.088:985): avc:  denied  { open } for  pid=3299 comm="sshd" path="/run/zincati/zincati.motd" dev="tmpfs" ino=69361 scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:var_run_t:s0 tclass=file permissive=1
> [35953.820596] audit: type=1400 audit(1593535777.088:986): avc:  denied  { getattr } for  pid=3299 comm="sshd" path="/run/zincati/zincati.motd" dev="tmpfs" ino=69361 scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:var_run_t:s0 tclass=file permissive=1
> ```
> 
> I'm not sure of a way around this upstream other than 1) providing some means of allowing certain sysusers to write directly into `/run/motd.d` (i.e. a group for writers to motd.d) or 2) adding an exception for `/run/zincati/public/zincati.motd` in the Fedora SELinux policy.
> 
> Otherwise, a quick way forward I see is installing a symlink `/run/console-login-helper-messages/motd.d/zincati.motd -> /run/zincati/public/zincati.motd` (which would be done in the Zincati Makefile/package). Then, `console-login-helper-messages-motdgen.service` should be able to follow symlinks to other directories and display the message from `/run/zincati/public/zincati.motd`. **Edit**: a problem now with this is letting CLHM detect the changes when a new write is done on `/run/zincati/public/zincati.motd` - will need to think about this.
> 
> @sohankunkerkar - could you update this to write to /run/zincati/public/zincati.motd (or another filename if @lucab has a preference) instead of /run/motd.d/zincati.motd?

Sure, I was wondering if you wanna take up this PR and push over the finish line.

> I'll then test to see whether the `console-login-helper-messages` symlink method above works. Happy to hear other thoughts too!

rfairley:
> Otherwise, a quick way forward I see is installing a symlink /run/console-login-helper-messages/motd.d/zincati.motd -> /run/zincati/public/zincati.motd (which would be done in the Zincati Makefile/package). Then, console-login-helper-messages-motdgen.service should be able to follow symlinks to other directories and display the message from /run/zincati/public/zincati.motd. Edit: a problem now with this is letting CLHM detect the changes when a new write is done on /run/zincati/public/zincati.motd - will need to think about this.

Tried this out, and the change detection with `PathChanged=/run/console-login-helper-messages/motd.d` works for a file outside of `/run/console-login-helper-messages/motd.d` pointed to by a symlink from the directory. So writing to `/run/zincati/public/zincati.motd`, and having systemd tmpfiles for a symlink `/run/console-login-helper-messages/motd.d/zincati.motd -> /run/zincati/public/zincati.motd` should work. Will update the PR for this.

lucab:
@rfairley thanks, great followup. A couple of additional doubts while we iron out the remaining details:
 * isn't the long term plan to drop CLHM in favor of plain `motd.d` eventually?
 * what about the state at startup and on non-deadends? How does the system cope with a broken link from `/run/motd.d/zincati.motd` to a non-existing target?
 * based on the above answer, should we use two symlinks? Like:
  - `/run/motd.d/zincati.motd` -> /run/zincati/public/zincati.motd -> `/dev/null` on startup and non-deadends (set up by tmpfiles)
  - `/run/motd.d/zincati.motd` -> /run/zincati/public/zincati.motd -> `/run/zincati/public/<something>` once a deadend is detected
 * what is the selinux context of `/run/motd.d/*`? I have a feeling that a sensible thing to do here could be having a `/run/zincati/public/motd.d/*` with a matching context

rfairley:
> @rfairley thanks, great followup. A couple of additional doubts while we iron out the remaining details:
> 
>     * isn't the long term plan to drop CLHM in favor of plain `motd.d` eventually?
> 

Eventually yes, we'd like any MOTD generators to write snippets in `/run/motd.d` rather than the CLHM directory (opened an issue to track this: https://github.com/coreos/console-login-helper-messages/issues/60). In general, plain `motd.d` should be tried first, and then CLHM `motd.d` as a last resort.

>     * what about the state at startup and on non-deadends? How does the system cope with a broken link from `/run/motd.d/zincati.motd` to a non-existing target?
> 

AIUI `pam_motd` will log any errors encountered in opening the files, but will still continue to try to read all the files in `/run/motd.d`. Agree - we should have something at the destination though, to avoid logging errors in PAM.

>     * based on the above answer, should we use two symlinks? Like:
> 
> 
>     * `/run/motd.d/zincati.motd` -> /run/zincati/public/zincati.motd -> `/dev/null` on startup and non-deadends (set up by tmpfiles)
> 
>     * `/run/motd.d/zincati.motd` -> /run/zincati/public/zincati.motd -> `/run/zincati/public/<something>` once a deadend is detected
> 

This SGTM - I guess that'd mean writing out directly to `/run/zincati/public/<something>`, then having Zincati modify the `/run/zincati/public/zincati.motd` symlink to point to it when a deadend is detected. If not detected, then modify the symlink to `/dev/null`.

I think we could avoid the extra indirection from the symlink by instead having `tmpfiles.d` write an empty file to `/run/zincati/public/zincati.motd`, and write the deadend information out in the way currently done in this PR to `/run/zincati/public/zincati.motd`. Instead of removing the file if deadend is not detected, write an empty string to it. This shouldn't affect the final MOTD message displayed - as long as the file has no characters (e.g. newlines) in it.

> 
>     * what is the selinux context of `/run/motd.d/*`? I have a feeling that a sensible thing to do here could be having a `/run/zincati/public/motd.d/*` with a matching context

`/run/motd.d` is `pam_var_run_t`. Setting the context of `/run/zincati/public/motd.d/*` seems reasonable to me as one way to do it.

Another alternative that comes to mind is adding a `/run/motd.d/zincati.motd` tmpfile entry, and have the tmpfile set its ownership to `zincati` while inheriting the correct SELinux context. This would avoid the symlink and changes in the Fedora SELinux policy. Will see if this works.

lucab:
> I think we could avoid the extra indirection from the symlink by instead having tmpfiles.d write an empty file to /run/zincati/public/zincati.motd, and write the deadend information out in the way currently done in this PR to /run/zincati/public/zincati.motd. Instead of removing the file if deadend is not detected, write an empty string to it. This shouldn't affect the final MOTD message displayed - as long as the file has no characters (e.g. newlines) in it.

This goes back to https://github.com/coreos/zincati/pull/276#issuecomment-624226211. Updating a symlink is an atomic operation, while changing the content of a file is not.

rfairley:
> > I think we could avoid the extra indirection from the symlink by instead having tmpfiles.d write an empty file to /run/zincati/public/zincati.motd, and write the deadend information out in the way currently done in this PR to /run/zincati/public/zincati.motd. Instead of removing the file if deadend is not detected, write an empty string to it. This shouldn't affect the final MOTD message displayed - as long as the file has no characters (e.g. newlines) in it.
> 
> This goes back to [#276 (comment)](https://github.com/coreos/zincati/pull/276#issuecomment-624226211). Updating a symlink is an atomic operation, while changing the content of a file is not.

Right, yes; before I was thinking we'd use `tempfile` eventually here for the writing. Using the symlink as the on/off switch makes sense too to provide the needed atomicity - and looks simpler to implement.

rfairley:
Did a test of this, exercising Zincati, this is how it looks:

zincati-deadend-MOTD-display

Steps to try this out, using coreos-assembler:

1. `cosa init https://github.com/coreos/fedora-coreos-config`. Then install Zincati to the rootfs to override in COSA:
    `cd /path/to/zincati`
    `make && make install DESTDIR=/path/to/fcos/overrides/rootfs/`
2. Override the `selinux-policy` and `selinux-policy-targeted` RPMs with the patch in https://github.com/fedora-selinux/selinux-policy/pull/398 applied, by dropping these RPMs into `/path/to/fcos/overrides/rpm/`. RPMs with the necessary patch can be downloaded from https://download.copr.fedorainfracloud.org/results/rfairley/selinux-policy/fedora-32-x86_64/01554192-selinux-policy/ (which I built in COPR).
3. Apply the following change to the config git in `/path/to/fcos/src/config` (request httpd to run a mock Cincinnati server off that):
    ```diff
    diff --git a/manifests/fedora-coreos.yaml b/manifests/fedora-coreos.yaml
    index c64e2dd..ec6ec60 100644
    --- a/manifests/fedora-coreos.yaml
    +++ b/manifests/fedora-coreos.yaml
    @@ -16,6 +16,7 @@ packages:
    - fedora-coreos-pinger
    # Updates
    - zincati
    +  - httpd
    
    etc-group-members:
    # Add the docker group to /etc/group
    ```
4. Run `cosa build`.
5. Once complete, do `cat builds/latest/x86_64/meta.json | grep -E "ostree-commit|ostree-version"`
6. Create the following FCC `deadend.fcc`, replacing the "payload" with the "ostree-commit" from the command above and "version" with "ostree-version" above, in the Cincinnati graph.
    ```yaml
    variant: fcos
    version: 1.0.0
    systemd:
      units:
        - name: zincati.service
          dropins:
            - name: verbose.conf
              contents: "[Service]\nEnvironment=ZINCATI_VERBOSITY=\"-vvvv\""
        - name: httpd.service
          enabled: true
    storage:
      files:
        - path: /etc/zincati/config.d/99-cincinnati-baseurl.toml
          contents:
            inline: |
              [cincinnati]
              base_url = "http://127.0.0.1:8080"
        - path: /etc/zincati/config.d/99-enable-updates.toml
          contents:
            inline: |
              updates.enabled = true
        - path: /var/www/html/v1/graph
          contents:
            inline: |
              {
                "nodes": [
                  {
                    "version": "32.20200715.dev.5",
                    "metadata": {
                      "org.fedoraproject.coreos.scheme": "checksum",
                      "org.fedoraproject.coreos.releases.age_index": "0",
                      "org.fedoraproject.coreos.updates.deadend_reason": "https://github.com/coreos/fedora-coreos-tracker/issues/215",
                      "org.fedoraproject.coreos.updates.deadend": "true"
                    },
                    "payload": "39a44ed16a26f588ee11a7381d239c8b8bf5b00e6893847823aa6a5041300357"
                  }
                ],
                "edges": []
              }
        - path: /etc/httpd/conf.d/port.conf
          contents:
            inline: |
              Listen 8080
    ```
7. Convert the FCC above to an Ignition config `podman run -i --rm quay.io/coreos/fcct:release --pretty --strict < deadend.fcc > deadend.ign`
8. Boot the image with `cosa run -i deadend.fcc`. The MOTD message as in the screenshot above should show.

rfairley:
Requires: https://github.com/fedora-selinux/selinux-policy/pull/398

dist/tmpfiles.d/zincati.conf
```diff
@@ -12,3 +12,9 @@ d     /run/zincati/public    0775 zincati zincati  -   -
 
 # Legacy symlink to metrics socket
 L+    /run/zincati/private/metrics.promsock  - - - - ../public/metrics.promsock
+
+# Runtime MOTD messages which Zincati can write to
+d     /run/zincati/public/motd.d  0775 zincati zincati  -   -
+
+# Symlink to deadend MOTD message
+L+    /run/motd.d/85-zincati-deadend.motd  - - - - ../zincati/public/motd.d/deadend.motd
```
lucab, line 20: The target here won't exist until the first successful graph fetch (and possibly never, in case of failures or if Zincati is disabled).

Should we start with a `/run/motd.d/85-zincati-deadend.motd -> /dev/null` and then update the symlink on each graph refresh maybe?

src/cincinnati/mod.rs
```diff
@@ -130,6 +132,59 @@ impl Cincinnati {
...
+            return;
+        }
+    };
+
+    if let Some(reason) = reason {
+        if let Err(e) = writeln!(
+            f,
+            "{} this release is a dead-end and won't update: {}",
+            // Fixed color 11 - a brighter shade of yellow.
+            Fixed(11).paint("Notice:"),
```
lucab, line 160: I'd stick this to plain ASCII without fancy colors.

src/cincinnati/mod.rs
```diff
@@ -130,6 +132,59 @@ impl Cincinnati {
...
+            reason
+        ) {
+            log::warn!("failed to write dead-end release MOTD: {}", e);
+            return;
+        }
+    }
+
+    if let Err(e) = f.persist("/run/zincati/public/motd.d/deadend.motd") {
+        log::warn!("failed to persist dead-end release MOTD: {}", e);
```
lucab, line 169: Here and in similar places above: perhaps we should just add context to errors, `?` them, and then log in a single place at the calling site.

src/cincinnati/mod.rs
```diff
@@ -130,6 +132,59 @@ impl Cincinnati {
     }
 }
 
+/// Write MOTD indicating a dead-end release, with the passed `reason`.
+/// If `reason` is `None`, the MOTD written is left as an empty file.
+fn refresh_deadend_motd(reason: Option<String>) {
+    // Avoid showing partially-written messages using tempfile and
+    // persist (rename).
+    let f = tempfile::Builder::new()
+        .prefix("deadend")
```
lucab, line 141: Nit: just for safety, I'd make this a dotfile.

lucab:
This looks amazingly great! I've left a few comments in review.

dist/tmpfiles.d/zincati.conf
rfairley: I had missed thinking about the initial state of the symlink (again) - yes, we should avoid the dangling symlink. Linking to `/dev/null` would be good, however, runs into a problem with the `zincati` user getting a `Permission denied` when removing and re-creating the symlink with the new location. AFAIK, there isn't a way to overwrite the symlink target without deleting/recreating the symlink file (until https://lore.kernel.org/linux-fsdevel/cover.1524549513.git.osandov@fb.com/ gets accepted).

As an alternative, could we create the destination file upfront via tmpfiles.d, initializing it as an empty file? I updated the tmpfiles.d config for this.

src/cincinnati/mod.rs
rfairley: Updated! Thought I'd try to make it stand out, but it is simpler to keep it plain characters.

src/cincinnati/mod.rs
rfairley: Done! `failure`'s `with_context` and using `Fallible` seemed most concise to chain the error, but let me know if there is a preferred way to do this.

src/cincinnati/mod.rs
rfairley: Done. Also added `.` on either side to make the tmpfile name more readable (`.deadend.abcdef.motd.partial` instead of `.deadendabcdefmotd.partial`) (not that it matters too much).

rfairley:
Also added an info-level log message for when dead-end is detected, in passing.

rfairley:
Updated - interdiff from last review: https://github.com/coreos/zincati/compare/41400a404e6e7fc996f8710f01c114fdacc090f4..ffc56f0697c5196dc6fa3770ae4e29bcb6af0b5c

dist/tmpfiles.d/zincati.conf
lucab: Ah, I forgot that we cannot atomically update an existing symlink. Yes, the tmpfiles way seems ok.

dist/tmpfiles.d/zincati.conf
```diff
@@ -12,3 +12,10 @@ d     /run/zincati/public    0775 zincati zincati  -   -
 
 # Legacy symlink to metrics socket
 L+    /run/zincati/private/metrics.promsock  - - - - ../public/metrics.promsock
+
+# Runtime MOTD messages which Zincati can write to
+d     /run/zincati/public/motd.d  0755 zincati zincati  -   -
```
lucab, line 17: I think this was `0775` in the previous iteration. What's the reason for changing it?

src/cincinnati/mod.rs
```diff
@@ -130,6 +131,52 @@ impl Cincinnati {
...
+fn refresh_deadend_motd(reason: Option<String>) -> Fallible<()> {
+    // Avoid showing partially-written messages using tempfile and
+    // persist (rename).
+    let mut f = tempfile::Builder::new()
+        .prefix(".deadend.")
+        .suffix(".motd.partial")
+        // Create the tempfile in the same directory as the final MOTD,
+        // to ensure proper SELinux labels are applied to the tempfile
+        // before renaming.
+        .tempfile_in("/run/zincati/public/motd.d")
```
lucab, line 145: Just double-checking: is this one created with `0664 - zincati/zincati` permission too?

dist/tmpfiles.d/zincati.conf
rfairley: Sorry, had been experimenting (seeing if `0664` worked), and typed the wrong value when fixing it back. Thanks for catching, fixed!

src/cincinnati/mod.rs
rfairley: tempfile creates the file with `0600` - had seen this but didn't think too much of it. Having the tmpfiles.d give it `0664` only to be replaced by a file with mode `0600` is confusing though - I updated this to set the mode of the tempfile to `0664`.

rfairley:
Updated - interdiff: https://github.com/coreos/zincati/compare/ffc56f0697c5196dc6fa3770ae4e29bcb6af0b5c..d45739231ece7d6b0d7a73a37c19598def803a72

src/cincinnati/mod.rs
lucab: Ack. Upstream library has this one open since some time: https://github.com/Stebalien/tempfile/issues/114.

lucab:
Code LGTM, thanks a lot! I'm holding this one for a bit while we wait for feedback on the SELinux proposed change.

sohankunkerkar:
Closing this PR as we're going ahead with https://github.com/coreos/zincati/pull/367

PR #286, opened 2020-05-14
update-agent: react to persistent deploy failure

lucab:
This augments update-agent state-machine to prevent locking into a faulty
`UpdateAvailable` state.
In particular, rpm-ostree may not agree to deploy the currently
detected update target. After a finite amount of failed attempts,
Zincati now abandons its update target and goes back to the
`NoNewUpdate` state, polling Cincinnati again for a new target.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/481
Closes: https://github.com/coreos/zincati/issues/288

---

New state-machine diagram:
diagram


bgilbert:
I haven't attempted to review the code in any detail, but the change SGTM.

src/update_agent/mod.rs
```diff
@@ -117,6 +117,51 @@ impl UpdateAgentState {
         self.transition_to(target);
     }
 
+    /// Record a deploy attempt in UpdateAvailable state.
+    ///
+    /// In case of persistent failure, this returns the update version which
+    /// is being abandoned.
+    fn deploy_attempted(&mut self, success: bool) -> Option<Release> {
+        // Maximum failed deploy attempts before declaring a persistent error.
+        const MAX_DEPLOY_ATTEMPTS: u8 = 60;
```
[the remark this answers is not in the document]
lucab: Good catch, I botched this. It was supposed to be 60 minutes. Retrying every 5 minutes, the watermark here should be at `12`.

This cap is a bit arbitrary, anything strictly larger than 1 would indeed work.
A single attempt instead would remove the re-entering edge. With the current ticking logic, it could result in an infinite fast-spinning loop where the agent is oscillating between `NoNewUpdate` and `UpdateaAvailable` as fast as it can.

lucab:
Ok, I manually tested this on top of the current `testing` rollout with a fine-tuned rollout wariness, and the agent is indeed able to recover and select the new edge once it appears:
```
[INFO ] starting update agent (zincati 0.0.11-alpha.0)
...
[INFO ] initialization complete, auto-updates logic enabled
[INFO ] target release '31.20200505.2.0' selected, proceeding to stage it
[ERROR] failed to stage deployment: rpm-ostree deploy failed: 
    error: Upgrade target revision 'a4395eae3e1844d806a79b5cf51d44e60c96c7ab261a715f5d7fd89584c6963b' with timestamp 'Wed May  6 16:42:59 2020' is chronologically older than current ...
[INFO ] target release '31.20200505.2.0' selected, proceeding to stage it
[ERROR] failed to stage deployment: rpm-ostree deploy failed:
    error: Upgrade target revision 'a4395eae3e1844d806a79b5cf51d44e60c96c7ab261a715f5d7fd89584c6963b' with timestamp 'Wed May  6 16:42:59 2020' is chronologically older than current ...
...
[WARN ] persistent deploy failure detected, target release '31.20200505.2.0' abandoned
[INFO ] target release '31.20200505.2.1' selected, proceeding to stage it
...
```

src/update_agent/actor.rs
```diff
@@ -120,15 +119,21 @@ impl UpdateAgent {
     /// This influences the pace of the update-agent refresh loop. Timing of the
     /// state machine is not uniform. Some states benefit from more/less
     /// frequent refreshes, or can be customized by the user.
-    fn refresh_delay(&self) -> Duration {
-        let default_delay = Duration::from_secs(super::DEFAULT_REFRESH_PERIOD_SECS);
+    fn refresh_delay(&self, prev_state: UpdateAgentState) -> Option<Duration> {
+        use std::mem::discriminant;
+
+        // State changes trigger immediate tick/action.
+        if discriminant(&prev_state) != discriminant(&self.state) {
```
jlebon, line 126: Huh neat, hadn't seen `discriminant()` before.

src/update_agent/mod.rs
```diff
@@ -36,7 +36,7 @@ enum UpdateAgentState {
     /// No further updates available yet.
     NoNewUpdate,
     /// Update available from Cincinnati.
```
jlebon, line 38: Would be good to describe here what the `u8` is for.

src/update_agent/mod.rs
```diff
@@ -117,6 +117,51 @@ impl UpdateAgentState {
         self.transition_to(target);
     }
 
+    /// Record a deploy attempt in UpdateAvailable state.
+    ///
+    /// In case of persistent failure, this returns the update version which
+    /// is being abandoned.
+    fn deploy_attempted(&mut self, success: bool) -> Option<Release> {
+        // Maximum failed deploy attempts before declaring a persistent error.
+        const MAX_DEPLOY_ATTEMPTS: u8 = 12;
```
jlebon, line 126: Minor: I'd put this at the top to raise visibility.

src/update_agent/actor.rs
```diff
@@ -222,7 +222,16 @@ impl UpdateAgent {
         let can_fetch = self.strategy.can_check_and_fetch(&self.identity);
         let state_change = actix::fut::wrap_future::<_, Self>(can_fetch)
             .then(|can_fetch, actor, _ctx| actor.locked_upgrade(can_fetch, release))
-            .map(|res, actor, _ctx| res.map(|release| actor.state.update_staged(release)));
+            .map(|res, actor, _ctx| {
+                let success = res.is_ok();
+                if let Some(abandoned) = actor.state.deploy_attempted(success) {
```
jlebon, line 227: The API here feels awkward. We pass in whether the upgrade was a success or not and `deploy_attempted` not only changes the state, but also returns whether the update was abandoned?

I think it'd be easier to grok if we kept doing `update_staged()` here in the `is_ok()` case and otherwise pass it to a function called e.g. `actor.state.maybe_abandon()`? (And then log the warning in that function). WDYT? (To be clear, this is fine as is too!)

src/update_agent/mod.rs
```diff
@@ -117,6 +117,51 @@ impl UpdateAgentState {
...
+        let target = UpdateAgentState::UpdateAvailable((update, failed));
+
+        self.transition_to(target);
+    }
+
+    /// Transition to the NoNewUpdate state after persistent deploy failure.
+    fn update_abandoned(&mut self) {
+        let target = UpdateAgentState::NoNewUpdate;
+
+        self.transition_to(target);
```
jlebon, line 162: So just to check, after we go back to `NoNewUpdate` here, do we ignore the same update again, or will we retry it even if that's still the latest update when we recheck? I think we want the latter behaviour, right? That way, e.g. if the OSTree remote is down for 1h or whatever, we'll still retry it again later.

lucab: Yes. Going back to `NoNewUpdate` clears the state, so on the next attempt we may be retrying the same update (that's the latter behavior).

If the graph changed in the meanwhile, best-edge selection will result in picking up any newer target update (that's the usual logic for choosing the freshest target among a set of valid edges).

src/update_agent/actor.rs
lucab: I'm honestly still looking forward for https://github.com/rust-lang/rust/issues/60553 to be implemented and stabilized.

src/update_agent/actor.rs
lucab: I see that it feels a bit strange. I'm trying to keep the state-machine logic split from the actor logic, so that the former is easier to test/inspect on its own.

I reworked this code a bit so that the actor owns the result-branching and thus it is still in charge of calling either `state.update_stage()` or `state.record_failed_deploy()`, how does it look now?

lucab:
Pushed a fixup commit to address the comments.

src/update_agent/actor.rs
jlebon: Yup, much better!

Issue #271, opened 2020-04-15
docs/devel: document required metadata for introspection


lucab:
Zincati currently derives its own [agent identity](https://github.com/coreos/zincati/blob/v0.0.10/src/identity/mod.rs#L32-L49) by introspecting run-time machine state and build-time OS metadata.

This means that it will refuse to run on systems where such introspection fails, e.g. because of some missing metadata.
While this is a design property which is good to keep, it makes the life of system integrators hard as they have to figure out which metadata may be missing in their OS.

There should be a dev docpage in this repository spelling out which metadata is required to be on the system and where it is sourced from.

PR #287, opened 2020-05-15
docs/devel: document required metadata for introspection

lucab:
This records all the details and OS metadata which are required by
Zincati in order to properly derive its own agent identity.

Fixes: https://github.com/coreos/zincati/issues/271

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,4 +1,10 @@
-# Development
+# Contributing
+
+## Project architecture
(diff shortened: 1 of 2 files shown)


lucab:
@gicmo this should be all the metadata needed to match current FCOS builds, but let me know in case I missed anything. Positive feedback appreciated too, as I should probably lift it to FCOS reference docs.

PR #292, opened 2020-05-22
docs/dev: add some details on periodic strategy design

lucab:
This adds some insights, rationale and history on the "periodic"
strategy.

### DIFF
--- a/docs/development/update-strategy-periodic.md
+++ b/docs/development/update-strategy-periodic.md
@@ -1,0 +1,44 @@
+# Periodic update strategy 
+
+The agent supports a `periodic` strategy, which allows gating reboots based on "reboot windows", defined on weekly basis.
+
+This strategy is a port of [locksmith reboot windows][locksmith], with a few differences:
+
+ * all times and dates are UTC-based to avoid skews and ambiguities
+ * multiple disjoint reboot windows are supported
+ * multiple configuration entries are assembled into a single weekly calendar
+ * weekdays need to be specified, in either long or abbreviated form
+ * length duration is always specified in minutes
+
+[locksmith]: https://github.com/coreos/locksmith/tree/v0.6.2#reboot-windows
+
+# Timing and configuration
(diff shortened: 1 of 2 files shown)


docs/development/update-strategy-periodic.md
```diff
@@ -0,0 +1,42 @@
+# Periodic update strategy 
+
+The agent supports a `periodic` strategy, which allows gating reboots based on "reboot windows", defined on weekly basis.
+
+This strategy is a port of [locksmith reboot windows][locksmith], with a few differences:
+
+ * all times and dates are UTC-based to avoid skews and ambiguities
```
ashcrow, line 7: nit: It may be worth noting it also uses the 24 hour clock format (IE: 01:00 == 1 am UTC). Most people will infer this with UTC, but still.

docs/development/update-strategy-periodic.md
```diff
@@ -0,0 +1,42 @@
...
+[locksmith]: https://github.com/coreos/locksmith/tree/v0.6.2#reboot-windows
+
+# Timing and configuration
+
+Window granularity is at the "minutes" level. For this reason, the configuration parameter `length_minutes` is a plain non-zero integer (instead of a free-form duration string).
+
+In order to ease the case where the same time-window has to be applied on multiple specific days, the `days` parameter accepts a set of weekdays (instead of a single day).
+
+A key part of this logic is that all times and dates are UTC-based, in order to guarantee the correctness of maintenance windows.
+In particular, UTC times are needed to avoid:
```
ashcrow, line 22: This explanation is awesome! :+1:

docs/development/update-strategy-periodic.md
lucab: Ack, good thing to note. It was true for locksmith as well and it's a different thing than UTC timezone, so I've put that info on a dedicated sentence on its own.

PR #294, opened 2020-05-25
docs/images: add a sequence diagram for locked reboots

lucab:
This adds a Message Sequence Chart ([MSC](https://en.wikipedia.org/wiki/Message_sequence_chart)) diagram showing the
interactions between Zincati, rpm-ostree and the lock-manager
when using reboot-slot locks (i.e. `fleet_lock` strategy).

---

Rendered:
msc


lucab:
@zonggen good points.

The user-docs show this from the point of view of finalizing an update. This graphs however is a never-ending loop, it should be really better viewed on a cylindrical display :smile:.

`revX` is the target update revision found via Cincinnati. I'll change that to `TargetRevision` for clarity.

lucab:
@zonggen rebased per comments above.

Issue #293, opened 2020-05-22
rpm-ostree: properly handle staged deployments


lucab:
https://github.com/coreos/zincati/pull/112 introduced some logic to detect local deployments. Its main goal was to avoid auto-updating (again) to deployments that had been already rolled-back.

However, it turned out that the logic there is a bit too naïf. In particular, it does not handle staged-but-not-finalized deployments in any special way, treating them as normal deployments.

This means that in some pathological corner cases, Zincati will skip an update if for some reason it has been already been staged but not finalized locally (e.g. due to manual interventions).
In most cases this should be a temporary condition which will converge to resolution as soon as an unrelated reboot happens OR the staged deployment is dropped OR a newer update is available.

However, it would be better to keep track of staged and non-staged local deployments separately. That in turn will reduce the rate of false positives in the rollback-detection logic.

This bugfix likely intersects with the rework due for https://github.com/coreos/zincati/issues/137.

PR #308, opened 2020-06-10
rpm-ostree/status: detect and handle staged deployments

lucab:
This augments rpm-ostree status querying logic in order to detected
staged-but-not-finalized deployments, i.e. those that haven't really
been persisted locally in the past.
Zincati uses this new logic in order to distinguish between previous
deployments (e.g. in case of rollbacks) and temporarily staged ones
(e.g. in case of manual interventions or half-applied updates). The
latter ones can be considered as future update targets, while the
former ones should not.

Fixes: https://github.com/coreos/zincati/issues/293

### DIFF
--- a/src/rpm_ostree/actor.rs
+++ b/src/rpm_ostree/actor.rs
@@ -65,7 +65,10 @@ impl Handler<FinalizeDeployment> for RpmOstreeClient {
 
 /// Request: query local deployments.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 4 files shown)


lucab:
@jlebon if you have time for a review pass on this, that would be appreciated.

lucab:
@jlebon yes, I think that is a possible path too (i.e. the "half-applied updates" above) and likely how I spotted this initially.

Issue #305, opened 2020-06-08
agent: please report current update strategy and some basic info when started


jlebon:
Right now, if I e.g. add a `/etc/zincati/config.d/90-updates.toml` with a config for the periodic strategy, there's no confirmation that this was picked up by Zincati when I `systemctl restart zincati`. I'd like something like:

```
...
[INFO ] initialization complete, auto-updates logic enabled
[INFO ] update strategy: periodic (next time window in 7h23m)
```

jlebon:
Related: https://github.com/coreos/zincati/issues/94. If we had that and made sure to report the update strategy in the `STATUS` notification, I think maybe we wouldn't need this (though it wouldn't hurt either).

lucab:
Agreed, I think this should be done on startup, regardless of what we do for #94. And we should forward this information to metrics too.

Regarding the periodic strategy, do you care about the next window in logs here specifically? I personally think that a more interesting info (from the point of monitoring and metrics too) is the total weekly schedule duration.

jlebon:
> Regarding the periodic strategy, do you care about the next window in logs here specifically? I personally think that a more interesting info (from the point of monitoring and metrics too) is the total weekly schedule duration.

Could we log both? Hypothetical: if a config is overriding another config to change the day the window occurs, but not the duration, it wouldn't be possible to tell them apart.

zonggen:
This also makes a lot of sense to me and could potentially very helpful when debugging and monitoring.

One thing I'm not entirely sure of regarding https://github.com/coreos/zincati/issues/94 is that when exposing the state of the agent, when should we log the current state? My assumption would be whenever the state changes and during startup.

lucab:
I took the metrics part in https://github.com/coreos/zincati/pull/314, but the "log on startup" part is still to be done.

There is a piece of underlying logic missing, which is adding a function (and tests) in the `weekly` module to compute the remaining time to the next window. I guess the result would be a `Option<NonZeroU32>`. Plus a human time formatter for that result (possibly `chrono` already provides the basic logic).

Other than that, I think this new code could sit next to the other logging we do on startup, in its own `Settings::log_startup_info()`.

PR #314, opened 2020-06-15
strategy: add metrics to track updates strategy configuration

lucab:
This adds two info metrics covering the configured update strategy:
 * `zincati_updates_strategy_mode{strategy="<MODE>"}`
 * `zincati_updates_strategy_periodic_schedule_length_minutes`

The first is an info metrics covering the strategy mode in all cases;
the second only applies to the "periodic" strategy and tracks the
schedule length.

Ref: https://github.com/coreos/zincati/issues/305

### DIFF
--- a/src/strategy/mod.rs
+++ b/src/strategy/mod.rs
@@ -7,6 +7,7 @@ use crate::identity::Identity;
 use failure::{bail, Fallible};
 use futures::prelude::*;
 use log::error;
+use prometheus::{IntGauge, IntGaugeVec};
 use serde::Serialize;
 
 mod fleet_lock;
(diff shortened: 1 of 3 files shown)


lucab:
/cc @jlebon @zonggen 

This address the metrics part of https://github.com/coreos/zincati/issues/305 (the logging up is still up for grab).

With the default "immediate" config, the result is:
```
# HELP zincati_updates_strategy_mode Update strategy mode in use
# TYPE zincati_updates_strategy_mode gauge
zincati_updates_strategy_mode{strategy="immediate"} 1
```

While with the "periodic" config from the [docs](./docs/usage/updates-strategy.md#periodic-strategy), the result is:
```
# HELP zincati_updates_strategy_mode Update strategy mode in use
# TYPE zincati_updates_strategy_mode gauge
zincati_updates_strategy_mode{strategy="periodic"} 1
# HELP zincati_updates_strategy_periodic_schedule_length_minutes Total length of the periodic strategy schedule in use
# TYPE zincati_updates_strategy_periodic_schedule_length_minutes gauge
zincati_updates_strategy_periodic_schedule_length_minutes 150
```

src/strategy/mod.rs
```diff
@@ -28,13 +42,24 @@ pub(crate) enum UpdateStrategy {
...
             "immediate" => UpdateStrategy::new_immediate()?,
             "periodic" => UpdateStrategy::new_periodic(cfg)?,
             "" => UpdateStrategy::default(),
             x => bail!("unsupported strategy '{}'", x),
         };
+
+        // Export info-metrics with details about current strategy.
+        STRATEGY_MODE
+            .with_label_values(&[strategy_name.as_ref()])
+            .set(1);
```
zonggen, line 57: I'm guessing this might be related to how `register_int_gauge_vec` works but when seeing a `zincati_updates_strategy_mode{strategy="immediate"} 1` the `1` after the current strategy makes me wonder if there's second meaning behind the integer other than the meaning of `enabled`. Just a thought definitely not a blocker.

lucab: Those are info metrics, basically just a boolean where the interesting information is encoded via label values (i.e. the `immediate` here). The following article describes well how they are used when querying and aggregating: https://www.robustperception.io/exposing-the-software-version-to-prometheus

Issue #35, opened 2019-06-13
strategy: overhaul internals


lucab:
Carrying over @jlebon's comment from https://github.com/coreos/zincati/pull/33#discussion_r293000323.

Once we settle on the final configuration for `strategy` section (and implement all strategies and options), there will likely be a few unnecessary bits due to initial overshooting/overdesigning features.

This is a reminder ticket to go through the whole `strategy` module to prune all unused bits and streamline the relevant pieces.

lucab:
Relevant places that could be cleaned up:
 * drop all the unused `Identity` arguments
 * verify whether `can_check_and_fetch` is effectively a no-op; if so, it can be safely dropped
 * check if we can and it makes sense to avoid boxed-pinned-results and directly return `impl Future<Output = bool>` in more places (unsure, I fear this will make the code messier)
 * try dropping the `#![allow(unused)]` annotation

PR #316, opened 2020-06-18
strategy: cleanup unused logic

zonggen:
Cleanup checklist:
 - [x] drop all the unused Identity arguments
 - [x] try dropping the #![allow(unused)] annotation
 - [x] verify whether can_check_and_fetch is effectively a no-op; if so, it can be safely dropped
 - [x] check if we can and it makes sense to avoid boxed-pinned-results and directly return impl Future<Output = bool> in more places
 - [x] check if introducing a trait helps in reducing the amount of matches and repetitive calls

Closes: https://github.com/coreos/zincati/issues/35
Signed-off-by: Allen Bai <abai@redhat.com>

### DIFF
--- a/src/cincinnati/client.rs
+++ b/src/cincinnati/client.rs
@@ -6,10 +6,7 @@
 
 // TODO(lucab): eventually move to its own "cincinnati client library" crate
 
-#![allow(unused)]
-
-use failure::{format_err, Error, Fail, Fallible, ResultExt};
-use futures::future;
+use failure::{Fail, Fallible, ResultExt};
 use futures::prelude::*;
 use reqwest::Method;
 use serde::{Deserialize, Serialize};
@@ -154,7 +151,7 @@ impl Client {
     }
 
     /// Map an HTTP response to a service result.
-    async fn map_response(mut response: reqwest::Response) -> Result<Graph, CincinnatiError> {
+    async fn map_response(response: reqwest::Response) -> Result<Graph, CincinnatiError> {
         let status = response.status();
 
         // On success, try to decode graph.
@@ -201,13 +198,6 @@ impl ClientBuilder {
     pub fn query_params(self, params: Option<HashMap<String, String>>) -> Self {
         let mut builder = self;
         builder.query_params = params;
-        builder
-    }
-
-    /// Set (or reset) the HTTP client to use.
-    pub fn http_client(self, hclient: Option<reqwest::Client>) -> Self {
-        let mut builder = self;
-        builder.hclient = hclient;
         builder
     }
 
--- a/src/strategy/immediate.rs
+++ b/src/strategy/immediate.rs
@@ -1,6 +1,6 @@
 //! Strategy for immediate updates.
 
-use failure::{Error, Fallible};
+use failure::Error;
 use futures::future;
 use futures::prelude::*;
 use log::trace;
@@ -8,23 +8,15 @@ use serde::Serialize;
 use std::pin::Pin;
 
 /// Strategy for immediate updates.
-#[derive(Clone, Debug, Serialize)]
-pub(crate) struct StrategyImmediate {
-    /// Whether to check for and fetch updates.
-    check: bool,
-    /// Whether to finalize updates.
-    finalize: bool,
-}
+#[derive(Clone, Debug, Default, Serialize)]
+pub(crate) struct StrategyImmediate {}
 
 impl StrategyImmediate {
     /// Check if finalization is allowed.
     pub(crate) fn can_finalize(&self) -> Pin<Box<dyn Future<Output = Result<bool, Error>>>> {
-        trace!(
-            "immediate strategy, can finalize updates: {}",
-            self.finalize
-        );
+        trace!("immediate strategy, can finalize updates: {}", true);
 
-        let res = future::ok(self.finalize);
+        let res = future::ok(true);
         Box::pin(res)
     }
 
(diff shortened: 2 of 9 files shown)


zonggen:
Couple of questions/thoughts:
1.  Is it still safe to get rid of `check` and `finalize` now, since https://github.com/coreos/zincati/pull/33#discussion_r293007594 was ~almost a year ago..~ more than a year ago
2. Tried to use more `return impl Future<Output = bool>` but got a lot of type errors and in the end it did make the code messier
3. Tried to define a trait that requires: `can_finalize`, `report_steady` and `can_check_and_fetch` and implement it for different strategy structs and let `UpdateStrategy` acts as a parent class of the individual structs.... but did't work like expected and I still needed to use matches at the end. Since inheritance is not available here in Rust, not sure if we could eliminate the match arms at all..

src/cincinnati/client.rs
```diff
@@ -205,7 +202,7 @@ impl ClientBuilder {
     }
 
     /// Set (or reset) the HTTP client to use.
-    pub fn http_client(self, hclient: Option<reqwest::Client>) -> Self {
+    pub fn _http_client(self, hclient: Option<reqwest::Client>) -> Self {
```
lucab, line 205: I did not realize this was actually unused. If so, we can simply drop it.

src/strategy/immediate.rs
```diff
@@ -31,6 +31,7 @@ impl StrategyImmediate {
     pub(crate) fn report_steady(&self) -> Pin<Box<dyn Future<Output = Result<bool, Error>>>> {
         trace!("immediate strategy, report steady: {}", true);
 
+        // TODO(zonggen): https://github.com/coreos/zincati/issues/35
```
lucab, line 34: You can drop this TODO line. Additionally, I think both the `check` and `finalize` fields on this struct can be removed, as I think they are anyway always set to `true` (they were initially supposed to be configurable, but we dropped that part during initial review).

src/strategy/mod.rs
```diff
@@ -96,7 +94,7 @@ impl UpdateStrategy {
     }
 
     /// Check if this agent is allowed to check for updates at this time.
-    pub(crate) fn can_check_and_fetch(&self, _identity: &Identity) -> impl Future<Output = bool> {
+    pub(crate) fn can_check_and_fetch(&self) -> impl Future<Output = bool> {
```
lucab, line 97: I think this function can be removed at all. If I'm not wrong (please double-check that), all the underlying methods always return `Ok(true)` so there is no useful logic performed here.

lucab:
@zonggen regarding 1, yes it should be safe to remove those, they never reached a point where they were actually configurable (we changed our mind mid-way during reviews). Regarding 2 and 3, that's good feebdack and not totally unexpected, good to have thrown pair of eyeballs into making sure of that.

src/cincinnati/client.rs
zonggen: Dropping this unused `http_client`

src/strategy/mod.rs
zonggen: The only two places that used this function are https://github.com/coreos/zincati/blob/master/src/update_agent/actor.rs#L188 and https://github.com/coreos/zincati/blob/master/src/update_agent/actor.rs#L219.

I've removed the function `can_check_and_fetch` and initialized the `ActorFuture` with a constant `actix::fut::ok::<bool, bool, Self>(true)`

zonggen: Also for both `can_finalize` and `report_steady`, strategies `Immediate` and `Periodic` had a hard-coded constant `true` as results. I didn't remove those parts for cleanness of the match block, since unlike `can_check_and_fetch`, they actually have functional values in `FleetLock`.

src/strategy/immediate.rs
zonggen: Dropped both `check` and `finalize`, and added a simple test for `can_finalize`.

src/strategy/immediate.rs
```diff
@@ -34,37 +26,19 @@ impl StrategyImmediate {
...
-    }
 }
 
 impl Default for StrategyImmediate {
     fn default() -> Self {
-        Self {
-            check: true,
-            finalize: true,
-        }
+        Self {}
```
lucab, line 33: You can now drop this `impl Default` block and directly put a `#[derive(Default)]` annotation on the struct.

src/update_agent/actor.rs
```diff
@@ -188,18 +188,16 @@ impl UpdateAgent {
     fn tick_check_updates(&mut self) -> ResponseActFuture<Self, Result<(), ()>> {
         trace!("trying to check for updates");
 
-        let can_check = self.strategy.can_check_and_fetch(&self.identity);
-        let state_change = actix::fut::wrap_future::<_, Self>(can_check)
-            .then(|can_check, actor, _ctx| actor.local_deployments(can_check))
+        let state_change = actix::fut::ok::<bool, bool, Self>(true)
```
lucab, line 191: This looks like a strange construct, would it be possible to start the chain with `self.local_deployments()` directly? If not, we can at least change the `true` into a `()`, and the `_can_check` into a plain `_`.

(I know future types are a bit overwhelming, so if your answer is "I tried but I'm not sure whether it's possible" I'm fine with just leaving a "TODO" note here to later re-evaluate whether it can be cleaned up further)

src/update_agent/actor.rs
```diff
@@ -220,9 +218,8 @@ impl UpdateAgent {
         trace!("trying to stage an update");
 
         let target = release.clone();
-        let can_fetch = self.strategy.can_check_and_fetch(&self.identity);
-        let deploy_outcome = actix::fut::wrap_future::<_, Self>(can_fetch)
-            .then(|can_fetch, actor, _ctx| actor.attempt_deploy(can_fetch, target));
+        let deploy_outcome = actix::fut::ok::<bool, bool, Self>(true)
```
lucab, line 221: Same comment as above regarding trying `self.attemp_deploy(target)`.

lucab:
This needs a rebase on current master too.

src/update_agent/actor.rs
zonggen: It turns out a `self.local_deployments()` directly will do it. Thanks a lot for the hint, I got quite confused about this part ..

src/update_agent/actor.rs
zonggen: Same as before, directly using `self.attempt_deploy(target)`:+1:

src/update_agent/actor.rs
lucab: I totally share the sentiment, there is still a huge developer-UX gap in the futures/async area. It's slowly but steadily improving though.

zonggen:
Updated according to reviews and rebased.

PR #340, opened 2020-08-11
docs: add initial scaffolding for GitHub Pages

travier:
Preview: https://travier.github.io/zincati/

If we are OK with this theme and direction, this will allow us to work on tweaking the doc layout.

With this PR we can start with the github.io domain website and then work on creating a custom domain name for the docs.


docs/_config.yml
```diff
@@ -0,0 +1,40 @@
+title: coreos/zincati
+description: Zincati documentation
+# Temporary baseurl
+baseurl: "/zincati"
+url: "https://coreos.github.io"
+# Comment above and use below for local development
+# url: "http://localhost:4000"
+permalink: /:title/
+
+remote_theme: pmarsceill/just-the-docs
```
lucab, line 10: We should consider forking this into our own repo under the `coreos` organization.

lucab:
LGTM. I've enabled the GH-pages integration. Let's merge and see how it looks like once live.

PR #348, opened 2020-08-27
docs: Fix development link in contributing doc

travier:
Unfortunately this make the link point to an "empty" page when viewing directly on GitHub. This could be improved by adding context in the development page.

Preview: https://travier.github.io/zincati/contributing/


lucab:
> Unfortunately this make the link point to an "empty" page when viewing directly on GitHub.

I think over time we can re-use that "empty" page to host miscellaneous details, like a glossary or some links to external resources.

lucab (approved):

Issue #119, opened 2019-09-03
metrics: expose process start timestamp


lucab:
Zincati should gauge (and expose as a metrics) the UTC timestamp of process start.
This will help remote observers in detecting service restarts or node reboots.
Suggested metrics name: `zincati_process_start_timestamp`.

lucab:
Following Prometheus conventions, it would be good to have this as an (integer) gauge called `process_start_time_seconds`.

PR #349, opened 2020-08-31
metrics: expose process start timestamp

bh7cw:
this gauges `process_start_time_seconds` following Prometheus conventions.

Fixes https://github.com/coreos/zincati/issues/119

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -32,8 +32,16 @@ mod weekly;
 use actix::Actor;
 use failure::ResultExt;
 use log::{info, trace};
+use prometheus::IntGauge;
 use structopt::clap::{crate_name, crate_version};
 use structopt::StructOpt;
+
+lazy_static::lazy_static! {
+    static ref PROCESS_START_TIME: IntGauge = register_int_gauge!(opts!(
+        "process_start_time_seconds",
+        "Start time of the process since unix epoch in seconds."
+    )).unwrap();
+}
 
 fn main() -> failure::Fallible<()> {
     // Parse command-line options.
@@ -68,6 +76,10 @@ fn run_agent() -> failure::Fallible<()> {
         settings.identity.node_uuid.lower_hex(),
         settings.identity.group
     );
+
+    // Expose process start timestamp.
+    let start_time = chrono::Utc::now();
+    PROCESS_START_TIME.set(start_time.timestamp());
 
     trace!("creating actor system");
     let sys = actix::System::builder()


bh7cw:
Tested on FCOS vm via AWS:
```
[core@ip ~]$ sudo socat - UNIX-CONNECT:/run/zincati/public/metrics.promsock
# HELP process_start_time_seconds UTC start time of the process in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1598904228
# HELP zincati_cincinnati_booted_release_is_deadend Whether currently booted OS release is a dead-end.
# TYPE zincati_cincinnati_booted_release_is_deadend gauge
zincati_cincinnati_booted_release_is_deadend 0
# HELP zincati_cincinnati_graph_edges_count Number of edges in Cincinnati update graph.
# TYPE zincati_cincinnati_graph_edges_count gauge
zincati_cincinnati_graph_edges_count 20
# HELP zincati_cincinnati_graph_nodes_count Number of nodes in Cincinnati update graph.
# TYPE zincati_cincinnati_graph_nodes_count gauge
zincati_cincinnati_graph_nodes_count 18
# HELP zincati_cincinnati_update_checks_total Total number of checks for updates to the upstream Cincinnati server.
# TYPE zincati_cincinnati_update_checks_total counter
zincati_cincinnati_update_checks_total 2
# HELP zincati_identity_os_info Information about the underlying booted OS
# TYPE zincati_identity_os_info gauge
zincati_identity_os_info{basearch="x86_64",os_version="32.20200809.3.0",platform="aws",stream="stable"} 1
# HELP zincati_update_agent_last_refresh_timestamp UTC timestamp of update-agent last refresh tick.
# TYPE zincati_update_agent_last_refresh_timestamp gauge
zincati_update_agent_last_refresh_timestamp 1598904231
# HELP zincati_update_agent_latest_state_change_timestamp UTC timestamp of update-agent last state change.
# TYPE zincati_update_agent_latest_state_change_timestamp gauge
zincati_update_agent_latest_state_change_timestamp 1598904232
# HELP zincati_update_agent_updates_enabled Whether auto-updates logic is enabled.
# TYPE zincati_update_agent_updates_enabled gauge
zincati_update_agent_updates_enabled 1
# HELP zincati_updates_strategy_mode Update strategy mode in use
# TYPE zincati_updates_strategy_mode gauge
zincati_updates_strategy_mode{strategy="immediate"} 1
```
Ready for review.

src/main.rs
```diff
@@ -32,9 +32,17 @@ mod weekly;
...
 use failure::ResultExt;
 use log::{info, trace};
+use prometheus::IntGauge;
 use structopt::clap::{crate_name, crate_version};
 use structopt::StructOpt;
 
+lazy_static::lazy_static! {
+    static ref PROCESS_START_TIME: IntGauge = register_int_gauge!(opts!(
+        "process_start_time_seconds",
+        "UTC start time of the process in seconds."
```
cgwalters, line 42: Minor nit: I'd clarify that this is an offset from the [Unix epoch](https://en.wikipedia.org/wiki/Unix_time).

The chrono docs say:

> Returns the number of non-leap seconds since January 1, 1970 0:00:00 UTC (aka "UNIX timestamp").

How about something like:

`Time the process started as non-leap seconds since the "Unix epoch" in UTC`

lucab: The recommend description for this gauge is `Start time of the process since unix epoch in seconds.`, as per https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics.

bh7cw: Thanks for review. I will change to clarify Unix epoch per Prometheus docs.

src/main.rs
```diff
@@ -47,6 +55,10 @@ fn main() -> failure::Fallible<()> {
         .try_init()
         .context("failed to initialize logging")?;
 
+    // Expose process start timestamp.
+    let start_time = chrono::Utc::now();
+    PROCESS_START_TIME.set(start_time.timestamp());
+
```
lucab, line 61: This metrics belongs to the agent service. As such, this whole block would be better moved into `run_agent()`, right after the `info!()` logging.

lucab (commented):
One minor suggestiong regarding code arrangement.

src/main.rs
bh7cw: Thanks for review. Good point. I will move to after the `starting update agent` info.

lucab (approved):

PR #350, opened 2020-09-02
docs/dev: add FleetLock protocol specs

lucab:
This adds a complete description of the FleetLock protocol used by
Zincati for cluster-wide coordinated reboots.

### DIFF
--- a/docs/development/fleetlock/protocol.md
+++ b/docs/development/fleetlock/protocol.md
@@ -47,7 +47,9 @@ All POST requests contain well-formed JSON body according to the following schem
    * `group` (string, mandatory, non-empty): reboot-group of the client
 
 Client ID is a case-sensitive textual label that uniquely identifies a lock holder. It is generated and persisted by each client.
-Client group is a textual label conforming to the regexp `^[a-zA-Z0-9.-]+$`. The recommended default group name is "`default`".
+Client group is a mandatory textual label, conforming to the regexp `^[a-zA-Z0-9.-]+$`. This labels can be configured on each client. A server SHOULD check this value and MAY use it to provide multiple reboot buckets (sorting a fleet of nodes into reboot tiers).
+
+By default, Zincati uses the group name "`default`" unless explicitly configured otherwise.
 
 ### Headers
 
--- a/docs/development/fleetlock/protocol.md
+++ b/docs/development/fleetlock/protocol.md
@@ -95,3 +95,16 @@ Error values carry a type-identifier and a textual description, according to the
 | value  | required | human-friendly error description, as a non-empty JSON string |
 
 This allows clients to show more specific error details to cluster administrators, instead of generic HTTP errors.
+
+For example, an error value like the following could be returned on `/v1/pre-reboot` when all available slots are already in use:
+
+```json
+{
+  "kind": "failed_lock_semaphore_full",
+  "value": "semaphore currently full, all slots are locked already"
+}
+```
+
+Zincati will log this error using the content of `value`, and it will track the `kind` label in metrics.
+
+A server MUST ensure that possible values for `kind` have a bounded/small cardinality.
--- a/docs/development/fleetlock/protocol.md
+++ b/docs/development/fleetlock/protocol.md
@@ -25,7 +25,7 @@ All operations are recursive, meaning that multiple unbalanced lock/unlock actio
 ## Client state-machine
 
 When a client starts it proceeds into either "initialization" or "finalization" state, based on system condition (i.e. whether a finalization is already in progress).
-In "**initialization**" state, the client tries to release any reboot slot it may have previously held.
+In the "**initialization**" state, the client tries to release any reboot slot it may have previously held.
 A successful unlock operation means that the client can proceed into its "**steady**" state and look for further updates.
 When an update is found and locally staged, the client proceed into its "**pre-reboot**" state and tries to lock a reboot slot.
 A successful lock operation means that the client can proceed into its "**finalization**" state and finalize a pending update, then reboot.
--- a/docs/development/fleetlock/protocol.md
+++ b/docs/development/fleetlock/protocol.md
@@ -24,7 +24,7 @@ All operations are recursive, meaning that multiple unbalanced lock/unlock actio
 
 ## Client state-machine
 
-When a client starts it proceeds into either "initialization" or "finalization" state, based on system condition (i.e. whether a finalization is already in progress).
+Clients start off in one of two states based on the system condition: "initialization" or "finalization". There are a number of states between "initialization" and "finalization" as well.
 In the "**initialization**" state, the client tries to release any reboot slot it may have previously held.
 A successful unlock operation means that the client can proceed into its "**steady**" state and look for further updates.
 When an update is found and locally staged, the client proceed into its "**pre-reboot**" state and tries to lock a reboot slot.
--- a/docs/development/fleetlock/protocol.md
+++ b/docs/development/fleetlock/protocol.md
@@ -11,7 +11,7 @@ It is modeled after a distributed counting semaphore with recursive locking and 
 
 ## Overview
 
-This is a request-response protocol, where operations are always initiated by the client (i.e. Zincati).
+The FleetLock protocol is a request-response protocol where operations are always initiated by the client (i.e. Zincati).
 Each operation consists of a JSON payload sent as a POST request to the server.
 
 At an high level, the client can perform two operations:
--- a/docs/development/fleetlock/protocol.md
+++ b/docs/development/fleetlock/protocol.md
@@ -1,0 +1,95 @@
+---
+parent: Development
+nav_order: 3
+layout: default
+---
+
+# FleetLock protocol
+
+This document describes an HTTP-based protocol for orchestrating fleet-wide reboots, used by Zincati.
+It is modeled after a distributed counting semaphore with recursive locking and lock-ownership.
+
+## Overview
+
+This is a request-response protocol, where operations are always initiated by the client (i.e. Zincati).
+Each operation consists of a JSON payload sent as a POST request to the server.
+
+At an high level, the client can perform two operations:
+
+ * `RecursiveLock`: try to reserve (lock) a slot for rebooting
+ * `UnlockIfHeld`: try to release (unlock) a slot that it was previously holding
+
+Semaphore locks are owned, so that only the client that created a lock can release it. All operations are recursive, so that multiple unbalanced lock/unlock actions are allowed.
+
+## Client state-machine
+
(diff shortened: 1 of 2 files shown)


lucab:
FYI @dghubble @rtlong

docs/development/fleetlock/protocol.md
```diff
@@ -0,0 +1,95 @@
...
+### Body
+
+All POST requests contain well-formed JSON body according to the following schema:
+
+ * `client_params` (object, mandatory)
+   * `id` (string, mandatory, non-empty): client identifier (e.g. node name or UUID)
+   * `group` (string, mandatory, non-empty): reboot-group of the client
+
+Client ID is a case-sensitive textual label that uniquely identifies a lock holder. It is generated and persisted by each client.
+Client group is a textual label conforming to the regexp `^[a-zA-Z0-9.-]+$`. The recommended default group name is "`default`".
```
dghubble, line 50: `group` also generated by the client right? Based on the Zincati config. Maybe this could be clearer. Will Zincati client ever send an empty group (requiring the server to do the defaulting?). Will it send a group that doesn't match the regex (i.e. is this info meant for the server implementer?).

docs/development/fleetlock/protocol.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+### Errors
+
+Errors on the service endpoints SHOULD be returned to the client as JSON objects, with a 4xx or 5xx HTTP status code.
+Error values carry a type-identifier and a textual description, according to the following schema:
+
+|  Key   | Optional | Description                                                  |
+|--------|----------|--------------------------------------------------------------|
+| kind   | required | error type identifier, as a non-empty JSON string            |
```
dghubble, line 92: Have an example thoughts on what the `kind` should look like to be reasonable? I'm assuming somewhere down the line Zincati logs will show some descriptive messages that include these fields?

dghubble: For future use cases with drain (or any pre-reboot action), I'm thinking to let a client take the lock but continue returning some error message until the drain (or other action) is complete. Seems like that plan would align with this and the messages could be useful for displaying that in Zincati journal logs.

docs/usage/updates-strategy.md
```diff
@@ -39,8 +39,10 @@ In short, it consists of two operations:
  * lock: before rebooting, a reboot slot must be locked (and confirmed) by the lock-manager.
  * unlock: after rebooting, any reboot slot owned by the node must be unlocked (and confirmed) by the lock-manager before proceeding further.
 
-This protocol is not coupled to any specific backend, and can be implemented on top of any suitable database.
-As an example, [airlock] is a free-software project which implements such protocol on top of [etcd3].
+This protocol is not coupled to any specific backend, and can be implemented on top of any suitable infrastructure:
+ * [airlock] is a free-software project which implements such protocol on top of [etcd3].
+ * a Kubernetes-based reboot-manager is provided as part of [Typhoon](https://github.com/poseidon/fleetlock).
```
dghubble, line 44: yay :tada:

docs/development/fleetlock/protocol.md
lucab: `kind` is meant to be tracked in metrics (as a label of `zincati_strategy_fleet_lock_errors_total`), while `value` is meant to be shown in logs. You can find some internal examples for `kind` at https://github.com/coreos/zincati/blob/v0.0.12/src/fleet_lock/mod.rs#L54-L56, please ensure that allowed values have a bounded/small cardinality. (I'll add these notes to the docs itself).

I didn't get all the details of your second comment, but yes you can use the `value` field to have something custom appearing in the logs. The protocol however does not have an explicit place suitable for a custom hook / locked draining point (I was assuming the out-of-band usage of a systemd-inhibitor for that).

docs/development/fleetlock/protocol.md
lucab: Yes, `group` is a client parameter in Zincati. It can be customized via configuration fragments. Zincati won't send a request with an empty or missing group value. Right now however it doesn't check validity with the regepx, that needs to be fixed, I've opened https://github.com/coreos/zincati/issues/351.

In general I would expect a server to only check the regexp and reject anything which does not conform. The "default" note is confusing here, so I'll drop that part.

docs/development/fleetlock/protocol.md
```diff
@@ -0,0 +1,95 @@
...
+---
+
+# FleetLock protocol
+
+This document describes an HTTP-based protocol for orchestrating fleet-wide reboots, used by Zincati.
+It is modeled after a distributed counting semaphore with recursive locking and lock-ownership.
+
+## Overview
+
+This is a request-response protocol, where operations are always initiated by the client (i.e. Zincati).
```
ashcrow, line 14:
```suggestion
The FleetLock protocol is a request-response protocol where operations are always initiated by the client (i.e. Zincati).
```

docs/development/fleetlock/protocol.md
```diff
@@ -0,0 +1,95 @@
...
+Each operation consists of a JSON payload sent as a POST request to the server.
+
+At an high level, the client can perform two operations:
+
+ * `RecursiveLock`: try to reserve (lock) a slot for rebooting
+ * `UnlockIfHeld`: try to release (unlock) a slot that it was previously holding
+
+Semaphore locks are owned, so that only the client that created a lock can release it. All operations are recursive, so that multiple unbalanced lock/unlock actions are allowed.
```
ashcrow, line 22:
```suggestion
Semaphore locks are owned, so that only the client that created a lock can release it. All operations are recursive meaning multiple unbalanced lock/unlock actions are allowed.
```

docs/development/fleetlock/protocol.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Client state-machine
+
+When a client starts it proceeds into either "initialization" or "finalization" state, based on system condition (i.e. whether a finalization is already in progress).
```
ashcrow, line 26:
```suggestion
Clients start off in one of two states based on the system condition: "initialization" or "finalization". There are a number of states between "initialization" and "finalization" as well.
```

docs/development/fleetlock/protocol.md
```diff
@@ -0,0 +1,95 @@
...
+In "**initialization**" state, the client tries to release any reboot slot it may have previously held.
```
ashcrow, line 27:
```suggestion
In the "**initialization**" state, the client tries to release any reboot slot it may have previously held.
```

docs/development/fleetlock/protocol.md
```diff
@@ -0,0 +1,95 @@
...
+All endpoints defined below are relative to a common deployment-specific base URL:
+
+ * `/v1/pre-reboot`: reserve/lock a reboot slot
+ * `/v1/steady-state`: release/unlock a reboot slot
+
+### Body
+
+All POST requests contain well-formed JSON body according to the following schema:
+
+ * `client_params` (object, mandatory)
```
ashcrow, line 45: nit/idea: It would be nice to be able to link to a jsonschema for this as well. Verified this against the example listed later.

Example:
```json
{
  "$id": "https://github.com/coreos/zincati/....../client_params.schema.json",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "FleetLock Protocol",
  "definitions": {
    "client_params": {
      "type": "object",
      "required": [ "id", "group" ],
      "properties": {
        "id": {
            "type": "string",
            "minLength": 1,
            "description": "client identifier (e.g. node name or UUID)"
        },
            "group": {
            "type": "string",
            "minLength": 1,
            "pattern": "^[a-zA-Z0-9.-]+$",
            "description": "reboot-group of the client"
        }
      }
    }
  },
  "type": "object",
  "required": ["client_params"],
  "properties": {
    "client_params": {
        "$ref": "#/definitions/client_params"
    }
  }
}
```

docs/development/fleetlock/protocol.md
```diff
@@ -0,0 +1,95 @@
...
+
+Errors on the service endpoints SHOULD be returned to the client as JSON objects, with a 4xx or 5xx HTTP status code.
+Error values carry a type-identifier and a textual description, according to the following schema:
+
+|  Key   | Optional | Description                                                  |
+|--------|----------|--------------------------------------------------------------|
+| kind   | required | error type identifier, as a non-empty JSON string            |
+| value  | required | human-friendly error description, as a non-empty JSON string |
+
+This allows clients to show more specific error details to cluster administrators, instead of generic HTTP errors.
```
ashcrow, line 95:
```suggestion
This allows clients to show more specific error details to cluster administrators, instead of generic HTTP errors.

Example:

```json

{
    "kind": "generic_http_500", 
    "value":  "(unknown/generic server error)"
}
```

ashcrow: I think showing an example is helpful but ... I don't think I picked the best example :sweat_smile:.

ashcrow (approved):
Thank you for ensuring there is developer documentation!!! :heart: 

I put in a few non-required suggestions but otherwise LGTM!

docs/development/fleetlock/protocol.md
lucab: Indeed. I was planning to lift https://github.com/opencounter/terraform-fleet-lock-dynamodb/blob/master/json_schema.json at some point after merging the initial spec. Your schema looks nice and fully fledged too!
I was also planning to have https://github.com/poseidon/fleetlock/blob/master/examples/body.json too.

Both are separate file that can land once the directory is in place, so I'd not bundle them in this PR right now.

docs/development/fleetlock/protocol.md
lucab: Good point. I'll add an example for "all slots already locked" negative response, which is a real case with more specific details.

lucab:
@ashcrow @dghubble thanks for the feedback, I've pushed a few more commits on top to clarify doubts and apply suggestions, PTAL.

docs/development/fleetlock/protocol.md
dghubble: Ok, that example of a metrics-style "kind" and a free form "value" dclears up my question I think.

And my 2nd comment probably isn't that important. I was thinking of a usage case. But I think it would fit properly within this protocol. Effectively, the server returning a success means a slot has been reserved. While an error might be used for many cases and that's up to the server (e.g. simple example is "no you can't get the simple lock", fancy example might be "no you can't get a lock yet bc some processing/draining/fireworks/anything are happening"). Seems pretty flexible.

ashcrow:
Looks good to me. Happy to merge when I see @dghubble give a :+1: as well!

dghubble:
My questions were answered, looks good! 👍

PR #352, opened 2020-09-08
cargo: update prometheus to latest version

lucab:
See changelog at https://github.com/tikv/rust-prometheus/blob/master/CHANGELOG.md#0100.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,7 +16,7 @@ dependencies = [
  "futures",
  "lazy_static",
  "log",
- "parking_lot",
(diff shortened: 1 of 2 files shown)


jlebon (approved):

lucab:
Also bumped minimum toolchain to `1.40.0`, as one of the dependencies is now using a feature stabilized there (`cfg(doctest)`).

lucab:
For reference, https://github.com/dependabot/dependabot-core/issues/2512.

PR #359, opened 2020-09-29
cargo: update actix to latest version

lucab:
This updates to actix-0.10 and adapts the existing code to the new API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,105 +2,27 @@
 # It is not intended for manual editing.
 [[package]]
 name = "actix"
-version = "0.9.0"
+version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


travier (approved):
LGTM but I do not know actix/futures well enough to understand the new -> pin change.

lucab:
> I do not know actix/futures well enough to understand the new -> pin change.

For reference, the change aligns to the new type constraints introduced with https://github.com/actix/actix/pull/355. In short, the new API requires that the future response chain can be pinned to a stable memory address, which we are upholding here.

dustymabe:
can we merge this? The version of actix in Fedora is already at 0.10.0.

lucab:
@dustymabe merged; do you need this quickly in a tagged release too?

dustymabe:
I don't think so. We can easily patch the rpm.

PR #360, opened 2020-09-29
rpm-ostree: avoid duplicating error causes

lucab:
This fixes error context on rpm-ostree failed runs in order to
avoid duplicating the underlying error cause. Failure chain already
includes the root cause as the next element, thus there is no need
to also embed it in the parent context.


travier (approved):
LGTM

PR #367, opened 2020-10-01
cincinnati: expose dead-end release information via MOTD

sohankunkerkar:
After checking whether the current Cincinnati node is a dead-end or
not, write out a new MOTD file at `/run/motd.d/deadend.motd`
reflecting the dead-end status. If the current release is a deadend,
a message is printed in the MOTD containing the reason for the
release being a deadend. An info message is also logged. If not a
dead-end release a blank MOTD is written and no message is logged.

To make the final write atomic, a tempfile is written within the
`/run/motd.d/` directory, and renamed to the location
of the complete MOTD to display.

In this case, "pkexec" utility is used to allow Zincati (which runs as
the `zincati` sysuser) to modify root-owned `/run/motd.d/` in order to
write dead-end release information into a MOTD file, and the polkit rule
will ensure that that utility runs only when its invoked by a `zincati-deadend`
binary and the corresponding user is `zincati`.

Resolves: #90
Co-authored by: Robert Fairley (rob.t.fairley@gmail.com)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ rand = "0.7"
 reqwest = { version = "0.10", features = ["json"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+tempfile = "^3.1"
 structopt = "0.3"
 tokio = "0.2"
 toml = "0.5"
--- a/dist/polkit-1/rules.d/zincati.rules
+++ b/dist/polkit-1/rules.d/zincati.rules
@@ -5,4 +5,13 @@ polkit.addRule(function(action, subject) {
         subject.user == "zincati") {
         return polkit.Result.YES;
     }
-})
+});
+
+// Allow Zincati to write dead-end release information as an MOTD fragment.
+polkit.addRule(function(action, subject) {
+    if (action.id == "org.coreos.zincati.deadend" &&  
+        subject.user == "zincati") {
+        return polkit.Result.YES;
+    }
+});
+
--- a/src/cincinnati/mod.rs
+++ b/src/cincinnati/mod.rs
@@ -10,13 +10,14 @@ mod mock_tests;
 use crate::config::inputs;
 use crate::identity::Identity;
 use crate::rpm_ostree::Release;
-use failure::{bail, Fallible};
+use failure::{bail, Fallible, ResultExt};
 use futures::prelude::*;
 use futures::TryFutureExt;
 use prometheus::{IntCounter, IntCounterVec, IntGauge};
 use serde::Serialize;
 use std::collections::BTreeSet;
 use std::pin::Pin;
+use std::sync::atomic::{AtomicU8, Ordering};
 
 /// Metadata key for payload scheme.
 pub static AGE_INDEX_KEY: &str = "org.fedoraproject.coreos.releases.age_index";
@@ -55,8 +56,37 @@ lazy_static::lazy_static! {
         "Total number of errors while checking for updates.",
         &["kind"]
     ).unwrap();
+    static ref DEADEND_STATE : DeadEndState = DeadEndState::default();
 }
 
+/// For tracking a dead-end release.
+pub struct DeadEndState(AtomicU8);
+
+impl Default for DeadEndState {
+    fn default() -> Self {
+        Self(AtomicU8::new(DeadEndState::UNKNOWN))
+    }
+}
+
+impl DeadEndState {
+    const FALSE: u8 = 0;
+    const TRUE: u8 = 1;
+    const UNKNOWN: u8 = 2;
+    pub fn is_not_deadend(&self) -> bool {
+        self.0.load(Ordering::SeqCst) == Self::FALSE
+            || self.0.load(Ordering::SeqCst) == Self::UNKNOWN
+    }
+    pub fn is_deadend(&self) -> bool {
+        self.0.load(Ordering::SeqCst) == Self::TRUE
+            || self.0.load(Ordering::SeqCst) == Self::UNKNOWN
+    }
+    pub fn set_deadend(&self) {
+        self.0.store(Self::TRUE, Ordering::SeqCst);
+    }
+    pub fn set_not_deadend(&self) {
+        self.0.store(Self::FALSE, Ordering::SeqCst);
+    }
+}
 /// Cincinnati configuration.
 #[derive(Debug, Serialize)]
 pub struct Cincinnati {
@@ -130,6 +160,39 @@ impl Cincinnati {
     }
 }
 
+/// Evaluate and record whether booted OS is a dead-end release, and
+/// log that information in a MOTD file.
+fn refresh_deadend_status(node: &Node) -> failure::Fallible<()> {
+    let deadend_reason = evaluate_deadend(node);
+    match &deadend_reason {
+        Some(reason) => {
+            if DEADEND_STATE.is_not_deadend() {
+                DEADEND_STATE.set_deadend();
+                log::info!("dead-end release detected: {}", reason);
+                std::process::Command::new("pkexec")
+                    .arg("/usr/libexec/zincati")
+                    .arg("deadend")
+                    .arg(reason)
+                    .output()
+                    .with_context(|_| "failed to write dead-end release information")?;
+            }
+            BOOTED_DEADEND.set(1);
+        }
+        None => {
+            if DEADEND_STATE.is_deadend() {
+                DEADEND_STATE.set_not_deadend();
+                std::process::Command::new("pkexec")
+                    .arg("/usr/libexec/zincati")
+                    .arg("deadend")
+                    .output()
+                    .with_context(|_| "failed to remove dead-end release MOTD file")?;
+            }
+            BOOTED_DEADEND.set(0);
+        }
+    };
+    Ok(())
+}
+
 /// Walk the graph, looking for an update reachable from the given digest.
 fn find_update(
     graph: client::Graph,
@@ -159,6 +222,9 @@ fn find_update(
     let cur_release = Release::from_cincinnati(cur_node.clone())
         .map_err(|e| CincinnatiError::FailedNodeParsing(e.to_string()))?;
 
+    if let Err(e) = refresh_deadend_status(&cur_node) {
+        log::warn!("failed to refresh dead-end status: {}", e);
+    }
     // Evaluate and record whether booted OS is a dead-end release.
     // TODO(lucab): consider exposing this information in more places
     // (e.g. logs, motd, env/json file in a well-known location).
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -32,4 +32,6 @@ impl CliOptions {
 pub(crate) enum CliCommand {
     #[structopt(name = "agent")]
     Agent,
+    #[structopt(name = "deadend")]
+    Deadend { reason: Option<String> },
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,9 +30,13 @@ mod update_agent;
 mod weekly;
 
 use actix::Actor;
-use failure::ResultExt;
+use failure::{bail, ResultExt};
 use log::{info, trace};
 use prometheus::IntGauge;
+use std::fs::Permissions;
+use std::io::Write;
+use std::os::unix::fs::PermissionsExt;
+use std::{fs, io};
 use structopt::clap::{crate_name, crate_version};
 use structopt::StructOpt;
 
@@ -64,6 +68,7 @@ fn run() -> i32 {
     // Dispatch CLI subcommand.
     let exit = match cli_opts.cmd {
         cli::CliCommand::Agent => run_agent(),
+        cli::CliCommand::Deadend { reason } => write_deadend_release_info(reason),
     };
 
     match exit {
@@ -124,5 +129,42 @@ fn run_agent() -> failure::Fallible<()> {
     trace!("starting actor system");
     sys.run().context("agent failed")?;
 
+    Ok(())
+}
+
+/// Deadend subcommand entry point.
+fn write_deadend_release_info(reason: Option<String>) -> failure::Fallible<()> {
+    if reason.is_some() {
+        // Avoid showing partially-written messages using tempfile and
+        // persist (rename).
+        let mut f = tempfile::Builder::new()
+            .prefix(".deadend.")
+            .suffix(".motd.partial")
+            // Create the tempfile in the same directory as the final MOTD,
+            // to ensure proper SELinux labels are applied to the tempfile
+            // before renaming.
+            .tempfile_in("/run/motd.d")
+            .with_context(|e| format!("failed to create temporary MOTD file: {}", e))?;
+        // Set correct permissions of the temporary file, before moving to
+        // the destination (`tempfile` creates files with mode 0600).
+        fs::set_permissions(f.path(), Permissions::from_mode(0o644))
+            .with_context(|e| format!("failed to set permissions of temporary MOTD file: {}", e))?;
+
+        if let Some(reason) = reason {
+            writeln!(
+                f,
+                "This release is a dead-end and won't auto-update: {}",
+                reason
+            )
+            .with_context(|e| format!("failed to write MOTD: {}", e))?;
+        }
+
+        f.persist("/run/motd.d/85-zincati-deadend.motd")
+            .with_context(|e| format!("failed to persist temporary MOTD file: {}", e))?;
+    } else if let Err(e) = std::fs::remove_file("/run/motd.d/85-zincati-deadend.motd") {
+        if e.kind() != io::ErrorKind::NotFound {
+            bail!("failed to remove dead-end release info file: {}", e);
+        }
+    }
     Ok(())
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2248,6 +2248,7 @@ dependencies = [
  "serde",
  "serde_json",
  "structopt",
+ "tempfile",
  "tokio",
  "toml",
  "url",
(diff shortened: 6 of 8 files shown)


src/main.rs
```diff
@@ -126,3 +131,34 @@ fn run_agent() -> failure::Fallible<()> {
 
     Ok(())
 }
+
+// Deadend subcommand entry point
```
lucab, line 135: Nit: docstrings require three slashes (`///`) and usually end with a dot.

src/main.rs
```diff
@@ -126,3 +131,34 @@ fn run_agent() -> failure::Fallible<()> {
...
+    if let Some(reason) = reason {
+        writeln!(
+            f,
+            "This release is a dead-end and won't auto-update: {}",
+            reason
+        )
+        .with_context(|e| format!("failed to write MOTD: {}", e))?;
+    }
+
+    f.persist("/run/motd.d/deadend.motd")
```
lucab, line 161: It would be good to give this a numerical sorting and the application prefix, e.g. `30-zincati-deadend.motd`.

lucab:
Thanks for the PR, I've left a few comment in review. The approach overall looks good, and indeed less messy than the symlinks chain. I think the biggest gaps in this WIP right now are 1) how to unset a previous deadend and 2) conditional logic in order to only act on state changes.

/cc @travier @jlebon @rfairley in case they want to have a look too.

travier:
Agree with @lucab here. Nice catch for the argv1.

dist/polkit-1/rules.d/zincati.rules
```diff
@@ -5,4 +5,12 @@ polkit.addRule(function(action, subject) {
         subject.user == "zincati") {
         return polkit.Result.YES;
     }
-})
+});
+
+// Allow Zincati to write dead-end release information into a MOTD file.
+polkit.addRule(function(action, subject) {
+    if (action.id == "org.freedesktop.policykit.exec" &&
+        action.lookup("program") == "/usr/libexec/zincati-deadend" &&  subject.user == "zincati") {
```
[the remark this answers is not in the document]
cgwalters: Yeah, seems likely.

src/cli/mod.rs
```diff
@@ -29,7 +29,11 @@ impl CliOptions {
 
 /// CLI sub-commands.
 #[derive(Debug, StructOpt)]
+#[structopt(rename_all = "kebab-case")]
```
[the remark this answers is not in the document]
cgwalters: Ah heh.  I advised the opposite and ended up doing https://github.com/coreos/zincati/pull/368

cgwalters:
Not to completely derail this but I did think of a potentially much simpler solution that doesn't involve polkit:
`ExecStartPre=+/usr/bin/setfacl -m u:zincati:rwx /run/motd.d`
Then zincati can just write to `/run/motd.d`.  (Now yes it could unlink or modify other motd files but eh)

rfairley:
:+1: I like the idea of keeping the final MOTD within `/run/motd.d` - avoids the need to have a corresponding SELinux rule added as the implementation in #276 needs. Note if this route is taken, this commit https://github.com/fedora-selinux/selinux-policy/pull/398/commits/760811d670d0399c38df64afe9f83ded4f96d5ef should be reverted.

One concern on the approach I have is: could the temporary `deadend.motd.partial` file be read by pam_motd callers, e.g. sshd, sometime before the final rename to `deadend.motd`, causing partial output to be displayed? It's unlikely that logins will hit this, but if this can happen it may take away from the original benefit from using the tmpfile+rename pattern.

If this is the case, one way around the issue may be to modify `pam_motd` to only display filenames with `.motd` suffix (so a `.motd.partial` file would never be displayed) - see https://github.com/linux-pam/linux-pam/blob/e21fd1f344a67844925ab9a06b0f54299c94b56c/modules/pam_motd/pam_motd.c#L171. `login` does filter MOTDs by `.motd` suffix (https://github.com/karelzak/util-linux/commit/5a528e2c6ff9735266fc2607c359e925b074bf2c#diff-77a7801e7ac65d85ef624043d0dd2647R250) .

src/cincinnati/mod.rs
```diff
@@ -55,6 +56,7 @@ lazy_static::lazy_static! {
         "Total number of errors while checking for updates.",
         &["kind"]
     ).unwrap();
+    static ref DEADEND_RELEASE_EXISTS : AtomicBool = AtomicBool::new(false);
```
lucab, line 59: As I mentioned in my [previous review](https://github.com/coreos/zincati/pull/367#discussion_r498668206), this is a [tristate](https://en.wikipedia.org/wiki/Three-valued_logic) (`unknown` / `true` / `false`), not a boolean.

In particular, when the service starts it doesn't know its deadend state (i.e. it starts in `unknown` mode) until the first fetch. This currently assumes the service is always starting from `false`, which is not guaranteed if the service was running before with a `true` state and then got restarted.

You can model this by wrapping an [AtomicU8](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU8.html) into a new object and exposing relevant set and query methods.

src/main.rs
```diff
@@ -126,3 +131,38 @@ fn run_agent() -> failure::Fallible<()> {
...
+                "This release is a dead-end and won't auto-update: {}",
+                reason
+            )
+            .with_context(|e| format!("failed to write MOTD: {}", e))?;
+        }
+
+        f.persist("/run/motd.d/85-zincati-deadend.motd")
+            .with_context(|e| format!("failed to persist temporary MOTD file: {}", e))?;
+    } else {
+        std::fs::remove_file("/run/motd.d/85-zincati-deadend.motd")?;
```
lucab, line 165: I think you'll need to special case the file not found case here, which is not an error condition in this case.

lucab:
@cgwalters nice. It can even be described declaratively via `systemd-tmpfiles`. However it goes in the same direction of the previous symlinking approach, i.e. the unprivileged `zincati` user being able to freely change (in particular, symlink) global content. After @travier feedback, I think there was some consensus about trying to avoid that.

@rfairley good point. The tmpfile is a dotfile (`.deadend[...]`) under the assumption that hidden files won't be sourced by the MOTD assembling logic. I didn't check all the consumers to make sure they actually behave that way, but it may be worth patching them in that direction if it isn't the case already.

sohankunkerkar:
@lucab Could you please take another look? 

> One concern on the approach I have is: could the temporary deadend.motd.partial file be read by pam_motd callers, e.g. sshd, sometime before the final rename to deadend.motd, causing partial output to be displayed? It's unlikely that logins will hit this, but if this can happen it may take away from the original benefit from using the tmpfile+rename pattern.

I'm not sure how to reproduce this issue locally (if at all it exists).

sohankunkerkar:
> If this is the case, one way around the issue may be to modify pam_motd to only display filenames with .motd suffix (so a .motd.partial file would never be displayed) - see https://github.com/linux-pam/linux-pam/blob/e21fd1f344a67844925ab9a06b0f54299c94b56c/modules/pam_motd/pam_motd.c#L171. login does filter MOTDs by .motd suffix (karelzak/util-linux@5a528e2#diff-77a7801e7ac65d85ef624043d0dd2647R250) .

This sounds like a good idea. I can submit a patch upstream if it's a viable option and can mitigate the issue related to motd displaying only partial output.

src/cincinnati/mod.rs
```diff
@@ -55,6 +57,26 @@ lazy_static::lazy_static! {
...
+#[derive(Debug, Clone)]
+pub enum DeadEndState {
+    Unknown = 0,
+    True = 1,
+    False = 2,
+}
+
+impl From<Arc<AtomicU8>> for DeadEndState {
+    fn from(x: Arc<AtomicU8>) -> Self {
+        match x.load(Ordering::Relaxed) {
```
lucab, line 73: Here and below, mind detailing why you went with `Relaxed` ordering?

src/main.rs
```diff
@@ -7,7 +7,8 @@
 extern crate fail;
 #[macro_use]
 extern crate prometheus;
-
+#[macro_use]
+extern crate failure;
```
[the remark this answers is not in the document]
sohankunkerkar: Agree, let me delete that line.

src/cincinnati/mod.rs
sohankunkerkar: I think that was a bad choice

dist/polkit-1/rules.d/org.freedesktop.policykit.pkexec.run-zincati-deadend.policy
```diff
@@ -0,0 +1,16 @@
...
+ "-//freedesktop//DTD PolicyKit Policy Configuration 1.0//EN"
+ "http://www.freedesktop.org/standards/PolicyKit/1/policyconfig.dtd">
+<policyconfig>
+  <action id="org.freedesktop.policykit.pkexec.run-zincati-deadend">
+    <description>Allow Zincati to write dead-end release information</description>
+    <defaults>
+      <allow_any>yes</allow_any>
+      <allow_inactive>yes</allow_inactive>
+      <allow_active>yes</allow_active>
+    </defaults>
```
travier, line 12: I think this should ideally be removed or at least set to no as in this current form this allows anyone to call the action.

dist/polkit-1/rules.d/org.freedesktop.policykit.pkexec.run-zincati-deadend.policy
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE policyconfig PUBLIC
+ "-//freedesktop//DTD PolicyKit Policy Configuration 1.0//EN"
+ "http://www.freedesktop.org/standards/PolicyKit/1/policyconfig.dtd">
+<policyconfig>
+  <action id="org.freedesktop.policykit.pkexec.run-zincati-deadend">
```
travier, line 6:
```suggestion
  <action id="org.coreos.zincati.deadend">
```

dist/polkit-1/rules.d/zincati.rules
```diff
@@ -5,4 +5,12 @@ polkit.addRule(function(action, subject) {
         subject.user == "zincati") {
         return polkit.Result.YES;
     }
-})
+});
+
+// Allow Zincati to write dead-end release information into a MOTD file.
+polkit.addRule(function(action, subject) {
+    if (action.id == "org.freedesktop.policykit.pkexec.run-zincati-deadend" &&
```
travier, line 12:
```suggestion
    if (action.id == "org.coreos.zincati.deadend" &&  
```

src/main.rs
```diff
@@ -126,3 +131,40 @@ fn run_agent() -> failure::Fallible<()> {
...
+            .prefix(".deadend.")
+            .suffix(".motd.partial")
+            // Create the tempfile in the same directory as the final MOTD,
+            // to ensure proper SELinux labels are applied to the tempfile
+            // before renaming.
+            .tempfile_in("/run/motd.d")
+            .with_context(|e| format!("failed to create temporary MOTD file: {}", e))?;
+        // Set correct permissions of the temporary file, before moving to
+        // the destination (`tempfile` creates files with mode 0600).
+        fs::set_permissions(f.path(), Permissions::from_mode(0o664))
```
travier, line 150:
```suggestion
        fs::set_permissions(f.path(), Permissions::from_mode(0o644))
```

Any particular not to use `644` here?

travier (commented):
Some changes requested but LGTM overall. I'm not sure about the polkit action changes and we will have to double check that this does not allow anyone from calling the action.

dist/polkit-1/rules.d/org.freedesktop.policykit.pkexec.run-zincati-deadend.policy
travier: From https://wiki.archlinux.org/index.php/Polkit#Actions

src/main.rs
sohankunkerkar: hmm... I just carried this change from the (previous)[https://github.com/coreos/zincati/pull/276] PR, so I don't have any strict opinion here.

dist/polkit-1/rules.d/org.freedesktop.policykit.pkexec.run-zincati-deadend.policy
sohankunkerkar: Yeah, makes sense. I was under the impression that this action will get called only for the `zincati` user when invoked by the `zincati.rules`(https://github.com/coreos/zincati/pull/367/files#diff-fb15afa4bf5dbf6d827a99f2d5d1af3eR13)

sohankunkerkar: WDYT about this change?
```
                        <allow_any>no</allow_any>
                        <allow_inactive>no</allow_inactive>
                        <allow_active>auth_admin_keep</allow_active>
```

travier: I think it should always be no for all entries here as this will never be called interactively (zincati is a daemon) and we only want to allow the case covered in the additional polkit rule we add.

dist/polkit-1/rules.d/org.freedesktop.policykit.pkexec.run-zincati-deadend.policy
```diff
@@ -0,0 +1,16 @@
...
+    <description>Allow Zincati to write dead-end release information</description>
```
travier, line 7:
```suggestion
    <description>Write dead-end release information as an MOTD fragment via Zincati</description>
```

src/main.rs
lucab: @sohankunkerkar is the resulting file owned by `root:root`? If so, this is likely not going to matter but `0644` should be enough.

sohankunkerkar: Yeah, it's owned by `root:root`
```
[core@localhost ~]$ ls -ltr /run/motd.d/85-zincati-deadend.motd 
-rw-rw-r--. 1 root root 109 Oct  9 15:43 /run/motd.d/85-zincati-deadend.motd

```

lucab:
Self-notes:
 * the global static `DEADEND_STATE` is not nice to have but carving it out requires a larger refactoring. It can be tackled later on after merging this feature PR first.
 * shelling out to pkexec is a synchronous operation which should not performed in the middle of the async chain. It can indefinitely block and stall the executor. In the happy path this should not be a big concern, so let's tackle this afterwards.
 * in the hot-path of update checking we don't really care if deadend-motd refresh failed for some reason, so a fire-and-forget approach may be enough. Or at least a timeout in order to make sure the core loop keeps going.

All points above can be likely solved by moving this logic to its own `SyncArbiter`.

lucab:
@sohankunkerkar can you please rebase on latest `master` and drop the WIP?

Cargo.toml
```diff
@@ -10,33 +10,34 @@ repository = "https://github.com/coreos/zincati"
 edition = "2018"
 
 [dependencies]
-actix = "0.10"
```
travier, line 13: Looks like a bad rebase here.

sohankunkerkar: @travier Just rectified it. Could you please take another look?

travier (changes requested):
Two minor nits, otherwise LGTM.

sohankunkerkar:
> Self-notes:
> 
>     * the global static `DEADEND_STATE` is not nice to have but carving it out requires a larger refactoring. It can be tackled later on after merging this feature PR first.
> 
>     * shelling out to pkexec is a synchronous operation which should not performed in the middle of the async chain. It can indefinitely block and stall the executor. In the happy path this should not be a big concern, so let's tackle this afterwards.
> 
>     * in the hot-path of update checking we don't really care if deadend-motd refresh failed for some reason, so a fire-and-forget approach may be enough. Or at least a timeout in order to make sure the core loop keeps going.
> 
> 
> All points above can be likely solved by moving this logic to its own `SyncArbiter`.

@lucab Should I capture this point in Jira? It would be helpful for tracking purposes.

travier (approved):

travier:
OK, how do we test that? Should we make a kola test?

lucab:
@sohankunkerkar no need to right now, I'll explore a bit the option later and then if I find an arrangement that seems to fit better I'll open a ticket here.

travier:
We also need to make a test to make sure only the Zincati user gets access to this action.

sohankunkerkar:
> OK, how do we test that? Should we make a kola test?
Yeah, good point. The kola test would definitely cover the corner cases.

lucab:
@travier I checked the polkit config (only, without binary changes) on latest FCOS `stable` this way:
```
$ id
uid=1000(core) gid=1000(core) groups=1000(core),4(adm),10(wheel),16(sudo),190(systemd-journal) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
$ pkexec /usr/libexec/zincati deadend
Error executing command as another user: Not authorized

$ id
uid=979(zincati) gid=979(zincati) groups=979(zincati) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
$ pkexec /usr/libexec/zincati deadend
USAGE:
    zincati [FLAGS] <SUBCOMMAND>

$ id
uid=979(zincati) gid=979(zincati) groups=979(zincati) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
$ pkexec /usr/libexec/zincati agent
==== AUTHENTICATING FOR org.freedesktop.policykit.exec ====
Authentication is needed to run `/usr/libexec/zincati' as the super user
```

I think it works as expected, do you have other things in mind to check?

lucab (changes requested):

lucab (changes requested):

lucab (approved):

lucab:
LGTM, thanks for the multiple iterations towards this!

travier:
Thanks @sohankunkerkar!

PR #368, opened 2020-10-02
rpm_ostree: Use rename_all = kebab-case

cgwalters:
A great thing about serde (as well as structopt) is that
you can achieve a high level of
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

Use `rename_all` here because that's the naming standard,
and if we start parsing other values then we don't
need to specify a rename attribute for each one.

### DIFF
--- a/src/rpm_ostree/cli_status.rs
+++ b/src/rpm_ostree/cli_status.rs
@@ -13,9 +13,9 @@ pub struct StatusJSON {
 
 /// Partial deployment object (only fields relevant to zincati).
 #[derive(Debug, Deserialize)]
+#[serde(rename_all = "kebab-case")]
 pub struct DeploymentJSON {
     booted: bool,
-    #[serde(rename = "base-checksum")]
     base_checksum: Option<String>,
     #[serde(rename = "base-commit-meta")]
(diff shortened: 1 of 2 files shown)


lucab:
In general, it isn't a matter of DRY but more about making sure than an internal field renaming does not trigger action at a distance subtly breaking serialization. And also making the source easier to grep for casual reader.
But I'm not very consistent myself, so I'm fine with going this way and the code change looks fine, happy to merge this.

I've just realized that none of the existing [fixtures](https://github.com/coreos/zincati/tree/master/tests/fixtures) are actually covering the `base-checksum` field. Can you please add a JSON and a test to exercise that?

cgwalters:
> In general, it isn't a matter of DRY but more about making sure than an internal field renaming does not trigger action at a distance subtly breaking serialization.

Right, but tests cover that.

> And also making the source easier to grep for casual reader.

Yeah, true.

> I've just realized that none of the existing fixtures are actually covering the base-checksum field. Can you please add a JSON and a test to exercise that?

Done.

lucab (approved):

PR #372, opened 2020-10-09
cargo: remove carets from semver requirements

lucab:
This removes all carets (^) from dependencies, resulting in
semver entries which are semantically equivalent from cargo
point of view.
This repository is starting to accumulate stale dependencies, as
dependabot is having troubles when trying to update them. The
issue seems related to `^` requirements confusing the bot.

Ref: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,38 +10,38 @@ repository = "https://github.com/coreos/zincati"
 edition = "2018"
 
 [dependencies]
-actix = "^0.10"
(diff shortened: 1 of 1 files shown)


lucab:
/cc @travier for review

See https://github.com/coreos/afterburn/pull/496 for more context.

travier (approved):

PR #379, opened 2020-10-14
cincinnati: swap order of dead-end refreshing logic

lucab:
This rearranges the deadend refreshing logic so that metrics are always
refreshed, and internal state updated only after the change has been
persisted to MOTD. This allows retrying to persist the state on the next
graph update, in case of transient local errors.
It also inverts condition checks to properly account for ternary logic and
unknown state.


cgwalters (approved):
Looks sane.

travier (approved):

PR #380, opened 2020-10-14
cli: split subcommands logic out of main

lucab:
This splits and moves subcommands logic out of main, into their own
`cli` modules.


cgwalters (approved):
Looks sane to me.

PR #385, opened 2020-10-19
docs/dev: fix fleetlock documentation

olivierlemasle:
- Fix an error in fleet lock request example (group named `default` in the example request, but named `workers` in the paragraph before)
- Update to new fleet lock spec location


lucab (approved):

lucab:
Thanks, good catch!

PR #387, opened 2020-10-20
docs/dev: add graph for actors and interactions

lucab:
This adds a graphviz diagram with all actors involved in Zincati
agent logic, and their interactions.

---

Rendered:
diagram


travier:
Looks great. Should we add this to a doc page?

lucab:
@travier yes, I plan to have an "Actor system" docpage which will include the diagram inline (still in-progress).

travier (approved):

PR #395, opened 2020-10-30
cli/deadend: split explicit set/unset verbs and logic

lucab:
This reworks the subcommand for deadend MOTD handling into two
explicit verbs:
 * `zincati deadend-motd set --reason <REASON>`
 * `zincati deadend-motd unset`

It also adds tests covering flags parsing for the subcommand.


lucab:
The previous implicit logic relying on `Option<String>` already caused subtle troubles in the original PR: https://github.com/coreos/zincati/pull/367#discussion_r504031200. Let's change it now before landing this subcommand in a released version.

travier (approved):
LGTM. Any particular _reason_ to use `'set --reason <reason>'` as it could also be `'set <reason>'`? (pun not intended :sweat_smile: )

lucab:
@travier just making clear what the argument is, in case some more things need to be added in the future.

Issue #398, opened 2020-11-02
ci: move Travis jobs to github actions


lucab:
Travis just implemented a [pricing change](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) which affects the whole `coreos` organization, and all jobs here are stuck as the free credit already run out.

It is probably time to move those jobs to GitHub actions.

PR #399, opened 2020-11-02
ci: move Travis jobs to GitHub actions

lucab:
This is a port of existing Travis rules to GitHub actions. The only one missing is the `aarch64` job, as we don't have arm64 runners.

Closes: https://github.com/coreos/zincati/issues/398


lucab:
This is green in my local fork: https://github.com/lucab/zincati/actions/runs/341737167

.github/workflows/rust.yml
```diff
@@ -0,0 +1,81 @@
...
+        uses: actions/checkout@v2
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.channel }}
+          default: true
+      - name: cargo build
+        run: cargo build
+      - name: cargo test
+        run: cargo test
```
jlebon, line 81: I noticed we split out the stable one out of the matrix and added additional steps there. Are you concerned that those extra steps might fail on beta/nightly? Would be nice if we could combine them.

jlebon (approved):
One comment, but LGTM as is too.

.github/workflows/rust.yml
lucab: I marked beta+nightly as fallible, but I had to split them because `continue-on-error` is a per-job boolean now. As these can fail anyway, I don't think it's worth doing the extended set of checks we have on stable.

jlebon: Ahh gotcha, missed the `continue-on-error` difference.

travier (approved):
LGTM

PR #402, opened 2020-11-05
workflows/rust: avoid duplicated runs

lucab:
This refines the triggering conditions for GH actions in order
to avoid running duplicated jobs on local branches.


lucab:
I noticed this on dependabot PRs, as they have double the amount of expected runs: https://github.com/coreos/zincati/pull/401.

travier (approved):
LGTM. I was wondering about that when reviewing the first CI PR but did not act on it 🙂.

PR #409, opened 2020-11-20
contrib: add dev container flow

lucab:
This adds a Dockerfile for a container development image, and wires
it to GitHub workflows.
Additionally, the workflow produces a rootfs-override artifact (with
full-debug binary) that can be directly plugged into coreos-assembler
overrides.

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -1,0 +1,25 @@
+---
+name: Containers
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  container-dev:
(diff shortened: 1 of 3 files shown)


lucab:
Artifacts sample from this PR: https://github.com/coreos/zincati/actions/runs/374211339

travier (approved):

travier:
Looks great. Maybe we could also or alternatively consider building RPMs with Packit (https://packit.dev/)?

lucab:
Yes, this is probably just a stopgap. @jlebon is looking into packit and RPMs/COPRs. However that brings up some further doubts regarding duplicating/diverging specfiles.

PR #410, opened 2020-11-23
strategy: record details on startup

lucab:
This logs an info message on startup, in order to record some
details related to the update strategy in use.
In particular, this is helpful when using the 'periodic' strategy
as it shows the remaining time till the next update window.

Closes: https://github.com/coreos/zincati/issues/305

### DIFF
--- a/src/strategy/mod.rs
+++ b/src/strategy/mod.rs
@@ -49,27 +49,47 @@ impl UpdateStrategy {
             x => bail!("unsupported strategy '{}'", x),
         };
 
+        Ok(strategy)
+    }
+
+    /// Record strategy details to metrics and logs.
+    pub(crate) fn record_details(&self) {
         // Export info-metrics with details about current strategy.
         STRATEGY_MODE
-            .with_label_values(&[strategy.configuration_label()])
+            .with_label_values(&[self.configuration_label()])
             .set(1);
-        if let UpdateStrategy::Periodic(p) = &strategy {
+
+        if let UpdateStrategy::Periodic(p) = self {
             let sched_length = p.schedule_length_minutes();
             PERIODIC_LENGTH.set(sched_length as i64);
         };
 
-        Ok(strategy)
+        log::info!("update strategy: {}", self.human_description());
     }
 
     /// Return the configuration label/name for this update strategy.
     ///
     /// This can be used to match back an instantiated strategy to the mode label
     /// from configuration.
-    pub(crate) fn configuration_label(&self) -> &'static str {
+    fn configuration_label(&self) -> &'static str {
         match self {
             UpdateStrategy::FleetLock(_) => StrategyFleetLock::LABEL,
             UpdateStrategy::Immediate(_) => StrategyImmediate::LABEL,
             UpdateStrategy::Periodic(_) => StrategyPeriodic::LABEL,
+        }
+    }
+
+    /// Return the human description for this strategy.
+    pub(crate) fn human_description(&self) -> String {
+        match self {
+            UpdateStrategy::FleetLock(_) => self.configuration_label().to_string(),
+            UpdateStrategy::Immediate(_) => self.configuration_label().to_string(),
+            UpdateStrategy::Periodic(p) => format!(
+                "{}, total schedule length {} minutes (next window {})",
+                self.configuration_label(),
+                p.schedule_length_minutes(),
+                p.human_remaining()
+            ),
         }
     }
 
--- a/src/strategy/periodic.rs
+++ b/src/strategy/periodic.rs
@@ -50,6 +50,17 @@ impl StrategyPeriodic {
         self.schedule.length_minutes()
     }
 
+    /// Return the remaining duration to next window, in human terms.
+    pub(crate) fn human_remaining(&self) -> String {
+        let datetime = chrono::Utc::now();
+        let remaining = self.schedule.remaining_to_datetime(&datetime);
+        match remaining {
+            None => "not found".to_string(),
+            Some(ref d) => WeeklyCalendar::human_remaining_duration(d)
+                .unwrap_or_else(|_| "unknown".to_string()),
+        }
+    }
+
     /// Check if finalization is allowed.
     pub(crate) fn can_finalize(&self) -> Pin<Box<dyn Future<Output = Result<bool, Error>>>> {
         let datetime_now = chrono::Utc::now();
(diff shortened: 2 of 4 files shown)


lucab:
Startup logs now look like this:
```
[INFO ] starting update agent (zincati 0.0.15-alpha.0)
[INFO ] Cincinnati service: https://updates.coreos.fedoraproject.org
[INFO ] agent running on node 'ABC', in update group 'XYZ'
[INFO ] initialization complete, auto-updates logic enabled
[INFO ] update strategy: periodic, total schedule length 30 minutes (next window in 6d 9h 28m)
```

lucab:
The pieces in the log message are already split, I tried shuffling them in a few different ways and this was the arrangement that looked best to me. Let me know if you'd prefer some different arrangement instead.

src/weekly/mod.rs
```diff
@@ -51,6 +51,71 @@ impl WeeklyCalendar {
...
+        let days = remaining.num_days() % 7;
+        if days > 0 {
+            human.push_str(&format!("{}d ", days));
+            earlier_output = true;
+        } else {
+            earlier_output = false;
+        }
+        let hours = remaining.num_hours() % 24;
+        if hours > 0 || earlier_output {
+            human.push_str(&format!("{}h ", hours));
```
cgwalters, line 111: Not that it really matters here but I think it's better to do
`write!(&mut human, "{}h", hours);`
to avoid box allocating an intermediate string.

src/weekly/mod.rs
```diff
@@ -51,6 +51,71 @@ impl WeeklyCalendar {
...
+        };
+        Some(chrono::Duration::minutes(remaining_mins))
+    }
+
+    /// Format remaining duration till the next window in human terms.
+    pub fn human_remaining_duration(remaining: &chrono::Duration) -> String {
+        let mut human = String::new();
+        if remaining.is_zero() {
+            human.push_str("now");
+            return human;
```
cgwalters, line 97: Potentially clearer to do:
```
if remaining.is_zero() {
    return "now".to_string();
}
let mut human = String::new();
...
```

Also I find the variable naming here would be clearer if it was `timestr` or `humanreadable` perhaps...a `human` variable feels very generic :smile:

src/strategy/mod.rs
```diff
@@ -49,30 +49,53 @@ impl UpdateStrategy {
...
 
+    /// Return the human description for this strategy.
+    pub(crate) fn human_description(&self) -> String {
+        match self {
+            UpdateStrategy::FleetLock(_) => self.configuration_label().to_string(),
+            UpdateStrategy::Immediate(_) => self.configuration_label().to_string(),
+            UpdateStrategy::Periodic(p) => {
+                let label = self.configuration_label();
+                let remaining = p.human_remaining();
+                let length = p.schedule_length_minutes();
```
cgwalters, line 90: Any reason for the intermediate variables here versus calling them inline in the `format!()`?

src/weekly/mod.rs
```diff
@@ -88,6 +153,11 @@ impl WeeklyCalendar {
         u64::from(measured)
     }
 
+    /// Return true if the calendar contains no time-windows.
+    pub fn is_empty(&self) -> bool {
+        self.windows.iter().count() == 0
```
cgwalters, line 158: `self.windows.iter().next().is_some()` is better for general iterators that don't have size hints I'd say.

src/weekly/mod.rs
```diff
@@ -51,6 +51,71 @@ impl WeeklyCalendar {
...
+        if self.contains_datetime(&datetime) {
+            return Some(chrono::Duration::zero());
+        }
+
+        let timepoint = utils::datetime_as_weekly_minute(datetime);
+        // Next window is this week, just subtract remaining minutes.
+        if let Some(next) = self
+            .windows
+            .iter_sorted()
+            .find(|x| x.range.start >= timepoint)
```
cgwalters, line 72: This logic confuses me; why are we looking at all windows versus just the next one?

dustymabe:
> Startup logs now look like this:
> 
> ```
> [INFO ] starting update agent (zincati 0.0.15-alpha.0)
> [INFO ] Cincinnati service: https://updates.coreos.fedoraproject.org
> [INFO ] agent running on node 'ABC', in update group 'XYZ'
> [INFO ] initialization complete, auto-updates logic enabled
> [INFO ] update strategy: periodic, total schedule length 30 minutes (next window in 6d 9h 28m)
> ```

The above LGTM

src/weekly/mod.rs
lucab: I'm not sure I get your question. This traverses the windows tree and stops at the [first one](https://doc.rust-lang.org/nightly/core/iter/trait.Iterator.html#method.find) which occurs after the given datetime, if any.
As such window may not exist, the lines afterwards take care of wrapping to the next week.

lucab:
@cgwalters thanks for the review! I added a fixup commit addressing the comments, and left a note on a question that I did not fully understand.

src/weekly/mod.rs
cgwalters: I didn't look in depth at the data structures here but wouldn't the window tree in sorted order give us the first window we're looking for, and we just need to determine whether or not it's this week?

cgwalters:
LGTM as is!

src/strategy/periodic.rs
```diff
@@ -50,6 +50,17 @@ impl StrategyPeriodic {
         self.schedule.length_minutes()
     }
 
+    /// Return the remaining duration to next window, in human terms.
+    pub(crate) fn human_remaining(&self) -> String {
+        let datetime = chrono::Utc::now();
+        let remaining = self.schedule.remaining_to_datetime(&datetime);
+        match remaining {
+            None => "unavailable".to_string(),
```
jlebon, line 58: Minor/optional: should this be e.g. "not found" so that the description reads "next window not found" instead?

src/weekly/mod.rs
```diff
@@ -409,4 +477,55 @@ mod tests {
...
+    fn human_remaining() {
+        use chrono::Duration;
+
+        let cases = vec![
+            (0, "now"),
+            (1, "in 1m"),
+            (59, "in 59m"),
+            (60, "in 1h 0m"),
+            (61, "in 1h 1m"),
+            (61, "in 1h 1m"),
```
jlebon, line 515: Repeated test case.

jlebon (approved):
Minor comments, LGTM as is!

src/weekly/mod.rs
lucab: I see, I think you are suggesting the `find_next` to be tree-traversing (log) instead of iter-traversing (linear). The library does not expose that primitive right now. I'll keep this code for now and see if I can augment the library for later.

lucab:
Thanks all for the feedback!

Issue #353, opened 2020-09-10
docs/dev: add installing/building/testing instructions


bh7cw:
Add more instructions in the docs, like:
  - installing the toolchain
  - building the binary/package
  - testing on a FCOS system

This helps new contributors start working smoothly.

PR #412, opened 2020-11-27
docs/dev: add a quickstart guide

lucab:
This adds a quickstart guide for developers, showing how to build/test
Zincati and how to embed it into an OS image via coreos-assembler.

Closes: https://github.com/coreos/zincati/issues/353

### DIFF
--- a/docs/development/quickstart.md
+++ b/docs/development/quickstart.md
@@ -1,0 +1,65 @@
+# Development quickstart
+
+This is quick start guide for developing and building this project from source on a Linux machine.
+
+## Get the source
+
+The canonical development location of this project is on GitHub. You can fetch the full source with history via `git`:
+
+```sh
+git clone https://github.com/coreos/zincati.git
+cd zincati
+```
+
+It is recommend to fork a copy of the project to your own GitHub account, and add it as an additional remote:
+
+```sh
+git remote add my-fork git@github.com:<YOURUSER>/zincati.git
+```
+
+## Install Rust toolchain
+
+This project is written in Rust, and requires a stable toolchain to build. Additionally, `clippy` and `rustfmt` are used by CI jobs to ensure that patches are properly formatted and linted.
+
+You can obtain a Rust toolchain via many distribution methods, but the simplest way is via [rustup](https://rustup.rs/):
+
+```sh
+rustup component add clippy
+rustup component add rustfmt
+rustup install stable
+```
+
+## Build and test
+
+Building and testing is handled via `cargo` and `make`:
+
+```sh
+make build
+make check
+```
+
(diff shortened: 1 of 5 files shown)


docs/development/quickstart.md
```diff
@@ -0,0 +1,65 @@
+# Development quickstart
```
travier, line 1: Can you include:
```
---
layout: default
nav_order: 1
parent: Development
---
```
And probably update the order of the other pages?

travier (approved):
LGTM with a small suggestion.

docs/development/quickstart.md
```diff
@@ -0,0 +1,65 @@
...
+
+```sh
+make build
+make check
+```
+
+If you prefer running builds in a containerized environment, you can use the provided `Dockerfile`:
+
+```sh
+docker build -f contrib/Dockerfile.dev -t coreos/zincati:dev .
```
jlebon, line 44: There's also the cosa buildroot image which has cosa + build deps we could link to here. This is actually what CI here itself uses: https://github.com/coreos/zincati/blob/9d5d5ebe941151aff1d1b5efbbb816553de94f9b/.cci.jenkinsfile#L3 -> https://github.com/coreos/coreos-ci-lib/blob/76bbd3d46503dea956c59eb02a079302de0d3abb/vars/cosaPod.groovy#L8.

Nice to have a minimal image as well though given that the buildroot image is massive.

docs/development/quickstart.md
```diff
@@ -0,0 +1,65 @@
...
+mkdir test-image
+cd test-image
+cosa init https://github.com/coreos/fedora-coreos-config
+popd
+docker build -f contrib/Dockerfile.dev -t coreos/zincati:dev .
+docker run --rm -v "$(pwd):/source" -v "/tmp/test-image:/assembler" coreos/zincati:dev
+pushd /tmp/test-image
+cosa fetch
+cosa build
+```
```
jlebon, line 65: Could we link to https://coreos.github.io/coreos-assembler/working/#using-overrides here as well for the "pet container" case where cosa and build tools are in the same container? That's IMO the nicest UX for development.

jlebon (approved):
Some comments, but LGTM overall!

docs/development/quickstart.md
lucab: Ah, good catch, I totally forgot about front-matters here.

lucab:
@jlebon @travier thanks for the feedback, I've pushed a fixup commit on top, PTAL.

travier (approved):
LGTM

jlebon (approved):

Issue #420, opened 2020-12-07
metrics: values from runtime config could be set earlier


lucab:
From a race condition observed in https://github.com/coreos/fedora-coreos-tracker/issues/686.

Most metrics are initially set in an ad-hoc fashion as soon as execution reaches the relevant codepath. In particular, this is the case for `zincati_update_agent_updates_enabled` which is first set when the actor [is started](https://github.com/coreos/zincati/blob/v0.0.14/src/update_agent/actor.rs#L35). This effectively means there is a non-zero timespan when the metrics server is already active but the metrics is yet unset (i.e. not present in the response to the clients).

This is generally not a problem for Prometheus (which is able to track this "unset" state), but simpler consumers could be tricked by this racing behavior.

The general case is not fixable. But specifically for info metrics coming from configuration, this issues can be avoided by eliminating the window of possible races. As those values are known right after parsing the config, they can be directly set into relevant metrics before starting the metrics service.

lucab:
FYI @jlebon @dustymabe @bh7cw

PR #421, opened 2020-12-14
config: set metrics values earlier

lucab:
This moves metrics initialization early on, for those values that
are set from configuration settings and are already known at
startup.

Closes: https://github.com/coreos/zincati/issues/420


cgwalters (approved):
(semi-superficial) LGTM

Issue #351, opened 2020-09-03
identity: some enhancements for group configuration


lucab:
Zincati supports an `identity.group` setting, which is used both in Cincinnati and FleetLock contexts.

There are few minor enhancements that could be done around it:
 * describe it in docpages; the whole `identity` section is missing, it should have a dedicated page
 * check custom values against the FleetLock regexp to error out early in case of misconfigurations

PR #425, opened 2020-12-17
identity: validate group settings

lucab:
This adds a validity check for the group label, making sure that
user-customized settings (if any) are valid and can be used in
Cincinnati and FleetLock requests.

Closes: https://github.com/coreos/zincati/issues/351

### DIFF
--- a/docs/usage/agent-identity.md
+++ b/docs/usage/agent-identity.md
@@ -1,0 +1,45 @@
+---
+layout: default
+parent: Usage
+---
+
+# Agent identity
+
+Zincati agent tries to derive an unique identity for the machine it is running on, by introspecting the underlying OS and reading user configuration.
+This includes assigning an ID and a group label specific to the agent, so that cluster-wide upgrades can be orchestrated via [phased rollouts][phased] and [lock-based][fleetlock-strategy] reboots.
+
+[phased]: auto-updates.md#phased-rollouts-client-wariness-canaries
+[fleetlock-strategy]:  updates-strategy.md#lock-based-strategy
+
+## Identity configuration
+
+All agent identity values are normally auto-detected at startup, and do not require user intervention.
+
+However, the following settings can be overridden through configuration fragments in the `identity` section:
+ * `group`: group label, used for graph fetching ([Cincinnati][cincinnati]) and reboot orchestration ([FleetLock][fleetlock])
+ * `node_uuid`: agent ID, used for graph fetching ([Cincinnati][cincinnati]) and reboot orchestration ([FleetLock][fleetlock])
+ * `rollout_wariness`: agent wariness to [phased rollouts][phased], used for graph fetching ([Cincinnati][cincinnati]).
+
+The following are defaults for each setting:
+- `group` (group label) is set to `default`
+- `node_uuid` (agent ID) is automatically generated, by hashing `/etc/machine-id` content
+- `rollout_wariness` is unset and the Cincinnati backend will assign a dynamic value to each request
+
+When the agent ID is not customized via configuration fragments, its default value is dynamically generated starting from `/etc/machine-id` content and from a Zincati specific application ID.
+For more details about such application-specific machine IDs, see [machine-id][machine-id] documentation.
+
+[machine-id]: https://www.freedesktop.org/software/systemd/man/machine-id.html
+
+## Example
+
+As an example, users can specify custom identity parameters by writing a configuration fragment to `/etc/lib/zincati/config.d/90-custom-identity.toml`:
+
+```toml
+[identity]
+group = "workers"
+```
+
+The fragment above will steer the node into the "workers" reboot group.
+
+[cincinnati]: ../development/cincinnati/protocol.md
+[fleetlock]: ../development/fleetlock/protocol.md
--- a/docs/usage/agent-identity.md
+++ b/docs/usage/agent-identity.md
@@ -5,7 +5,7 @@ parent: Usage
 
 # Agent identity
 
-Zincati agent tries to derive an unique identity for the machine it is running on, by introspecting the underlying OS and reading user configuration.
+Zincati agent tries to derive a unique identity for the machine it is running on by introspecting the underlying OS and reading user configuration.
 This includes assigning an ID and a group label specific to the agent, so that cluster-wide upgrades can be orchestrated via [phased rollouts][phased] and [lock-based][fleetlock-strategy] reboots.
 
 [phased]: auto-updates.md#phased-rollouts-client-wariness-canaries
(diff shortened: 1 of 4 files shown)


lucab:
/cc @ashcrow @dghubble @rtlong

Some docs proof-reading would be appreciated.

docs/usage/agent-identity.md
```diff
@@ -0,0 +1,37 @@
+---
+layout: default
+parent: Usage
+---
+
+# Agent identity
+
+Zincati agent tries to derive an unique identity for the machine it is running on, by introspecting the underlying OS and reading user configuration.
```
ashcrow, line 8:
```suggestion
Zincati agent tries to derive a unique identity for the machine it is running on by introspecting the underlying OS and reading user configuration.
```

ashcrow: `a` versus `an` for words like `unique` is admittedly a difference between American English and British English.

docs/usage/agent-identity.md
```diff
@@ -0,0 +1,37 @@
...
+This includes assigning an ID and a group label specific to the agent, so that cluster-wide upgrades can be orchestrated via [phased rollouts][phased] and [lock-based][fleetlock-strategy] reboots.
+
+[phased]: auto-updates.md#phased-rollouts-client-wariness-canaries
+[fleetlock-strategy]:  updates-strategy.md#lock-based-strategy
+
+## Identity configuration
+
+All agent identity values are normally auto-detected at startup, and do not require user intervention.
```
ashcrow, line 16:
```suggestion
All agent identity values are normally auto-detected at startup and do not require user intervention.
```

docs/usage/agent-identity.md
```diff
@@ -0,0 +1,37 @@
...
+
+However, the following settings can be overridden through configuration fragments in the `identity` section:
+ * `group`: group label, used for graph fetching ([Cincinnati][cincinnati]) and reboot orchestration ([FleetLock][fleetlock])
+ * `node_uuid`: agent ID, used for graph fetching ([Cincinnati][cincinnati]) and reboot orchestration ([FleetLock][fleetlock])
+ * `rollout_wariness`: agent wariness to [phased rollouts][phased], used for graph fetching ([Cincinnati][cincinnati]).
+
+By default, the agent ID is automatically derived from `/etc/machine-id` in an application-specific way, group label is set to `default`, and rollout wariness is unset (Cincinnati backend will assign a dynamic value to each request).
```
ashcrow, line 23: Something like this may work a bit better.

```suggestion
The following are defaults for each setting:
- `group` (group label) is set to `default`
- `node_uuid` (agent ID) is automatically derived from `/etc/machine-id` in an application-specific way
- `rollout_wariness` is unset and the Cincinnati backend will assign a dynamic value to each request
```

ashcrow (commented):
Wording suggestions

docs/usage/agent-identity.md
lucab: I think I changed this at least three time before committing. Next time I'll just flip a coin and save some time ;)

kelvinfan001 (approved):
LGTM

lucab:
@kelvinfan001 no need to.

src/identity/mod.rs
```diff
@@ -154,6 +157,22 @@ impl Identity {
             stream: "mock-stable".to_string(),
         }
     }
+
+    /// Validate the group label value in current identity.
+    fn validate_group_label(&self) -> Fallible<()> {
+        static VALID_GROUP: &str = "^[a-zA-Z0-9.-]+$";
```
travier, line 163: Do we want to support groups starting with `.`, `-` or a number? i would recommend we only allow letters for the first char.

lucab: I wouldn't disallow them (especially numbers only), but not strong opinion either way. However, this is coming from an already-defined regex https://coreos.github.io/zincati/development/fleetlock/protocol/#body. I should probably add the reference in comments.

lucab: I've just added the specs reference in the docstring, but kept the regex as-is to match previous docs.

lucab:
Rebased, PTAL.

docs/usage/agent-identity.md
```diff
@@ -0,0 +1,40 @@
+---
+layout: default
+parent: Usage
+---
+
+# Agent identity
```
cgwalters, line 6: This doc is great!

docs/usage/agent-identity.md
```diff
@@ -0,0 +1,40 @@
...
+
+All agent identity values are normally auto-detected at startup and do not require user intervention.
+
+However, the following settings can be overridden through configuration fragments in the `identity` section:
+ * `group`: group label, used for graph fetching ([Cincinnati][cincinnati]) and reboot orchestration ([FleetLock][fleetlock])
+ * `node_uuid`: agent ID, used for graph fetching ([Cincinnati][cincinnati]) and reboot orchestration ([FleetLock][fleetlock])
+ * `rollout_wariness`: agent wariness to [phased rollouts][phased], used for graph fetching ([Cincinnati][cincinnati]).
+
+The following are defaults for each setting:
+- `group` (group label) is set to `default`
```
cgwalters, line 24: Any reason not to use the empty string instead?  (A user interface could render it as "(default)") or so?

cgwalters (approved):

docs/usage/agent-identity.md
lucab: The empty string is not a valid value. If unset in user configuration, the literal `default` value is sent to the FleetLock backend.

(yes, we could have made that field optional and let the backend auto-bucket clients without an explicit setting, when we defined the protocol. But we didn't foresee strong usecases for that, and now it's a bit late)

docs/usage/agent-identity.md
```diff
@@ -0,0 +1,40 @@
...
+- `node_uuid` (agent ID) is automatically derived from `/etc/machine-id` in an application-specific way
```
dghubble, line 25: When you say application-specific, who is the application here?

It would be useful if `node_uuid` exactly matched the machine id, since other systems like Kubernetes nodes use that as node metadata as well. Saying its "derived from" in an "application-specific" way seems pretty open ended.

lucab: The application here is Zincati. The term "application-specific" is lifted from `sd_id128_get_machine_app_specific` [docs](https://www.freedesktop.org/software/systemd/man/sd_id128_get_machine.html#), which is the implementation used under the hood. I should probably add this reference directly in the docpage.

The machine-id is considered confidential and should not be exposed on the network, see https://www.freedesktop.org/software/systemd/man/machine-id.html.

dghubble: Fair enough. In future, if a protocol implementer needed to associate Zincati identify with nodes I suppose a different mechanism would be needed.

lucab: The auto-generated default is meant to be enough to cover a scenario where the network flows are trusted and the node-id does not need to be cross-linked with other inventory systems.
Relaxing the former would need some kind of authN mechanism. 
For the latter, a fragment with `identity.node_uuid = <your_custom_id>` should already be enough to cover that.

lucab:
Rebased to master and updated some further wording on the docpage. CI is green, keeping previous stamps and merging.

Issue #94, opened 2019-08-02
update-agent: expose current state-machine state via sd_notify


miabbott:
I booted the FCOS 30.20190801.0 qemu image via `libvirt` and noticed that `zincati` logged a message about not being able to check for updates due to name resolution issues.  

```
$ journalctl -b -u zincati --no-pager
-- Logs begin at Fri 2019-08-02 17:38:50 UTC, end at Fri 2019-08-02 17:41:24 UTC. --
Aug 02 17:39:00 localhost systemd[1]: Started Zincati Update Agent.
Aug 02 17:39:00 localhost zincati[963]: [INFO ] starting update agent (zincati 0.0.3)
Aug 02 17:39:00 localhost zincati[963]: [INFO ] agent running on node '4346819196334e4780e3383378457d3d' in group 'default'
Aug 02 17:39:00 localhost zincati[963]: [ERROR] failed to check for updates: failed to query Cincinnati: https://updates.coreos.stg.fedoraproject.org/v1/graph?node_uuid=4346819196334e4780e3383378457d3d&platform=qemu&basearch=x86_64&group=default&current_os=a9c8d66d3628d1b9b4c4690777e8b730d08329b4359410cb410a2003296af1ca&stream=testing: error trying to connect: failed to lookup address information: Name or service not known
[core@localhost ~]$ ping -c 3 updates.coreos.stg.fedoraproject.org
PING stg.fedoraproject.org (209.132.181.5) 56(84) bytes of data.
64 bytes from stg.fedoraproject.org (209.132.181.5): icmp_seq=2 ttl=37 time=89.6 ms
64 bytes from stg.fedoraproject.org (209.132.181.5): icmp_seq=3 ttl=37 time=89.9 ms

--- stg.fedoraproject.org ping statistics ---
3 packets transmitted, 2 received, 33.3333% packet loss, time 35ms
rtt min/avg/max/mdev = 89.576/89.739/89.902/0.163 ms
```

Digging into the journal, it looks like `zincati` was running before the network was fully online.

https://paste.fedoraproject.org/paste/tuuMJ0BZ6nqmofnMvjrG7Q

I discussed this on #fedora-coreos with @jlebon who pointed out that the service probably retried, but never printed a message about successfully checking for an update.

So maybe the service could log a message when it successfully checks after a failed check?

Or reorder the service to wait for `network-online.target`?

jlebon:
With `-vvv`, you do see the subsequent try happen (and finish without errors):

```
[...]

PR #450, opened 2021-01-26
Register as update driver for rpm-ostree

kelvinfan001:
Following https://github.com/coreos/rpm-ostree/pull/2459, rpm-ostree
now has a mechanism for third-party update drivers to register as
its update driver.

Register Zincati as the update driver in the "initialization" stage.

### DIFF
--- a/src/rpm_ostree/cli_deploy.rs
+++ b/src/rpm_ostree/cli_deploy.rs
@@ -1,8 +1,11 @@
-//! Interface to `rpm-ostree deploy --lock-finalization`.
+//! Interface to `rpm-ostree deploy --lock-finalization` and
+//! `rpm-ostree deploy --register-driver`.
 
 use super::Release;
 use failure::{bail, Fallible, ResultExt};
 use prometheus::IntCounter;
+
+const DRIVER_NAME: &str = "Zincati";
 
 lazy_static::lazy_static! {
     static ref DEPLOY_ATTEMPTS: IntCounter = register_int_counter!(opts!(
@@ -19,7 +22,7 @@ lazy_static::lazy_static! {
 pub fn deploy_locked(release: Release, allow_downgrade: bool) -> Fallible<Release> {
     DEPLOY_ATTEMPTS.inc();
 
-    let result = invoke_cli(release, allow_downgrade);
+    let result = invoke_cli_deploy(release, allow_downgrade);
     if result.is_err() {
         DEPLOY_FAILURES.inc();
     }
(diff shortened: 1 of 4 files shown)


src/rpm_ostree/cli_deploy.rs
```diff
@@ -19,27 +23,42 @@ lazy_static::lazy_static! {
...
-fn invoke_cli(release: Release, allow_downgrade: bool) -> Fallible<Release> {
+fn invoke_cli(release: Release, allow_downgrade: bool, register_driver: bool) -> Fallible<Release> {
     fail_point!("deploy_locked_err", |_| bail!("deploy_locked_err"));
     fail_point!("deploy_locked_ok", |_| Ok(release.clone()));
 
     let mut cmd = std::process::Command::new("rpm-ostree");
     cmd.arg("deploy")
         .arg("--lock-finalization")
-        .arg(format!("revision={}", release.checksum))
+        .arg(release.checksum.to_string())
```
lucab, line 54: Hmm, it looks like this argument should be optional/conditional maybe?

src/rpm_ostree/cli_deploy.rs
```diff
@@ -19,27 +23,42 @@ lazy_static::lazy_static! {
...
+        checksum: String::from(""),
+        age_index: None,
+    };
+    invoke_cli(empty_release, false, true)?;
+    Ok(())
+}
+
 /// CLI executor.
-fn invoke_cli(release: Release, allow_downgrade: bool) -> Fallible<Release> {
+fn invoke_cli(release: Release, allow_downgrade: bool, register_driver: bool) -> Fallible<Release> {
```
lucab, line 47: It feels like this could be split into a `invoke_cli_deploy` plus a `invoke_cli_register`.

lucab:
Thanks, looks sane overall. I guess we'd have to wait for an rpm-ostree release with the new API before landing this.

lucab:
Additionally, can you check what's the resulting behavior when configuration contains https://coreos.github.io/zincati/usage/auto-updates/#disabling-auto-updates? Does `rpm-ostree status` output matches your expectation?

src/rpm_ostree/cli_deploy.rs
kelvinfan001: Sure! I was thinking that they share mostly the same code, but I guess forcing them together like this does feel a bit unnatural.

kelvinfan001:
> can you check what's the resulting behavior when configuration contains https://coreos.github.io/zincati/usage/auto-updates/#disabling-auto-updates?

My expectation is that we still consider Zincati to be the update driver even when Zincati's auto-updates logic is disabled by configuration. After https://github.com/coreos/zincati/issues/94, I'm planning on getting rpm-ostree to check the systemd status of its update driver and also display that status in `rpm-ostree status`. So when Zincati's present on the system but its auto-updates logic is disabled, `rpm-ostree status` might look something like this:
```
AutomaticUpdates: driven by Zincati
UpdateDriverStatus: auto-updates disabled
```

src/rpm_ostree/cli_deploy.rs
lucab: Yes, I got the same feeling while reading through. The fake `Release` item in particular is a good indicator that this arrangement could be re-calibrated.

lucab:
@kelvinfan001 that seems reasonable to me. When we finally get to glue all the pieces together, let's do a final manual test to make sure the UX matches the plan.

cgwalters:
v2021.2 is out https://github.com/coreos/rpm-ostree/releases/tag/v2021.2 and I'll add the overrides to FCOS next-devel soon.

src/update_agent/actor.rs
```diff
@@ -169,30 +169,49 @@ impl UpdateAgent {
...
-                log::warn!("{}", status);
-                actor.state.end();
-            };
-            update_unit_status(status);
-            notify_ready();
-            Ok(())
-        });
+                Self::register_as_driver(actor)
+            })
+            .map(|_r, actor, _ctx| {
```
kelvinfan001, line 198: I'm not sure if this is the right thing to do. We don't actually need to wait for the results of `local_deployments()` before we `register_as_driver()`, so chaining like this seems unnecessary. But OTOH, I'm not sure how to combine `local_deployments()` and `register_as_driver()` into one elegantly. 
@lucab

kelvinfan001: Hmm or does all of this not really matter since we're really just chaining all the actions within a tick into one big future..

kelvinfan001:
Minor change: we no longer register as the updates driver if auto-updates logic is disabled. This is so that rpm-ostree's CI can freely use `rpm-ostree upgrade`, `rpm-ostree deploy`, etc. without having to unregister/disable Zincati first (since I believe on non-production streams/dev machines Zincati is disabled only through Zincati's configuration, i.e. Zincati's initialization still happens).
@lucab Let me know if this logic doesn't make sense, though. Thanks!

lucab:
@kelvinfan001 that's an interesting scenario. Yes I think it is correct to not register as a driver in that case.

lucab (approved):

PR #454, opened 2021-01-28
update-agent: expose current state via sd_notify

kelvinfan001:
Closes: https://github.com/coreos/zincati/issues/94


tests/kola/libtest.sh
```diff
@@ -0,0 +1,73 @@
+# Source library for shell script tests
```
cgwalters, line 1: (At some point we should probably copy this into coreos-assembler; but definitely something after this PR)

tests/kola/misc/test-status.sh
```diff
@@ -0,0 +1,30 @@
...
+set -xeuo pipefail
+
+. ${KOLA_EXT_DATA}/libtest.sh
+
+cd $(mktemp -d)
+
+systemctl stop zincati.service
+echo "updates.enabled = false" > /etc/zincati/config.d/disable-in-test.toml
+systemctl start zincati.service
+systemctl status zincati.service | grep 'Status:' > zincati_disabled_status.txt
```
cgwalters, line 14: There's also e.g. `systemctl show -p StatusText zincati.service` which is more precise.

cgwalters (commented):
Gave this only a skim but it looks great to me!

src/update_agent/actor.rs
```diff
@@ -229,16 +263,23 @@ impl UpdateAgent {
         let can_finalize = self.strategy.can_finalize();
         let state_change = actix::fut::wrap_future::<_, Self>(can_finalize)
             .then(|can_finalize, actor, _ctx| actor.finalize_deployment(can_finalize, release))
-            .map(|res, actor, _ctx| res.map(|release| actor.state.update_finalized(release)));
+            .map(|res, actor, _ctx| {
+                res.map(|release| {
+                    update_unit_status("update finalized");
```
lucab, line 268: This could be augmented with the release version.

src/update_agent/actor.rs
```diff
@@ -210,8 +232,20 @@ impl UpdateAgent {
         let deploy_outcome = self.attempt_deploy(target);
         let state_change = deploy_outcome.map(move |res, actor, _ctx| {
             match res {
-                Ok(_) => actor.state.update_staged(release),
-                Err(_) => actor.deploy_attempt_failed(release),
+                Ok(_) => {
+                    actor.state.update_staged(release);
+                    update_unit_status("update staged");
```
lucab, line 237: This could be augmented with the release version.

tests/kola/misc/test-status.sh
```diff
@@ -0,0 +1,30 @@
...
+# Simple sanity checks for systemd unit status updates.
+
+set -xeuo pipefail
+
+. ${KOLA_EXT_DATA}/libtest.sh
+
+cd $(mktemp -d)
+
+systemctl stop zincati.service
+echo "updates.enabled = false" > /etc/zincati/config.d/disable-in-test.toml
```
lucab, line 12: Please add a numeric prefix to this, possibly `99-test-status-updates-enabled.toml`.

tests/kola/misc/test-status.sh
```diff
@@ -0,0 +1,30 @@
...
+systemctl start zincati.service
+systemctl status zincati.service | grep 'Status:' > zincati_disabled_status.txt
+assert_file_has_content zincati_disabled_status.txt \
+'initialization complete, auto-updates logic disabled by configuration'
+ok "status show initialization"
+
+systemctl stop zincati.service
+rm /etc/zincati/config.d/*
```
lucab, line 20: This seems a bit aggressive and may have other side-effects. Just overwriting your ow config fragment should be enough.

tests/kola/misc/test-status.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "updates.enabled = true" > /etc/zincati/config.d/disable-in-test.toml
+systemctl start zincati.service
+systemctl status zincati.service | grep 'Status:' > zincati_polling_status.txt
+assert_file_has_content zincati_polling_status.txt 'periodically polling for updates'
+ok "status show polling"
+# Give Zincati some time to check for updates.
+sleep 30
```
lucab, line 27: In order to reduce the test-run time, you may consider replacing this with a while loop with a more granular sleep, and assert on the file content once out of the loop.

src/update_agent/actor.rs
```diff
@@ -193,8 +203,20 @@ impl UpdateAgent {
...
+                        update_unit_status(&format!("update available: {}", release.version));
+                        actor.state.update_available(release);
+                    }
+                    None => {
+                        actor.state.no_new_update();
+                        update_unit_status(&format!(
+                            "periodically polling for updates (last checked {})",
+                            &actor
+                                .state_changed
+                                .with_timezone(&Local)
```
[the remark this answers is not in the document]
kelvinfan001: Hmm yeah I hadn't considered old/stale TZ data. I was thinking that this is more informational and for convenience, so it'd be nice if the time were more easily readable by the admin. (I originally tried to make it show something like "last checked x min ago", but I guess that can't be very easily done here. 
But UTC works for me :)

tests/kola/misc/test-status.sh
kelvinfan001: So for example if I use a local cosa build, then it'd have e.g. `90-disable-on-non-production-stream.toml` and `95-disable-on-dev.toml` right?

src/update_agent/actor.rs
```diff
@@ -143,11 +145,18 @@ impl UpdateAgent {
 
         let initialization = self.nop().map(|_r, actor, _ctx| {
             if actor.enabled {
-                log::info!("initialization complete, auto-updates logic enabled");
+                let status = "initialization complete, auto-updates logic enabled";
+                log::info!("{}", status);
+                update_unit_status(status);
+                notify_ready();
```
jlebon, line 151: Minor: seems like we could move those three lines to have the if-statement since they're in common with the else block? You'd have to declare `status` outside too.

src/update_agent/actor.rs
```diff
@@ -193,8 +203,20 @@ impl UpdateAgent {
             })
             .map(|res, actor, _ctx| {
                 match res {
-                    Some(release) => actor.state.update_available(release),
-                    None => actor.state.no_new_update(),
+                    Some(release) => {
+                        update_unit_status(&format!("update available: {}", release.version));
```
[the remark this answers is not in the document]
jlebon: Would `found update on remote: {}` fit better?

src/update_agent/actor.rs
```diff
@@ -327,3 +369,35 @@ impl UpdateAgent {
...
+/// configuration is loaded.
+fn notify_ready() {
+    match notify(false, &[NotifyState::Ready]) {
+        Err(e) => log::error!(
+            "failed to notify service manager that service is ready: {}",
+            e
+        ),
+        Ok(sent) => {
+            if !sent {
+                log::error!("status notifications not supported for this service");
```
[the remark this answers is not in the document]
kelvinfan001: hmm, this log message should only be sent when [`libsystemd::daemon::notify()`](https://docs.rs/libsystemd/0.2.1/libsystemd/daemon/fn.notify.html) returns `false` right? So shouldn't it log the same message in both functions, `update_unit_status` (the one above) and `update_unit_status`?

src/update_agent/actor.rs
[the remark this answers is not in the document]
jlebon: Personally, I find human-relative timestamps much more useful. UTC makes sense to start though ideally it'd be nice if we printed both (see e.g. https://github.com/coreos/rpm-ostree/blob/536bfc62ebb4d44b0126bed4e5050a08291614be/src/app/rpmostree-builtin-status.cxx#L1216).

src/update_agent/mod.rs
```diff
@@ -136,9 +136,9 @@ impl UpdateAgentState {
 
     /// Record a failed deploy attempt in UpdateAvailable state.
     ///
-    /// This returns whether a persistent deploy failure was detected
-    /// and the target update abandoned.
-    fn record_failed_deploy(&mut self) -> bool {
+    /// This returns the total number of failed deployment attempts
+    /// including the newly recorded failed attempt.
+    fn record_failed_deploy(&mut self) -> u8 {
```
jlebon, line 141: Hmm, I guess the function could return a tuple of `(bool, u8)` too so you don't have to redo the comparison in the callee.

jlebon (commented):
Nice work!

src/update_agent/actor.rs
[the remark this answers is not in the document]
kelvinfan001: Since currently we only update the unit status when we change states (instead of in a loop like rpm-ostree) and we don't know when users will call `systemctl status zincati.service` (unlike `rpm-ostree status`), is it still possible to do human-relative timestamps? 
Could we make rpm-ostree builtin-status extract the time from the StatusText of zincati.service and then compute the human-relative time when `rpm-ostree status` is called? ...but we also can't guarantee that all potential update drivers output their time in the same format with their StatusText.

src/update_agent/mod.rs
kelvinfan001: oh right, I forgot that was possible :)

tests/kola/misc/test-status.sh
lucab: In most cases you will end up with multiple fragments, yes. But only the highest priority setting will matter.

kelvinfan001: Haha thanks! I realized this right after I typed that comment.

src/update_agent/actor.rs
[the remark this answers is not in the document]
lucab: Right, which should be basically never happen. If it happens, it is likely printed twice quickly. If it is printed only once, it's weird and actually interesting to see which one is ok/not-ok.
Bottom line, this should be only useful for debugging weird cases, at which point having reducing all possible confusion is better.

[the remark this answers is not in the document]
kelvinfan001: Ah, thanks for pointing that out!

src/update_agent/actor.rs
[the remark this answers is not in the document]
jlebon: Yeah you're right, the architecture doesn't allow it very easily. I guess we could shove the timestamp as a u64 in a separate custom `sd_notify` variable that rpm-ostree picks up on, but it's not worth the complexity IMO.

Maybe we could print in both UTC and local time? But anyway, just UTC is fine for now!

kelvinfan001:
Thanks all for the review! @lucab  I've updated the code with your suggestions. Should be ready for another look :)

lucab (approved):
LGTM!

Issue #338, opened 2020-08-10
RFE: log an info level message when update is available, but present on disk


miabbott:
During the workshop session at Fedora Nest, the ability to rollback to a previous deployment was demonstrated.  As part of the demo, we showed that `zincati` was running but did not try to download, change deployments, or reboot the host.

I believe it would be useful if an INFO level message was logged during this situation, telling users that there is an update available but it is already on disk.

lucab:
I see the usability gain in surfacing this information. However Zincati performs such check every time it manages to fetch the graph from Cincinnati (i.e. on average every five minutes), so it doesn't look like a great idea spamming the logs each time.

I wonder whether we should:
 * print this information more succinctly only at startup (is it possible for new deployments to appear afterwards?)
 * put this information in metrics (only?)
 * lower the logging level below the default threshold (i.e. DEBUG or TRACE)

miabbott:
> I see the usability gain in surfacing this information. However Zincati performs such check every time it manages to fetch the graph from Cincinnati (i.e. on average every five minutes), so it doesn't look like a great idea spamming the logs each time.
> 
> I wonder whether we should:
> 
>     * print this information more succinctly only at startup (is it possible for new deployments to appear afterwards?)

I think logging at startup would be great.  There's a chance a user may manually rebase/deploy a new ostree commit manually, but I think that is a corner case that doesn't need to be addressed.

lucab:
Additional notes on this:
 * we don't know on start whether a locally-found deployment will (or will not) be a target for future updates, so the best we can do is printing how many and which has been found on startup (even if that includes older deployments). An `info` level message would work here.
 * the currently booted deployment is not interesting in this context, so care should be taken to **not** count that in.
 * the case where there aren't other deployments around is also not very interesting to a final user, so at most such case can be logged with a quick and dirty `debug` message.
 * as we are already querying rpm-ostree on each Cincinnati check, we should also put this gauged number into a metric, on top of printing and setting it on startup.

miabbott:
@lucab @bh7cw thanks for the additional thoughts around how to handle this RFE.  I agree with the notes above and it looks like a well thought out approach on how to better inform an end user in the situation.

PR #456, opened 2021-01-29
cincinnati: record ignored update targets in logs and metrics

kelvinfan001:
Zincati ignores possible update targets that have previously been
deployed. Log this explicitly so there's no confusion when e.g. users
do a rollback.

Closes: https://github.com/coreos/zincati/issues/338

### DIFF
--- a/src/cincinnati/mod.rs
+++ b/src/cincinnati/mod.rs
@@ -276,6 +276,16 @@ fn find_update(
     // Exclude target already deployed locally in the past.
     let new_updates = updates.difference(&local_releases);
 
+    // Log that we will avoid updating to already deployed releases.
+    let prev_deployed_excluded = updates.intersection(&local_releases).count();
+    if prev_deployed_excluded > 0 {
+        log::debug!(
+            "Found {} possible update target{} already deployed locally in the past; ignoring",
+            prev_deployed_excluded,
+            if prev_deployed_excluded > 1 { "s" } else { "" }
+        );
+    }
+
     // Pick highest available updates target (based on age-index).
     let next = match new_updates.last().cloned() {
         Some(rel) => rel,
--- a/src/cincinnati/mod.rs
+++ b/src/cincinnati/mod.rs
@@ -47,6 +47,10 @@ lazy_static::lazy_static! {
         "zincati_cincinnati_booted_release_is_deadend",
         "Whether currently booted OS release is a dead-end."
     ).unwrap();
+    static ref UPDATE_TARGETS_IGNORED: IntGauge = register_int_gauge!(
+        "zincati_cincinnati_ignored_update_targets",
+        "Number of ignored targets among update targets found."
+    ).unwrap();
     static ref UPDATE_CHECKS: IntCounter = register_int_counter!(opts!(
         "zincati_cincinnati_update_checks_total",
         "Total number of checks for updates to the upstream Cincinnati server."
@@ -285,6 +289,7 @@ fn find_update(
             if prev_deployed_excluded > 1 { "s" } else { "" }
         );
     }
+    UPDATE_TARGETS_IGNORED.set(prev_deployed_excluded as i64);
 
     // Pick highest available updates target (based on age-index).
     let next = match new_updates.last().cloned() {
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -138,17 +138,52 @@ impl UpdateAgent {
         std::time::Duration::from_secs(secs.saturating_add(jitter))
     }
 
+    /// Log at INFO level how many and which deployments will be excluded from being
+    /// future update targets.
+    fn log_excluded_depls(depls: &BTreeSet<Release>, actor: &UpdateAgent) {
+        // Exclude booted deployment.
+        let mut other_depls = depls.clone();
+        if !other_depls.remove(&actor.identity.current_os) {
+            log::error!("could not find booted deployment in deployments");
+            return; // Early return since this really should not happen.
+        }
+
+        let excluded_depls_count = other_depls.len();
+        if excluded_depls_count > 0 {
+            log::info!(
+                "found {} other finalized deployment{}",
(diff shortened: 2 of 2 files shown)


kelvinfan001:
I remember watching the FCOS tutorial from the last Nest with Fedora and this was a suggestion to Zincati. 
Before reading the code, I was also curious myself about whether Zincati is smart enough to not try to update to previously deployed deployments :)

lucab:
@kelvinfan001 yes, I think you may want to read https://github.com/coreos/zincati/issues/338 which should be the same topic. That ticket also contains some further discussion about not logging it at every check, but early on, to avoid spamming logs.
Although it could be a good idea keeping a live-metrics related to this, too.

kelvinfan001:
ah I see. Didn't realize there was already an issue for this. Lots of things I didn't consider.

src/update_agent/actor.rs
```diff
@@ -138,17 +144,48 @@ impl UpdateAgent {
         std::time::Duration::from_secs(secs.saturating_add(jitter))
     }
 
+    /// Log at INFO level how many and which deployments will be excluded from being
+    /// future update targets.
+    fn log_excluded_depls(deployments: Result<BTreeSet<Release>, ()>, actor: &UpdateAgent) {
+        if let Ok(depls) = deployments {
+            let excluded_depls_count = depls.len() - 1; // Minus current booted deployment.
```
lucab, line 151: This can be explicitly written as a `saturating_sub()`.

src/update_agent/actor.rs
```diff
@@ -138,17 +144,48 @@ impl UpdateAgent {
...
+            LOCAL_DEPLOYMENTS_COUNT.set(excluded_depls_count.try_into().unwrap());
```
lucab, line 152: Bunch of comments on this:
 * This should be periodically refreshed too, ideally close to the querying logic itself
 * Would a `as` work here instead?

src/update_agent/actor.rs
```diff
@@ -138,17 +144,48 @@ impl UpdateAgent {
         std::time::Duration::from_secs(secs.saturating_add(jitter))
     }
 
+    /// Log at INFO level how many and which deployments will be excluded from being
+    /// future update targets.
+    fn log_excluded_depls(deployments: Result<BTreeSet<Release>, ()>, actor: &UpdateAgent) {
```
lucab, line 149: This argument feels a bit weird, I think you should take a `&BTreeSet<Release>` here and wrap the calling site for this with a `if let Ok(ref depls) = foo() { log_excluded_depls(depls) }`.

src/update_agent/actor.rs
```diff
@@ -9,13 +9,19 @@ use libsystemd::daemon::*;
...
 use std::time::Duration;
 
 lazy_static::lazy_static! {
     static ref LAST_REFRESH: IntGauge = register_int_gauge!(opts!(
         "zincati_update_agent_last_refresh_timestamp",
         "UTC timestamp of update-agent last refresh tick."
     )).unwrap();
+
+    static ref LOCAL_DEPLOYMENTS_COUNT: IntGauge = register_int_gauge!(
+        "zincati_local_deployments_count",
```
lucab, line 22: I think this metrics really belongs to the `rpm-ostree` module, and could be prefixed as `zincati_rpm_ostree_status_[...]`.

src/update_agent/actor.rs
```diff
@@ -9,13 +9,19 @@ use libsystemd::daemon::*;
...
+        "Total number of local finalized deployments found, excluding the current booted deployment."
```
lucab, line 23: I'm not sure whether there is a better name for those. "Local inactive deployments" maybe?

/cc @jlebon

src/update_agent/actor.rs
kelvinfan001: ah yes, this helper function's argument is kind of weird, thanks!

kelvinfan001:
@lucab As per https://github.com/coreos/zincati/pull/472/files#r574551578, I think it might be more suitable to put the metrics-related stuff in a separate PR or in https://github.com/coreos/zincati/pull/472/ because I think doing metrics (re: rpm-ostree status) properly would require the changes in the other PR.

src/update_agent/actor.rs
jlebon: Hmm, do we even need "local"? All deployments are local. :)

Maybe

> Total number of deployments found, excluding any staged and booted one.

?

src/cincinnati/mod.rs
```diff
@@ -47,6 +47,10 @@ lazy_static::lazy_static! {
         "zincati_cincinnati_booted_release_is_deadend",
         "Whether currently booted OS release is a dead-end."
     ).unwrap();
+    static ref UPDATE_TARGETS_IGNORED: IntGauge = register_int_gauge!(
+        "zincati_cincinnati_update_targets_ignored_total",
```
lucab, line 51: This is not a total/counter, i.e. it does not accumulate values over time; `ignored_update_targets` would be more suitable. See https://prometheus.io/docs/practices/naming/ for guidance.

src/update_agent/actor.rs
```diff
@@ -138,17 +138,49 @@ impl UpdateAgent {
         std::time::Duration::from_secs(secs.saturating_add(jitter))
     }
 
+    /// Log at INFO level how many and which deployments will be excluded from being
+    /// future update targets.
+    fn log_excluded_depls(depls: &BTreeSet<Release>, actor: &UpdateAgent) {
+        let excluded_depls_count = depls.len().saturating_sub(1); // Minus current booted deployment.
```
lucab, line 144: It may be cleaner to just build a local vector excluding the booted deployment, and then use that for the rest of this function.

src/cincinnati/mod.rs
kelvinfan001: right, thanks!

lucab (approved):

Issue #462, opened 2021-02-03
Packaging fails due to recursive symlink


lucab:
```
   Packaging zincati v0.0.16 (/[...]/coreos/zincati)
error: failed to prepare local package for uploading

Caused by:
  failed to open for archiving: `/[...]/coreos/zincati/tests/kola/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data/misc/data`

Caused by:
  Too many levels of symbolic links (os error 40)
```

PR #463, opened 2021-02-03
tests/kola: fix recursive symlink

lucab:
This fixes packaging by removing the recursive symlink in kola tests.

Closes: https://github.com/coreos/zincati/issues/462


kelvinfan001 (approved):
Ah sorry about that. Didn't know this would break Rust packaging. In this case, we'll need to copy around `libtest.sh` if we create new subdirectories within `kola/`. 
LGTM.

Issue #137, opened 2019-09-18
rpm-ostree: optimize idle-state by caching status


lucab:
Followup from https://github.com/coreos/zincati/pull/112#discussion_r324343125.

Zincati is currently not very idle-friendly, as it continuously queries rpm-ostree for current status, causing the daemon to wake up from its idling state.

Suggestion is to cache the rpm-ostree status to optimize the idle-state, and only refresh the cache if anything changed on ostree side.

Plan is:
 * add a `status_cache` in the rpm_ostree actor
 * cache the current booted deployment plus additional local deployments in there, and attach the mtime to it
 * make querying a bit smarter, by always checking the [mtime on /ostree/deploy](https://github.com/ostreedev/ostree/blob/653fc6a125e789692d8f8cd47ca4b1a124766a61/src/libostree/ostree-sysroot.c#L271) first and using cached content if still fresh

PR #472, opened 2021-02-10
rpm-ostree: cache status to minimize daemon wakeups

kelvinfan001:
Zincati needs to invoke the rpm-ostree binary to get the current
deployments when starting up and checking for updates.
Try to minimize the number of times we wake up rpm-ostreed and
cache rpm-ostree's responses whenever possible.

Closes: https://github.com/coreos/zincati/issues/137


jlebon (approved):
Looks reasonable to me!

src/rpm_ostree/cli_status.rs
```diff
@@ -109,8 +118,31 @@ fn booted_json(status: StatusJSON) -> Fallible<DeploymentJSON> {
     Ok(booted)
 }
 
-/// Introspect deployments (rpm-ostree status).
-fn status_json(booted_only: bool) -> Fallible<StatusJSON> {
+/// Introspect deployments (rpm-ostree status) using rpm-ostree client actor client's
+/// cache if possible.
+fn status_json(client: &mut RpmOstreeClient, booted_only: bool) -> Fallible<StatusJSON> {
+    let ostree_depls_data = fs::metadata(OSTREE_DEPLS_PATH)?;
```
lucab, line 124: A `with_context()` with the file path would be helpful here.

src/rpm_ostree/cli_status.rs
```diff
@@ -1,18 +1,24 @@
...
+use super::actor::{RpmOstreeClient, StatusCache};
 use super::Release;
 use failure::{bail, ensure, format_err, Fallible, ResultExt};
+use filetime::FileTime;
+use log::trace;
 use serde::Deserialize;
 use std::collections::BTreeSet;
+use std::fs;
+
+const OSTREE_DEPLS_PATH: &str = "/ostree/deploy";
```
lucab, line 12: A docstring would be nice here, noting why we bypass rpm-ostree and directly go to ostree internals.

src/rpm_ostree/actor.rs
```diff
@@ -1,14 +1,25 @@
...
+pub struct StatusCache {
+    pub status: StatusJSON,
+    pub mtime: FileTime,
+}
+
 /// Client actor for rpm-ostree.
 #[derive(Debug, Default, Clone)]
-pub struct RpmOstreeClient {}
+pub struct RpmOstreeClient {
+    pub status_cache: Option<StatusCache>,
```
lucab, line 21: For reference, I believe that this works only because this actor is spawned on a single thread.
If we move to a n>1 threadpool, it will result in each actor-thread having its own cache.

There is no correctness issue in doing this and it isn't a real problem right now, so I think we can keep this. Perhaps we should add a quite note as a reminder, though.

src/rpm_ostree/cli_status.rs
```diff
@@ -109,8 +118,31 @@ fn booted_json(status: StatusJSON) -> Fallible<DeploymentJSON> {
...
+
+    if let Some(cache) = &client.status_cache {
+        if cache.mtime == ostree_depls_data_mtime {
+            trace!("cache fresh, using cached rpm-ostree status");
+            return Ok(cache.status.clone());
+        }
+    }
+
+    trace!("cache stale, invoking rpm-ostree to retrieve local deployments");
+    let status = invoke_cli_status(booted_only)?;
```
lucab, line 135: I'm mildly concerned by this, as we can technically cache a wrong/incomplete state depending on the boolean that gets here when priming the cache initially.

I'd rather switch to cache the full state and do the booted filtering ourselves, or query `rpm-ostree status` twice in here and cache the booted/non-booted details separately (depending on the ergonomics of the consumers that needs to be massaged).

lucab (commented):
Overlook looks nice. I have one concern on the caching logic, and a bunch of minor comments.

src/rpm_ostree/cli_status.rs
kelvinfan001: Thanks for pointing that out. I think we should always cache the full state in the `status_json()` function, since I don't think asking `rpm-ostree` to give us the full state would be a lot more expensive than only asking for `--booted`. Currently, I think whenever we use this `status_json()` function, we're always doing `booted_only=false` anyways.

cgwalters (approved):
This looks good to me, though I wonder if rpm-ostree should just serialize its status JSON into e.g. `/run/rpm-ostree/status.json` automatically; we could also add a systemd path unit to wakeup and refresh.  Then e.g. `rpm-ostree status --json` client side also wouldn't wake up the daemon.  But eh, this is totally fine too!

cgwalters:
Also if we did any more things in this area we'd probably benefit from https://github.com/coreos/rpm-ostree/issues/2389 since the logic for loading the cached status could be encapsulated there.

src/rpm_ostree/cli_status.rs
lucab: SGTM, it looks we can drop the `booted_only` args from the lower layers if that's true, and keep the filtering logic higher up.

While at it (or maybe later as you wish) you may also consider adding metrics regarding total-queries, failed-queries, and some deployments stats. It would probably ease some of the work in https://github.com/coreos/zincati/pull/456 too.

kelvinfan001: Yeah, maybe I'll create a separate PR for the metrics-related things, since I could see a lot of that would depend on the changes in this PR.

lucab (approved):

PR #474, opened 2021-02-16
rpm-ostree: expose metrics for status querying and caching

kelvinfan001:
Expose metrics for the number of deployments (excluding the booted
deployment and staged deployments). This is essentially the number
of deployments that are invalid candidates for targets of future
updates.

Also expose the total number of times Zincati attempted to query
local deployments (potentially utilizing the rpm-ostree actor's
cache), the total number of times `rpm-ostree status` was invoked,
and the total number of failed attempts to invoke `rpm-ostree status`.

### DIFF
--- a/src/rpm_ostree/cli_status.rs
+++ b/src/rpm_ostree/cli_status.rs
@@ -5,6 +5,7 @@ use super::Release;
 use failure::{bail, ensure, format_err, Fallible, ResultExt};
 use filetime::FileTime;
 use log::trace;
+use prometheus::IntCounter;
 use serde::Deserialize;
 use std::collections::BTreeSet;
 use std::fs;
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Not a zincati or Prometheus expert but this LGTM!

src/rpm_ostree/cli_status.rs
```diff
@@ -14,6 +15,25 @@ use std::fs;
 /// of the current state of deployments.
 const OSTREE_DEPLS_PATH: &str = "/ostree/deploy";
 
+lazy_static::lazy_static! {
+    static ref DEPLOYMENTS: IntGauge = register_int_gauge!(opts!(
+        "zincati_deployments_count",
```
kelvinfan001, line 20: Actually thinking about this a bit more, would this metric be useful? I think in most cases it would be `1`, since rpm-ostree saves 1 other deployment by default; but this metric doesn't really tell us any useful information, like for example whether there has likely been a rollback (this information would be more accurately predicted by the `zincati_cincinnati_update_targets_ignored_total` metric from https://github.com/coreos/zincati/pull/456/files).
I think that, at best, a high `zincati_deployments_count` could indicate to us that a user may be doing manual deployments with `rpm-ostree deploy` behind Zincati's back (or perhaps just configured rpm-ostree to save many deployments)? What do you think @lucab?

lucab: > but this metric doesn't really tell us any useful information

I agree with that, "how many target updates are we ignoring" is the real interesting signal.

src/rpm_ostree/cli_status.rs
```diff
@@ -14,6 +15,25 @@ use std::fs;
...
+        "Total number of deployments found, excluding any staged deployments and the currently booted deployment."
+    )).unwrap();
+    static ref STATUS_ATTEMPTS: IntCounter = register_int_counter!(opts!(
+        "zincati_status_attempts_total",
```
lucab, line 24: This prefix should be something like `zincati_rpm_ostree_status_cache_[...]`.

I think you may want to have two metrics here, in order to compute the cache-miss rate. You can do that by keeping a `cache_requests_total` plus a `cache_misses_total`, or a single `cache_requests_total{result=<hit|miss>}`.

This will also work as a double-check that in the steady-state we have almost zero cache misses, and all misses result in a status query and no failures.

kelvinfan001: I was thinking we can get an approximate cache-miss rate with `cache_requests_total` and `zincati_rpm_ostree_status_attempts_total`. But yeah, there are times when `rpm_ostree_status` is called directly without even attempting to check the cache (mainly because there isn't access to the actor), so `cache_requests_total{result=<hit|miss>}` would be better :)

src/rpm_ostree/cli_status.rs
kelvinfan001: I've removed that metric then

lucab (approved):

PR #478, opened 2021-02-22
update_agent: fix timestamp in status updates

lucab:
This fixes status updating logic, in order to display the timestamp
of the last check (instead of the last state change).

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -222,6 +222,11 @@ impl UpdateAgent {
         let state_change = self
             .local_deployments()
             .then(|res, actor, _ctx| {
+                let timestamp_now = chrono::Utc::now();
+                update_unit_status(&format!(
+                    "periodically polling for updates (last checked {})",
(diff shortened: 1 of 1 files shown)


kelvinfan001 (approved):
Hmm, I wonder how I got the idea to use the last state change time, since that would incorrectly display that e.g. the last update check was when we ~finished initialization~ reached 'steady'.
LGTM.

lucab:
@kelvinfan001 it's a subtle bug, because on the first poll the output is correct, it only goes out of sync on subsequent polls.
I guess you inferred it from the FSM graph, and to be honest it's debatable whether a reentering edge should be considered a state change (I remember that after some thoughts on this I decided it was more useful to ignore self-edges to that extent).

kelvinfan001:
> I decided it was more useful to ignore self-edges to that extent

Ohh, admittedly, I hadn't even realized that, but I went back to read the code a bit more carefully and confirmed that this is the case.
In any case, putting the timestamp closer to when we actually do a check would be much better (previously, even for the first and second polls, the output is slightly early).

Issue #114, opened 2019-08-29
agent: broadcast to consoles before rebooting


lucab:
This is a followup to https://github.com/coreos/fedora-coreos-tracker/issues/239.

Right now Zincati triggers a reboot in a silent way, which could surprise any logged users while they are in middle of some interactive maintenance.

For better user experience, it should somehow notify the users before triggering a reboot. Current proposal is to broadcast (i.e. `wall`) to all active consoles. This would match locksmith behavior.

lucab:
Blocked until:
 * https://github.com/nix-rust/nix/issues/1111 is fixed
 * a new version of `nix` crate is published with the fix

dustymabe:
+1 - I just had this idea today and there's already an open issue for it :heart:

lucab:
Small update here, this is unfortunately blocked on a couple more levels of dependencies that need to be patched too, see https://github.com/rust-lang/libc/pull/1496#issuecomment-529109067.

kelvinfan001:
Update: The implementation in #479 simply shells out to `loginctl` to get the list of logged in users. No longer requiring https://github.com/nix-rust/nix/issues/1111.

Issue #115, opened 2019-08-29
agent: delay reboot if ongoing interactive sessions


lucab:
This is a followup to https://github.com/coreos/fedora-coreos-tracker/issues/239.

Right now Zincati triggers a reboot in a silent way, which could surprise any logged users while they are in middle of some interactive maintenance.

For better user experience, it should somehow delay the reboot if it notices any active sessions. Current proposal is to count the number of active consoles before finalizing, and introduce a delay if the result is non-zero. This would match locksmith behavior.

Locksmith used to just sleep for a fixed amount of time. A better option to explore here would be implementing the [inhibitor-lock protocol](https://www.freedesktop.org/wiki/Software/systemd/inhibit/).

lucab:
This requires https://github.com/coreos/zincati/issues/114 to be implemented first.

openstacker:
@lucab  Now we(OpenStack Magnum) are using Fedora CoreOS as nodes for k8s cluster. But because of the auto update, we have to totally disable it now to avoid interrupting the k8s bootstrapping.  However, I would like to see a delayed reboot or manual reboot to leave the option to the owner of the server. Looking forward to seeing your thoughts. Cheers.

lucab:
Hi @openstacker! This specific ticket here is not what you are looking for. The topic however has been previously discussed at https://github.com/coreos/fedora-coreos-tracker/issues/147#issuecomment-606526951.

The approaches from there are: 
 1. have a your own first-boot bootstrapping unit to delay `boot-complete.target`
 2. use a temporary [shutdown inhibitor](https://www.freedesktop.org/software/systemd/man/systemd-inhibit.html)
 3. bring your own [reboot manager](https://github.com/coreos/zincati/blob/0.0.6/docs/usage/updates-strategy.md#lock-based-strategy)
 4. write a config fragment with `updates.enabled = false`, then remove it after bootstrapping and restart `zincati.service`
 5. implement and use https://github.com/coreos/zincati/issues/245 (this is not really targeted at your case)

jlebon:
@lucab Re. https://github.com/coreos/zincati/issues/115#issuecomment-645861581, can we document 1 and 2 in https://github.com/coreos/zincati/blob/master/docs/usage/auto-updates.md? This topic came up in the Nest with Fedora Hands-On Lab.

kelvinfan001:
[...]

PR #479, opened 2021-02-22
agent: Broadcast message to logged in users before rebooting

kelvinfan001:
Broadcast to users before reboot by writing a message to their
ttys.

Closes: #114

### DIFF
--- a/src/update_agent/mod.rs
+++ b/src/update_agent/mod.rs
@@ -9,7 +9,10 @@ use crate::rpm_ostree::{Release, RpmOstreeClient};
 use crate::strategy::UpdateStrategy;
 use actix::Addr;
 use chrono::prelude::*;
+use failure::{bail, Fallible, ResultExt};
 use prometheus::IntGauge;
+use serde::{Deserialize, Deserializer};
+use std::fs;
 use std::time::Duration;
 
 /// Default refresh interval for steady state (in seconds).
@@ -35,6 +38,28 @@ lazy_static::lazy_static! {
         "zincati_update_agent_updates_enabled",
         "Whether auto-updates logic is enabled."
     )).unwrap();
+}
+
+/// JSON output from `loginctl list-sessions --output=json`
+#[derive(Debug, Deserialize)]
+pub struct SessionsJSON {
+    user: String,
+    #[serde(deserialize_with = "empty_string_as_none")]
+    tty: Option<String>,
+}
+
+/// Function to deserialize field to `Option<String>`, where empty strings are
+/// deserialized into `None`.
+fn empty_string_as_none<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    if s.is_empty() {
+        Ok(None)
+    } else {
+        Ok(Some(s))
+    }
 }
 
 /// State machine for the agent.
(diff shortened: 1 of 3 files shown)


kelvinfan001:
This doesn't seem particularly useful on itself. Mainly for paving the way towards https://github.com/coreos/zincati/issues/115.

src/update_agent/mod.rs
```diff
@@ -238,6 +248,53 @@ impl UpdateAgent {
...
+        .with_context(|_| "failed to run 'loginctl' binary")?;
+
+    if !cmdrun.status.success() {
+        bail!(
+            "loginctl failed to list current sessions:\n{}",
+            String::from_utf8_lossy(&cmdrun.stderr)
+        );
+    }
+
+    let sessions: Vec<SessionsJSON> = serde_json::from_slice(&cmdrun.stdout)?;
```
lucab, line 268: It would be good to split the logic up to here to its own function, and pass the sessions vector as input to the rest of the logic.

src/update_agent/mod.rs
```diff
@@ -238,6 +248,53 @@ impl UpdateAgent {
...
+        tty.insert_str(0, "/dev/");
+
+        log::trace!(
+            "Attempting to broadcast a message to user {} at {}",
+            user,
+            tty
+        );
+
+        {
+            let broadcast = format!(
```
lucab, line 284: You can format this once outside of the loop.

src/update_agent/mod.rs
```diff
@@ -238,6 +248,53 @@ impl UpdateAgent {
...
+                "Broadcast message from Zincati at {}:\n{}\n",
+                chrono::Utc::now().format("%a %Y-%m-%d %H:%M:%S %Z"),
+                msg
+            );
+            fs::write(&tty, broadcast)?;
```
lucab, line 289: I'd rather re-arrange this so that in case of errors we do a `log::error!()` but we keep trying the other sessions.

At the end you can return a `Some(tty_written, tty_total)` or a `None` if there were no sessions.

src/update_agent/mod.rs
```diff
@@ -238,6 +248,53 @@ impl UpdateAgent {
...
+        }
+
+        sessions_count += 1;
```
lucab, line 292: This is a `saturating_add()`.

src/update_agent/mod.rs
kelvinfan001: I opted to return a Result still, since this would let us propagate any errors and also 'no sessions' doesn't seem like a special enough case to be `None` IMO. What do you think?

src/update_agent/mod.rs
```diff
@@ -238,6 +263,78 @@ impl UpdateAgent {
     }
 }
 
+/// Attempt to broadcast msg to all sessions registered in systemd's login manager.
+/// Returns a Result with a tuple of total sessions found and sessions broadcasted to,
+/// if no error.
+fn broadcast(msg: &str) -> Fallible<(u8, u8)> {
+    let sessions = get_user_sessions()?;
+    let sessions_total = sessions.len() as u8;
```
lucab, line 271: Keeping everything as `usize` seems fine to me. Any specific reasons for casting to `u8` here instead?

src/update_agent/mod.rs
```diff
@@ -238,6 +263,78 @@ impl UpdateAgent {
...
+    // Iterate over sessions and attempt to write to each session's tty.
+    for session in sessions.into_iter() {
+        let user = session.user;
+        let tty_dev = match session.tty {
+            Some(mut tty) => {
+                tty.insert_str(0, "/dev/");
+                tty
+            }
+            None => {
+                log::info!(
```
lucab, line 289: I think this can be demoted to `debug!()`. It seems to be a quite common scenario related to non-interactive users, and there is no need to spam the logs about that.

src/update_agent/mod.rs
lucab: That's fine as well.

src/update_agent/mod.rs
kelvinfan001: Thanks! I wasn't too familiar with _logged in_ non-interactive users, so I thought they were quite rare.

src/update_agent/mod.rs
kelvinfan001: No specific reason. I noticed instances of `u8`s used elsewhere in Zincati and I thought smaller is probably better.
(Just read up a bit on it though; in this case it should probably be smarter to let the compiler choose for us).

lucab (approved):

PR #481, opened 2021-02-22
update-agent: add some delay after first Cincinnati check

kelvinfan001:
Generally, state changes trigger immediate ticks; make an exception
for the case where we're transitioning from ReportedSteady to
NoNewUpdates. This will prevent Zincati from performing the first
two polls for updates consecutively with no delay in between.

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -113,7 +113,12 @@ impl UpdateAgent {
 
         // State changes trigger immediate tick/action.
         if discriminant(&prev_state) != discriminant(&self.state) {
-            return None;
+            // Unless we're transitioning from ReportedSteady to NoNewUpdate.
+            if !(prev_state == UpdateAgentState::ReportedSteady
+                && self.state == UpdateAgentState::NoNewUpdate)
+            {
+                return None;
+            }
         }
 
         let delay = match self.state {


src/update_agent/actor.rs
```diff
@@ -113,7 +113,12 @@ impl UpdateAgent {
 
         // State changes trigger immediate tick/action.
         if discriminant(&prev_state) != discriminant(&self.state) {
-            return None;
+            // Unless we're transitioning from ReportedSteady to NoNewUpdate.
+            if !(prev_state == UpdateAgentState::ReportedSteady
+                && self.state == UpdateAgentState::NoNewUpdate)
```
kelvinfan001, line 118: We can simplify this, but I feel like this is a bit more readable/intuitive, but let me know if I'm wrong.

kelvinfan001:
Also not sure if this was intentional, I just found the consecutive checks a bit odd when I noticed it.

lucab:
Good catch! No, it wasn't intentional.

However this ad-hoc condition seems a bit hacky. Perhaps a better alternative would be to cut the `ReportedSteady -> UpdateAavailable` edge, so that we are left with a single edge out of `ReportedSteady` into `NoNewUpdates`. At that point, the state transition will immediately trigger a Cincinnati check which could confirm that there are no new updates, or detect an update and move to the `UpdateAavailable` state.

kelvinfan001:
Cool, makes sense. Briefly looked over https://github.com/coreos/zincati/issues/229, and I think cutting the edge _shouldn't_ undo the work that was done in that ticket.

kelvinfan001:
That makes sense. But looking at https://github.com/coreos/zincati/issues/229; would cutting the edge  undo the work that was done in that ticket? If we cut that edge, it seems like we introduce a state of "don't actually know if there's an update available", except we just shift that state to `NoNewUpdates`, instead of the original `Steady`.

lucab:
Uhm, fair point.

lucab (approved):

PR #485, opened 2021-03-01
update_agent: delay reboot if ongoing interactive sessions

### LINKED ISSUE #115: agent: delay reboot if ongoing interactive sessions

kelvinfan001:
Allow update finalizations (reboots) to be postponed a number of times for an amount of time if it is detected that there are users (with a tty) currently logged in on the system. Add a counter for number of postponements to the UpdateStaged state of the update agent; if the maximum number of allowed postponements has been reached, proceed with the finalization, disregarding active users.

Closes #115


src/update_agent/actor.rs
```diff
@@ -381,27 +394,39 @@ impl UpdateAgent {
...
-                    log::warn!(
-                        "{} sessions found, but only broadcasted to {}",
-                        sessions_total,
-                        sessions_broadcasted
-                    );
+
+        // Postpone reboot if there are interactive users logged in.
+        match super::get_interactive_user_sessions() {
+            Ok(interactive_sessions) => {
+                DETECTED_ACTIVE_USERS.set(interactive_sessions.len().try_into().unwrap());
```
kelvinfan001, line 415: I recall in a previous PR, @lucab suggested casting the `usize` here `as i64`. However, wouldn't this cast potentially lead to data loss since `usize` is likely a `u64`? This time, I've opted to use `std::convert::TryInto::try_into()`, but let me know if `as` would actually work just fine.

lucab: Yes, it's a minor concern (you'll never see 2⁶³ user sessions), but technically right.
An `as` could be good enough.
Or you can write a small `saturating_as` helper which clamps the input to the proper output range, before doing the conversion.

src/update_agent/mod.rs
```diff
@@ -420,4 +530,49 @@ mod tests {
...
+        assert!(!should_postpone);
+        assert_eq!(machine, UpdateAgentState::UpdateStaged((update.clone(), 0)));
+
+        // Set up dummy interactive sessions.
+        let foo_session = InteractiveSession {
+            user: String::from("fakeuser"),
+            tty_dev: String::from("/dev/tty1"),
+        };
+        let interactive_sessions_present: Vec<InteractiveSession> = vec![foo_session];
+        // MAX-1 postponements.
```
kelvinfan001, line 558: Note to self: This should be `MAX postponements`, not `MAX-1`.

src/update_agent/actor.rs
```diff
@@ -381,27 +394,39 @@ impl UpdateAgent {
         release: Release,
     ) -> ResponseActFuture<Self, Result<Release, ()>> {
         if !can_finalize {
+            // Reset number of postponements if finalization attempt failed due to
```
lucab, line 397: There are at least two state transitions nested very deeply inside this function, and I think it would better to split the concerns so that in the top-level `tick_finalize_update` it is clearly visible that there three cases:
 * `strategy_can_finalize: false` -> reset the number of attempts
 * `strategy_can_finalize: true` -> handle interactive users, then:
   * `usersessions_can_finalize: false` -> update the number of attempts
   * `usersessions_can_finalize: true` -> call `finalize_deployment(true)`

src/update_agent/mod.rs
```diff
@@ -195,9 +222,70 @@ impl UpdateAgentState {
...
+    fn broadcast_reboot_warning(
+        &mut self,
+        postponements: u8,
+        release_ver: &str,
+        interactive_sessions: &[InteractiveSession],
+    ) {
+        let warning_msg;
+        if postponements == 0 {
+            let max_reboot_delay_sec =
+                (MAX_FINALIZE_POSTPONEMENTS as u64).saturating_mul(DEFAULT_POSTPONEMENT_TIME_SECS);
```
lucab, line 276: This feels a bit odd to read. Did you try inverting the direction of `postponements` counter, going from MAX to 0, so that it gives you directly the  correct number?

kelvinfan001: Hmm yeah that would spare us from having to `MAX_FINALIZE_POSTPONEMENTS`.saturating_sub(1)` and instead just check of whether `postponements_left == 1`.

src/update_agent/actor.rs
kelvinfan001: I've tried to address this in the latest WIP commit. I've made it a WIP commit because I feel like I actually made the code more difficult to read. It'd be great if you can look at the diff of just that commit and see if I'm going in the right general direction. Thanks!

src/update_agent/actor.rs
```diff
@@ -321,6 +348,36 @@ impl UpdateAgent {
         Box::pin(state_change)
     }
 
+    /// If there are interactive users logged in, return an error and postpone reboot.
+    fn check_interactive_sessions(
+        &mut self,
+        release: Release,
+    ) -> ResponseActFuture<Self, Result<Release, ()>> {
```
kelvinfan001, line 355: I think it'd be much clearer if I could return a `ResponseActFuture<Self, bool>` here, so then later, in the calling function (`tick_finalize_update()`), we don't need to `match` on a `Result` type to determine whether we should postpone the reboot. Essentially, we're communicating a `bool` through `Ok()`/`Err()`. 
The Actix-related bits were a bit tricky for me so I wasn't able to figure out how to do this.

src/update_agent/actor.rs
```diff
@@ -308,9 +308,36 @@ impl UpdateAgent {
...
+                    // No interactive sessions detected.
+                    Ok(release) => {
+                        log::info!(
+                            "staged deployment '{}' available, proceeding to finalize it",
+                            release.version
+                        );
+                        actor.finalize_deployment(release)
+                    }
+                    // Interactive sessions detected or strategy does not permit finalization.
+                    Err(_) => Box::pin(actix::fut::err(())),
```
kelvinfan001, line 338: This could have happened due to `!strategy_can_finalize` OR `check_interactive_sessions()` returned an error (there are interactive users and a reboot should be postoned).

kelvinfan001:
Restructured a lot of the original code and addressed comments. Ready for another round of review :)

/cc @lucab

src/update_agent/mod.rs
```diff
@@ -433,4 +584,73 @@ mod tests {
...
+
+    #[test]
+    fn test_format_seconds() {
+        assert_eq!("1 second", format_seconds(1));
+        assert_eq!("2 seconds", format_seconds(2));
+        assert_eq!("1 minute", format_seconds(60));
+        assert_eq!("1 minute and 1 second", format_seconds(61));
+        assert_eq!("1 minute and 30 seconds", format_seconds(90));
+        assert_eq!("2 minutes", format_seconds(120));
+        assert_eq!("42 minutes and 23 seconds", format_seconds(2543));
```
jlebon, line 654: It's easier to believe these are correctly written if you write e.g. 2543 as `42*60 + 23`. :)

kelvinfan001: haha, right, just `2543` really isn't too useful there :)

src/update_agent/mod.rs
```diff
@@ -38,16 +47,31 @@ lazy_static::lazy_static! {
         "zincati_update_agent_updates_enabled",
         "Whether auto-updates logic is enabled."
     )).unwrap();
+    static ref POSTPONED_FINALIZATIONS: IntCounter = register_int_counter!(opts!(
+        "zincati_update_agent_postponed_finalizations_total",
+        "Total number of update finalization postponements due to active users."
+    )).unwrap();
+    static ref DETECTED_ACTIVE_USERS: IntGauge = register_int_gauge!(opts!(
+        "zincati_update_agent_detected_active_users",
```
lucab, line 55: Minor suggestion: something like `finalization_detected_active_users` makes it easier to understand when this is relevant.

src/update_agent/mod.rs
```diff
@@ -198,9 +228,86 @@ impl UpdateAgentState {
...
+    ///
+    /// Returns a boolean indicating whether finalization is permitted.
+    fn handle_interactive_sessions(&mut self, interactive_sessions: &[InteractiveSession]) -> bool {
+        if interactive_sessions.is_empty() {
+            return true;
+        }
+
+        let (release, postponements_remaining) = match self.clone() {
+            UpdateAgentState::UpdateStaged((r, p)) => (r, p),
+            _ => unreachable!(
```
lucab, line 268: Self-note: later on we should probably try to re-arrange the internals of the FSM data-structure so that it doesn't require this matching.

lucab (approved):

PR #490, opened 2021-03-04
rpmostree: Use Rc for status caching

cgwalters:
Prep for updating to use the rpm-ostree client bindings; I hit
the problem that the current struct doesn't implement `Clone`.

I could have easily added that, but conceptually it's a bit wrong
to have our cache gather some data, immediately do a full copy,
then discard the original.

I tried fighting the borrow checker here to have the function
return a reference; doesn't seem possible.  In the end with this
type of cache I think it's just cleaner to use reference counting;
there's no concerns about circular references etc.

### DIFF
--- a/src/rpm_ostree/cli_status.rs
+++ b/src/rpm_ostree/cli_status.rs
@@ -9,6 +9,7 @@ use prometheus::IntCounter;
 use serde::Deserialize;
 use std::collections::BTreeSet;
 use std::fs;
+use std::rc::Rc;
 
 /// Path to local OSTree deployments. We use its mtime to check for modifications (e.g. new deployments)
 /// to local deployments that might warrant querying `rpm-ostree status` again to update our knowledge
@@ -143,9 +144,8 @@ fn booted_json(status: &StatusJSON) -> Fallible<DeploymentJSON> {
     Ok(booted)
 }
 
-/// Introspect deployments (rpm-ostree status) using rpm-ostree client actor client's
-/// cache if possible.
-fn status_json(client: &mut RpmOstreeClient) -> Fallible<StatusJSON> {
+/// Ensure our status cache is up to date; if empty or out of date, run `rpm-ostree status` to populate it.
+fn status_json(client: &mut RpmOstreeClient) -> Fallible<Rc<StatusJSON>> {
     STATUS_CACHE_ATTEMPTS.inc();
     let ostree_depls_data = fs::metadata(OSTREE_DEPLS_PATH)
         .with_context(|e| format_err!("failed to query directory {}: {}", OSTREE_DEPLS_PATH, e))?;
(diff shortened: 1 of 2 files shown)


src/rpm_ostree/cli_status.rs
```diff
@@ -143,24 +144,23 @@ fn booted_json(status: &StatusJSON) -> Fallible<DeploymentJSON> {
...
             return Ok(cache.status.clone());
         }
     }
 
     STATUS_CACHE_MISSES.inc();
     trace!("cache stale, invoking rpm-ostree to retrieve local deployments");
-    let status = invoke_cli_status(false)?;
+    let status = Rc::new(invoke_cli_status(false)?);
     client.status_cache = Some(StatusCache {
         status: status.clone(),
```
cgwalters, line 165: Note that while there's still a `clone()` call here that boils down to (inlined) an integer increment, not a full `memcpy()` of the status data

src/rpm_ostree/cli_status.rs
```diff
@@ -143,24 +144,23 @@ fn booted_json(status: &StatusJSON) -> Fallible<DeploymentJSON> {
     Ok(booted)
 }
 
-/// Introspect deployments (rpm-ostree status) using rpm-ostree client actor client's
-/// cache if possible.
-fn status_json(client: &mut RpmOstreeClient) -> Fallible<StatusJSON> {
+/// Ensure our status cache is up to date; if empty or out of date, run `rpm-ostree status` to populate it.
+fn status_json(client: &mut RpmOstreeClient) -> Fallible<Rc<StatusJSON>> {
```
cgwalters, line 148: I think it's really neat how in Rust one can change a function like this to return an `Rc<>` and for all the callers it's transparent due to the auto-deref.

kelvinfan001: It took some getting used to. I remember earlier on when I almost forgot that `&` are just references instead of a magical symbol used for borrowing stuff.

src/rpm_ostree/cli_status.rs
kelvinfan001: Would this be more obvious if we used `Rc::clone(&status)`?

kelvinfan001 (approved):
LGTM!
> gather some data, immediately do a full copy,
then discard the original.

Now that I think about it, I think I might've done this in other places as well just to appease functions that like to take ownership of its arguments...

src/rpm_ostree/cli_status.rs
lucab: https://rust-lang.github.io/rust-clippy/master/#clone_on_ref_ptr (not enabled by default because it has too many false positives)

cgwalters: I see the value in the lint; updated to use `Rc::clone`.

PR #491, opened 2021-03-04
rpm-ostree: Use upstream Rust bindings

cgwalters:
rpm-ostree: Use upstream Rust bindings

Depends: coreos/rpm-ostree#2636

Use the upstream client bindings for the status data.  Note
that some functions switched to returning borrowed `&String`
because there's no need to clone unnecessarily.

An interesting note: In the upstream API, the commit
metadata is exposed as a generic `HashMap<>` because
that's how it's used by both ostree and rpm-ostree.  It'd
be a layering violation for us to hardcode `coreos-assembler.basearch`
in the rpm-ostree git for example, not to mention `fedora-coreos.stream`.
So those constants stay here in zincati.

I dropped the `_json` terminology from various functions
because that's weird - we parsed the data from JSON, but
that's not really very relevant except as an implementation
detail.  It's just a `Deployment`, not a `DeploymentJSON`.

Also I dropped for now the optimization of using `--booted`;
it's not a huge amount of data.  If we care we can re-add that
later.  Note it makes caching more complex because then we
need to carefully check the booted state too.

### DIFF
--- a/src/rpm_ostree/cli_status.rs
+++ b/src/rpm_ostree/cli_status.rs
@@ -2,14 +2,15 @@
 
 use super::actor::{RpmOstreeClient, StatusCache};
 use super::Release;
-use failure::{bail, ensure, format_err, Fallible, ResultExt};
+use failure::{format_err, Fallible, ResultExt};
 use filetime::FileTime;
 use log::trace;
 use prometheus::IntCounter;
-use serde::Deserialize;
 use std::collections::BTreeSet;
 use std::fs;
 use std::rc::Rc;
+
+use super::CLI_CLIENT;
 
 /// Path to local OSTree deployments. We use its mtime to check for modifications (e.g. new deployments)
 /// to local deployments that might warrant querying `rpm-ostree status` again to update our knowledge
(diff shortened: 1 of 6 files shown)


cgwalters:
Draft until the rpm-ostree PR merges.  I also only compile tested this.

src/rpm_ostree/cli_status.rs
```diff
@@ -36,82 +39,70 @@ lazy_static::lazy_static! {
...
-            checksum: self.base_revision(),
-            version: self.version,
+            checksum: d.base_checksum.clone().unwrap_or_else(|| d.checksum.clone()),
+            version: d.version.clone().unwrap_or_default(),
             age_index: None,
         }
     }
-
-    /// Return the deployment base revision.
-    pub fn base_revision(&self) -> String {
```
cgwalters, line 80: This might be a handy API to add upstream.

src/rpm_ostree/cli_status.rs
```diff
@@ -36,82 +39,70 @@ lazy_static::lazy_static! {
...
 }
 
 /// Parse base architecture for booted deployment from status object.
-pub fn parse_basearch(status: &StatusJSON) -> Fallible<String> {
-    let json = booted_json(status)?;
-    Ok(json.base_metadata.basearch)
+pub fn parse_basearch(status: &rpmostree_client::Status) -> Fallible<&String> {
+    let d = booted_deployment(status)?;
+    let v = d.base_commit_meta.get(COSA_BASEARCH);
+    if let Some(basearch) = v {
```
cgwalters, line 56: Maybe extract a helper function `find_nonempty_string_key()` or so?

src/rpm_ostree/cli_status.rs
```diff
@@ -36,82 +39,70 @@ lazy_static::lazy_static! {
...
+    } else {
+        failure::bail!("No {} metadata key", COSA_BASEARCH);
+    }
 }
 
 /// Parse the booted deployment from status object.
-pub fn parse_booted(status: &StatusJSON) -> Fallible<Release> {
-    let json = booted_json(status)?;
-    Ok(json.into_release())
+pub fn parse_booted(status: &rpmostree_client::Status) -> Fallible<Release> {
```
cgwalters, line 72: Hm, this should probably be `find_booted()` (we're not parsing anything now) and potentially be upstream API too.

src/rpm_ostree/cli_status.rs
```diff
@@ -170,41 +159,26 @@ fn status_json(client: &mut RpmOstreeClient) -> Fallible<StatusJSON> {
 }
 
 /// CLI executor for `rpm-ostree status --json`.
-pub fn invoke_cli_status(booted_only: bool) -> Fallible<StatusJSON> {
+pub fn invoke_cli_status() -> Fallible<rpmostree_client::Status> {
     RPM_OSTREE_STATUS_ATTEMPTS.inc();
 
-    let mut cmd = std::process::Command::new("rpm-ostree");
-    cmd.arg("status").env("RPMOSTREE_CLIENT_ID", "zincati");
```
cgwalters, line 177: We lose this client ID - need to add that to upstream API.  It doesn't matter much for status querying but it will for other work.

src/rpm_ostree/cli_status.rs
```diff
@@ -170,41 +159,26 @@ fn status_json(client: &mut RpmOstreeClient) -> Fallible<StatusJSON> {
...
-    let status: StatusJSON = serde_json::from_slice(&cmdrun.stdout)?;
-    Ok(status)
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
 
-    fn mock_status(path: &str) -> Fallible<StatusJSON> {
+    fn mock_status(path: &str) -> Fallible<rpmostree_client::Status> {
```
cgwalters, line 178: It's not clear to me we even need these tests anymore, since the upstream library has this.

cgwalters:
OK rebased :surfer: and cleaned up to use the latest upstream client API.  I think this turns out cleaner.  Only compile tested though.

cgwalters:
OK so the `cargo publish --dry-run` is going to fail with this.  Is it really worth publishing zincati to crates.io?  It doesn't make sense to `cargo install` it, etc.

src/rpm_ostree/cli_status.rs
```diff
@@ -37,83 +38,31 @@ lazy_static::lazy_static! {
...
+impl From<&rpmostree_client::Deployment> for Release {
+    fn from(d: &rpmostree_client::Deployment) -> Self {
         Release {
-            checksum: self.base_revision(),
-            version: self.version,
+            checksum: d
+                .base_checksum
+                .clone()
+                .unwrap_or_else(|| d.checksum.clone()),
+            version: d.version.clone().unwrap_or_default(),
```
lucab, line 48: This looks sketchy, and may break other logic in the app.

I don't know in which cases we can end up with a version-less commit, but it may be good to log a warning and filter it out.

src/rpm_ostree/cli_status.rs
```diff
@@ -226,18 +147,4 @@ mod tests {
             assert_eq!(deployments.len(), 1);
         }
     }
-
-    #[test]
-    fn mock_booted_basearch() {
```
lucab, line 231: These two are some basic unit tests to ensure we don't regress on parsing FCOS-specific metadata, I'd rather not drop it.

src/rpm_ostree/mod.rs
```diff
@@ -16,6 +17,16 @@ use failure::{ensure, format_err, Fallible, ResultExt};
 use serde::Serialize;
 use std::cmp::Ordering;
 
+/// Metadata key the base (RPM) architecture; injected by coreos-assembler
```
lucab, line 20:
```suggestion
/// Metadata key for the base (RPM) architecture, injected by coreos-assembler.
```

lucab:
The crates.io tarball is the main artifact consumed by the [Fedora specfile](https://src.fedoraproject.org/rpms/rust-zincati/blob/rawhide/f/rust-zincati.spec#_15).

We can back away from that by adjusting the release steps. But that would introduce maintenance skew compared to the other (coreos+rust) projects we maintain, and it would still be blocked by `cargo package`.

Issue #95, opened 2019-08-05
agent: should warn when run by other users


arithx:
When `zincati agent` is ran by other users (e.x. `core`) instead of the `zincati` user it will fail with a `Permission Denied` error. We should detect the user and warn if it's not correct.

This is what the error looks like when being ran from the `core` user:
```
$ zincati -v agent
[INFO ] starting update agent (zincati 0.0.3)
[INFO ] agent running on node 'c5673158671d4c5996b8c1d0da025ca6' in group 'default'
Error: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }

failed to bind metrics service
```

NOTE: I stumbled across this while just playing around with the OS and saw the binary.

lucab:
I agree with this.

The agent should be checking that it is running as the `zincati` user.
The proper place where to put this check would be within [`run_agent`](https://github.com/coreos/zincati/blob/0.0.4/src/main.rs#L53).
Helper library method to use is https://docs.rs/users/0.9.1/users/fn.get_current_username.html.

PR #495, opened 2021-03-08
cli: Check if run as `zincati` user

kelvinfan001:
Error out if not run by the `zincati` user for the `agent` and
`deadend-motd` subcommands, since the update agent should only be
running as the `zincati` user and `deadend-motd` should not be used
externally.
The binary is currently already placed in `/usr/libexec` in Fedora, but doing
this for extra safety and clarity.

Closes: https://github.com/coreos/zincati/issues/95

Example output:
```
$ /usr/libexec/zincati agent
[WARN ] zincati binary should not be run directly
[ERROR] critical error: update agent not running as `zincati` user

```
```
$ /usr/libexec/zincati deadend-motd unset
[WARN ] zincati binary should not be run directly
[ERROR] critical error: deadend-motd subcommand must be run by `zincati` user
```

### DIFF
--- a/src/cli/deadend.rs
+++ b/src/cli/deadend.rs
@@ -1,5 +1,6 @@
 //! Logic for the `deadend` subcommand.
 
+use super::ensure_zincati_user;
 use failure::{bail, Fallible, ResultExt};
 use std::fs::Permissions;
 use std::io::Write;
@@ -28,6 +29,7 @@ pub enum Cmd {
 impl Cmd {
     /// `deadend-motd` subcommand entry point.
     pub(crate) fn run(self) -> Fallible<()> {
+        ensure_zincati_user("deadend-motd subcommand must be run by `zincati` user")?;
         match self {
             Cmd::Set { reason } => refresh_motd_fragment(reason),
             Cmd::Unset => remove_motd_fragment(),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2109,6 +2109,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "users"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "24cc0f6d6f267b73e5a2cadf007ba8f9bc39c6a6f9666f8cf25ea809a153b032"
+dependencies = [
+ "libc",
+ "log",
+]
+
+[[package]]
 name = "uuid"
 version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2364,4 +2374,5 @@ dependencies = [
  "tokio",
  "toml",
  "url",
+ "users",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ structopt = "0.3"
 tokio = "0.2"
 toml = "0.5"
 url = { version = "2.2", features = ["serde"] }
+users = "0.11.0"
 
 [dev-dependencies]
 http = "0.2"
--- a/src/cli/agent.rs
+++ b/src/cli/agent.rs
@@ -1,5 +1,6 @@
 //! Logic for the `agent` subcommand.
 
+use super::ensure_zincati_user;
 use crate::{config, metrics, rpm_ostree, update_agent};
 use actix::Actor;
 use failure::{Fallible, ResultExt};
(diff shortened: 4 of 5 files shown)


cgwalters:
Looks OK to me, though I think a stricter check is "running under systemd".  We could implement that by adding e.g. `Environment=ZINCATI_SYSTEMD=1` to the unit file and verify that.  It wouldn't require any additional crates.

cgwalters:
> Just for my information, why is adding an environment variable stricter than checking the user? Isn't it easier to add an environment variable to a command than to run the command as another user?

Oh for sure, but we're just trying to protect against running it *accidentally* right?

As for it being stricter...eh, it's indeed debatable.

So anyways again this LGTM overall!

kelvinfan001:
> Oh for sure, but we're just trying to protect against running it accidentally right?

Yep, makes sense. Thanks :)

cgwalters (approved):

lucab:
> Looks OK to me, though I think a stricter check is "running under systemd".

Not a bad idea. More specifically, we can check that the agent got spawned with an invocation ID https://www.freedesktop.org/software/systemd/man/systemd.exec.html#%24INVOCATION_ID.

src/cli/deadend.rs
```diff
@@ -28,6 +29,7 @@ pub enum Cmd {
 impl Cmd {
     /// `deadend-motd` subcommand entry point.
     pub(crate) fn run(self) -> Fallible<()> {
+        ensure_zincati_user("deadend-motd subcommand must be run by `zincati` user")?;
```
lucab, line 32: Uhm, this doesn't look correct though. This helper is run through `pkexec`, so I would expect this to always run as `root`. Did you check this either manually or via kola tests?

kelvinfan001: ah I see, my bad. I only checked that the command couldn't be run as my normal user. Hmm I think the invocation ID environment variable won't work here either because `pkexec` doesn't pass on environment variables in the `exec`'ed program.

kelvinfan001: Could we get `pkexec` to execute as the `zincati` user then?

lucab: No, wait, those are separate concerns:
 * `zincati agent` runs as `zincati` user under a systemd service unit
 * `zincati deadend-motd` runs as `root` user under pkexec

Their respective setup is already ok, it's only this specific check which is wrong. The original ticket was correctly scoped to the `agent` verb only, but the PR dragged the check into an additional place where it doesn't belong.

kelvinfan001: The problem with this PR (the `deadend-motd` check) is that it'll fail even when it was the `zincati agent` (running as the `zincati` user) that `pkexec`'ed `/usr/libexec/zincati deadend-motd` right? The check will cause `/usr/libexec/zincati deadend-motd` to exit because it sees that it is running as `root`. So I thought that if we made `pkexec` run as `zincati` user, then the happy path would work; and we'll still be able to exit early if a user tries to run `/usr/libexec/zincati deadend-motd`.

kelvinfan001: Or we could just skip the check for `zincati deadend-motd` and if a user tries to run that subcommand, then s/he'll get 
```
$ /usr/libexec/zincati deadend-motd set --reason foo
[ERROR] critical error: failed to create temporary MOTD file under '/run/motd.d/'
[ERROR]  -> Permission denied (os error 13) at path "/run/motd.d/.deadend.acAhvM.motd.partial"
```
(I think this is what you mean right?)

lucab: Yes, we can either drop the check at all (previous status quo), or fix it to match the real expected user.

cgwalters: Sorry about this BTW!  I really went back and forth with myself for a minute before clicking the merge button.  My thought process flowed like: "I don't know zincati really well, but this patch looks sane.  And I think it's covered by CI tests.  Should it wait for luca?  It could, but on the other hand it's good to build up a sense of shared ownership in projects and not have one person every PR waits for.  Hmm, well if in the end something isn't right he should see this and be able to review post-merge."

Which is basically what happened :smile:  Curious what you all think about this.

kelvinfan001: To be fair, I really should've understood what `pkexec` does first... apologies for that. This PR also unfortunately wasn't covered by CI (working on that now).
I think having non-Luca people review (and approve) smaller PRs like this one makes sense, so we don't spam him with every little thing. 
But I'll defer to Luca on whether we should defer to Luca for everything.

lucab: No worries, I think the flow here was fine, and as Kelvin said we mostly hit a CI gap.

Overall kola does cover the critical path, so in the worst case a green merge should at most regress some minor/UX feature, but the core upgrading logic should be fine.

PR #503, opened 2021-03-16
tests/kola: add dead-end MOTD kola test

kelvinfan001:
tests/kola: add dead-end MOTD kola test

  Create a new kola test directory that contains a Fedora CoreOS
  config that sets up an HTTP server and configure Zincati to use
  it as its Cincinnati server.
  Include a test to check that Zincati correctly detects a deadend
  release and successfully generates a MOTD.

docs/development: add testing docs

### DIFF
--- a/docs/development/testing.md
+++ b/docs/development/testing.md
@@ -1,0 +1,45 @@
+---
+layout: default
+nav_order: 7
+parent: Development
+---
+
+# Testing
+
+## Unit Tests
+Unit tests can be run using `make check` (via `cargo test`).
+
+## External Kola Tests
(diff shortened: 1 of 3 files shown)


kelvinfan001:
I believe CI is failing because `kolet` does not yet have the `httpd` command and so the mock Cincinnati server is not started in the newly added test.
`httpd` command added here: https://github.com/coreos/coreos-assembler/pull/2087
Blocked until new CoreOS Assembler release.

lucab (approved):

lucab:
Thanks for the test and for the docs! I guess this can be freely merged now, right?

PR #514, opened 2021-03-24
Add new actor for D-Bus server interface

kelvinfan001:
This adds an experimental D-Bus server interface at `org.coreos.zincati.Experimental` for Zincati.
It includes:
- a `Moo` method purely for testing to demonstrate that a new `dbus` actor running on its own thread can handle incoming D-Bus messages.
- a `LastRefreshTime` method that gets the unix timestamp of the update agent actor's internal state's last refresh time to demonstrate that the new `dbus` actor can communicate properly with the `update_agent` actor.

Also includes a CLI for calling these two experimental methods.


src/update_agent/actor.rs
```diff
@@ -33,6 +33,21 @@ impl Actor for UpdateAgent {
     }
 }
 
+pub struct LastRefresh {}
+
+impl Message for LastRefresh {
+    type Result = Fallible<i64>;
```
cgwalters, line 39: Does this need to be `Fallible`?

src/dbus/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+impl Actor for DBusService {
+    type Context = SyncContext<Self>;
+
+    fn started(&mut self, _ctx: &mut Self::Context) {
+        trace!("D-Bus service actor started");
+
+        let connection = zbus::Connection::new_system().unwrap();
```
cgwalters, line 68: Isn't there a way to more properly report errors from actor startup, like a way to return a `Result`?  I am not finding it offhand; looks like `MetricsService` is calling `expect()` (which is an unwrap-wrapper).

At least one thing I'd do in this situation is factor out a helper function that returns a `Result`, and then centralize something like a `log::error!(e); return;` call from that.

src/dbus/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+        object_server
+            .at(
+                &ObjectPath::try_from("/org/coreos/zincati1").unwrap(),
+                experimental_interface,
+            )
+            .unwrap();
+
+        loop {
+            if let Err(err) = object_server.try_handle_next() {
+                eprintln!("{}", err);
```
cgwalters, line 90: Looks like other bits of the code are using `log::warn!`.

src/dbus/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn last_refresh_time(&self) -> Result<i64, zbus::fdo::Error> {
+        let msg = LastRefresh {};
+        let refresh_time_fut = self
+            .agent_addr
+            .send(msg)
+            .unwrap_or_else(|e| Err(e.into()))
+            .map_err(|e| {
+                let err_msg = format!("failed to get last refresh time from agent actor: {}", e);
+                log::error!("{}", err_msg);
+                zbus::fdo::Error::Failed(err_msg)
```
cgwalters, line 39: I think it's generally better to "log xor propagate" i.e. don't both log and propagate an error, because that usually leads to double-logging at least (and if the *callers* also log and propagate you can get many copies).

Although this case is tricky because we're also sending an error to the client, but it's useful to log in the journal too.  So...this is probably OK; it's what rpm-ostree does too.

cgwalters (commented):
I'm not an actix/zincati expert but LGTM overall.

One thing to consider too is double-checking that the caller is uid 0 in the methods using
https://docs.rs/zbus/1.8.0/zbus/fdo/struct.DBusProxy.html#method.get_connection_unix_user
- this is what bootupd does with its own socket.

src/dbus/mod.rs
kelvinfan001: Thanks for pointing that out! But just for my information, in this case, do we still count it as "propagating"? My intuition is that here we are sending a message to a client that an error occurred; but on the server side, the error was never really propagated anywhere, and wouldn't be double logged elsewhere right?

src/update_agent/actor.rs
kelvinfan001: Nope, it doesn't 😅

kelvinfan001:
> One thing to consider too is double-checking that the caller is uid 0 in the methods using
https://docs.rs/zbus/1.8.0/zbus/fdo/struct.DBusProxy.html#method.get_connection_unix_user
This what bootupd does with its own socket, although here we need to ask the central bus daemon for the creds.

@cgwalters how important is it that we check for the caller UID in the method calls, i.e. can the busconfig policy be easily bypassed?
IIUC, to get the caller UID, we first need to get the bus name of the caller from the message sent by the caller. However, `zbus` has this nice `ObjectServer` concept that has methods like [`try_handle_next()`](https://docs.rs/zbus/1.8.0/zbus/struct.ObjectServer.html#method.try_handle_next) that abstracts away the actual D-Bus message and all we need to do is use macros like [`dbus_interface`](https://docs.rs/zbus/1.8.0/zbus/attr.dbus_interface.html) to implement a D-Bus interface.
But I think we can also not use the macros/`ObjectServer` and do things in a more low level approach.

cgwalters:
> @cgwalters how important is it that we check for the caller UID in the method calls, i.e. can the busconfig policy be easily bypassed?

It's not necessary.  I chose to do it in bootupd because it was easy and cheap - and security layers like this are often helpful.

src/dbus/mod.rs
cgwalters: It makes sense here - the error is being returned to the client, but we don't control their logging, and in some cases the admin would like to be able to reliably see errors from the journal.

So fine as is!

cgwalters:
Looks like zbus doesn't build against the MSRV:
```
 error[E0432]: unresolved import `proc_macro`
 --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/zvariant_derive-2.6.0/src/lib.rs:8:5
  |
8 | use proc_macro::TokenStream;
  |     ^^^^^^^^^^ help: a similar path exists: `syn::proc_macro`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: could not compile `zvariant_derive`.
```

1.40 is pretty old now.   Looks like current `registry.redhat.io/rhel8/rust-toolset` is shipping `rust-1.47.0-1.module+el8.3.1+8546+07bf6eb2.x86_64`.  I'd say bumping to at least 1.47 is totally safe.

(Clearly we need some sort of automation and consistency around our MSRV across repos)

dist/dbus-1/system.d/org.coreos.zincati1.conf
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?> <!-- -*- XML -*- -->
+
+<!DOCTYPE busconfig PUBLIC
+          "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
+          "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
+<busconfig>
+
+  <!-- Only user zincati can own the service -->
+  <policy user="zincati">
+    <allow own="org.coreos.zincati1"/>
```
cgwalters, line 10: This is cosmetic of course, but IMO the chance of us ever introducting a `zincati2` bus name is zero.  Personally I'd just name this `zincati` - we don't expect to make this a widely used stable API anyways.

From a quick inspection here in `busctl`: systemd, rpm-ostree and firewalld use the `1` suffix, other services like NetworkManager, fwupd, don't.  But tellingly there are zero `2` or more services :smile:

cgwalters (approved):
One cosmetic bit, otherwise this looks good to me!

src/update_agent/actor.rs
```diff
@@ -76,10 +138,13 @@ impl Handler<RefreshTick> for UpdateAgent {
...
                     pause.as_secs()
                 );
-                Self::tick_later(ctx, pause);
+                let handle = Self::tick_later(ctx, pause);
+                actor.tick_later_handle = Some(handle);
             } else {
                 let update_timestamp = chrono::Utc::now();
                 actor.state_changed = update_timestamp;
+                // Cancel scheduled ticks, if any.
+                actor.cancel_scheduled_ticks(&msg, ctx);
```
kelvinfan001, line 147: I think here it is possible to have a race where a self tick msg and a command msg is sent at the same time, and the command msg arrvies first.
If this happens, we can't really cancel the scheduled tick like this because the message had alrady been sent. Though I don't think it is too big of a deal to have an extra tick; in practice, this results in two continuous ticks without any delay in between when Zincati is in a steady state (`NoNewUpdates` and `UpdateStaged`) where it periodically sends itself refresh ticks.

lucab:
@kelvinfan001 let's maybe stick to the experimental interface here only, and leave the update part out? I fear this PR/discussion may become quite large otherwise.

kelvinfan001:
@lucab I've dropped the `Updates` interface bits. This PR currently contains CLI commands to invoke the two experimental D-Bus methods, but I think there needs to be further discussion on whether we should even have a CLI for Zincati to invoke upcoming useful methods e.g. `CheckUpdate`.

src/dbus/experimental.rs
```diff
@@ -0,0 +1,36 @@
...
+            String::from("moo.")
+        }
+    }
+
+    /// Get update_agent actor's last refresh time.
+    fn last_refresh_time(&self) -> fdo::Result<i64> {
+        let msg = LastRefresh {};
+        let refresh_time_fut = self.agent_addr.send(msg).map_err(|e| {
+            let err_msg = format!("failed to get last refresh time from agent actor: {}", e);
+            log::error!("{}", err_msg);
```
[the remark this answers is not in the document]
kelvinfan001: From [the docs](https://docs.rs/zbus/1.8.0/zbus/struct.ObjectServer.html#method.try_handle_next), it says `try_handle_next()` "returns an error if the message is malformed or an error occurred". So I think `try_handle_next()` returns errors that arise from receiving/interpreting D-Bus messages, and not errors that come out of the functions associated with the D-Bus methods. 
Also "If the message was handled by an an interface, it returns Ok(None)." I think in this case, by the time we get to `self.agent_addr.send()`, the message is already handled by our `Experimental` interface's `LastRefreshTime` method. The error that we're logging here is an error from within the `last_refresh_time()` function. I don't think `try_handle_next()` would know about it/consider it an error even if we return it (i.e. via mapping the error into `fdo::Error::Failed(err_msg)` and returning it). I just tested this out using the `Updates` interface and trying to `CheckUpdate` when the update agent isn't in a state that allows it. The [`check_update()` function](https://github.com/kelvinfan001/zincati/blob/b9d568285b90297ca34441637e9e04aee8f141d3/src/dbus/updates.rs#L50) returned an `fdo::Error` and `try_handle_next()` never logged it to the journal.
Let me know if I'm making any sense :)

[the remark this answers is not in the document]
lucab: I am quite puzzled then, who handles the `fdo::Result` which is returned here?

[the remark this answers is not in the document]
kelvinfan001: I understood that the D-Bus client calling this method (e.g. `busctl`) handles it. `fdo::Error::Failed` implements [`org.freedesktop.DBus.Error.Failed`](https://gitlab.freedesktop.org/dbus/dbus/-/blob/master/dbus/dbus-protocol.h#L359).

src/dbus/experimental.rs
```diff
@@ -0,0 +1,35 @@
...
+    /// Get update_agent actor's last refresh time.
+    fn last_refresh_time(&self) -> fdo::Result<i64> {
+        let msg = LastRefresh {};
+        let refresh_time_fut = self.agent_addr.send(msg).map_err(|e| {
+            let err_msg = format!("failed to get last refresh time from agent actor: {}", e);
+            log::error!("LastRefreshTime D-Bus method call: {}", err_msg);
+            fdo::Error::Failed(err_msg)
+        });
+
+        System::new("LastRefreshTime System").block_on(refresh_time_fut)
```
kelvinfan001, line 33: @lucab I'm not sure if this is a proper usage of `actix::System`. It seems quite heavy-handed that we need to create a new system just to block the current thread to resolve a future, and then throw the system away.
Am I using this correctly?

src/dbus/experimental.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// Get update_agent actor's last refresh time.
+    fn last_refresh_time(&self) -> fdo::Result<i64> {
+        let msg = LastRefresh {};
+        let refresh_time_fut = self.agent_addr.send(msg).map_err(|e| {
+            let err_msg = format!("failed to get last refresh time from agent actor: {}", e);
+            log::error!("LastRefreshTime D-Bus method call: {}", err_msg);
+            fdo::Error::Failed(err_msg)
+        });
+
+        Runtime::new()
```
kelvinfan001, line 34: Following up to this comment https://github.com/coreos/zincati/pull/514#discussion_r605931720, I've opted to use tokio's `Runtime::new()` directly.

lucab: Ah, I think you are right. I was hoping to reuse the existing `System` but it looks like it can't be done that way.
There is possibly a smarter way to share the runtime here (without creating a new one from scratch each time), but this looks good enough for now.

src/update_agent/actor.rs
```diff
@@ -33,6 +33,21 @@ impl Actor for UpdateAgent {
...
+pub struct LastRefresh {}
+
+impl Message for LastRefresh {
+    type Result = i64;
+}
+
+impl Handler<LastRefresh> for UpdateAgent {
+    type Result = i64;
+
+    fn handle(&mut self, _msg: LastRefresh, _ctx: &mut Self::Context) -> Self::Result {
```
lucab, line 45: Self-note: it would be interesting to check whether this is handled in a concurrent way (i.e. on the spot) even when there is a long-running `RefreshTick` already being processed (e.g. an rpm-ostree deploy taking quite some time to download the new content).

lucab (approved):

src/dbus/experimental.rs
kelvinfan001: Coming back to this after some additional reading on async and actix: 
Is the reason why we can't use the existing `System` because the DBus actor is hosted on a separate `SyncArbiter`, which is not async? Notably, we don't have access to an `AsyncContext`. It looks to me like there is no executor for this actor, so we have to create our own.
For some context, the DBus actor is started via `SyncArbiter::start(threads, move || DBusService::new(agent_addr.clone()))`

src/update_agent/actor.rs
kelvinfan001: It looks like this is _not_ handled in a concurrent way. I think it is because when we handle `RefreshTick`s, we use `actix::AsyncContext::wait()`, and so it stops processing any incoming events until the future resolves, essentially making it synchronous.
I've done some experimentation here: https://gist.github.com/kelvinfan001/14f933776c51ca6e455b5797b72531c8

PR #515, opened 2021-03-29
github/workflows/rust.yml: bump MSRV toolchain

kelvinfan001:
Since `registry.redhat.io/rhel8/rust-toolset` is shipping
`rust-1.47.0-1.module+el8.3.1+8546+07bf6eb2.x86_64` already, it
should be safe to bump MSRV toolchain to `1.47.0`.
Otherwise, the `zbus` crate (which we would like to use soon) is
unable to build.

xref: https://github.com/coreos/zincati/pull/514#issuecomment-809526987


travier (approved):
LGTM

Issue #519, opened 2021-04-01
cargo: move from 'failure' (now deprecated) to 'anyhow'


lucab:
When this project started, https://github.com/rust-lang-nursery/failure was the most ergonomic approach to error-handling.

In the meanwhile, lot of things have been modernized and improved in the stdlib. The library is now marked as deprecated. https://github.com/dtolnay/anyhow is a good replacement for applications like Zincati.

lucab:
Closed via https://github.com/coreos/zincati/pull/533

PR #521, opened 2021-04-07
[WIP] Add new D-Bus Updates interface

kelvinfan001:


src/update_agent/actor.rs
```diff
@@ -91,31 +139,50 @@ impl Handler<RefreshTick> for UpdateAgent {
...
+            self.tick_later_handle = None;
+        }
+    }
+
     /// Schedule an immediate refresh of the state machine.
-    pub fn tick_now(ctx: &mut Context<Self>) {
-        ctx.notify(RefreshTick {})
+    pub fn tick_now(&mut self, ctx: &mut Context<Self>) {
+        // Cancel scheduled ticks, if any.
+        self.cancel_scheduled_ticks(ctx);
```
kelvinfan001, line 171: I think here it is possible to have a race where a self tick msg and a command msg is sent at the same time, and the command msg arrives first.
If this happens, we can't really cancel the scheduled tick like this because the message had alrady been sent. Though I don't think it is too big of a deal to have an extra tick; in practice, this results in two continuous ticks without any delay in between when Zincati is in a steady state (NoNewUpdates and UpdateStaged) where it periodically sends itself refresh ticks.

src/update_agent/actor.rs
```diff
@@ -91,31 +139,50 @@ impl Handler<RefreshTick> for UpdateAgent {
...
 
-        Box::pin(actix::fut::ok(()))
+        Box::pin(actix::fut::ok(self.state.clone()))
     }
 }
 
 impl UpdateAgent {
+    /// Cancel the scheduled refresh tick whose handle is stored in the update agent's
+    /// `tick_later_handle` field, if any.
+    fn cancel_scheduled_ticks(&mut self, ctx: &mut Context<Self>) {
```
kelvinfan001, line 161: When transitioning "naturally" (no messages received from the D-Bus actor) from a "steady state" (e.g. `NoNewUpdates` or `UpdateAvailable`) to a new state, `tick_later_handle` is a handle to
a scheduled msg that has already been sent, so there is no need to cancel it. But it doesn't look like there is any harm in calling `ctx.cancel_future()` on a handle that is already sent, so for simplicity, always attempt to cancel.
But let me know if this doesn't make sense.

kelvinfan001 (commented):
Some notes on parts I'm iffy about.

Cargo.toml
```diff
@@ -17,6 +17,7 @@ env_logger = "0.8"
 envsubst = "0.2"
 fail = "0.4"
 failure = "0.1"
+failure_derive = "0.1.8"
```
lucab, line 20: Let's do https://github.com/coreos/zincati/issues/519 instead of going this way.

src/dbus/updates.rs
```diff
@@ -0,0 +1,82 @@
...
+            .map_err(|e| {
+                let err_msg = format!("failed to create runtime to execute future: {}", e);
+                log::error!("{}", err_msg);
+                Failed(err_msg)
+            })
+            .and_then(|mut runtime| runtime.block_on(refresh_time_fut))
+    }
+}
+
+#[dbus_interface(name = "org.coreos.zincati.Updates")]
```
lucab, line 38: I'd rather keep this under the existing `Experimental` interface for now, and iterate on the design there.

src/update_agent/actor.rs
```diff
@@ -48,16 +48,64 @@ impl Handler<LastRefresh> for UpdateAgent {
     }
 }
 
-pub(crate) struct RefreshTick {}
+/// Error thrown when the command that attempted to initiate a refresh tick
+/// is not permitted to do so in the current state of the update agent.
+#[derive(Debug, Fail)]
+struct TickPermissionError {}
```
lucab, line 54: `Permission` usually has a different semantic nuance, related to a lack of privileges. In this case this is mostly related to an unsatisfiable request or an unfulfilled precondition. I don't a proper suggestion for this right now, but it would be better to rename this.

cgwalters: `InvalidStateError`?

Issue #301, opened 2020-06-04
updates/strategy: local timezone aware reboot windows


lucab:
Zincati 0.0.11 gained a new `periodic` strategy for scheduling reboot windows on a weekly calendar.

Its design decisions are noted [here](https://github.com/coreos/zincati/blob/c0d2f264d3960e3c2b8a74d6517c675d0f1782b9/docs/development/update-strategy-periodic.md) and the configuration options [here](https://github.com/coreos/zincati/blob/c0d2f264d3960e3c2b8a74d6517c675d0f1782b9/docs/usage/updates-strategy.md#periodic-strategy).

There are three main points that pinned the `periodic` strategy to use UTC times and dates:
 * avoiding ambiguities for human observers
 * avoiding duplicating/skipping a reboot window on Daylight Saving Time (DST) bumps
 * guaranteeing the length of the reboot windows even when DST/TZ change

However this comes with a few usability drawbacks:
 * people may not be used to handle/reason about UTC times (especially in places far from GMT)
 * some users may be more interested about guaranteeing local-time invariants, accepting occasional duplication/skips on time changes

One idea that was floated is to add a new strategy which uses systemd time labels as documented in https://www.freedesktop.org/software/systemd/man/systemd.time.html.
That format however is meant for edge-triggers (i.e. precise points in time) instead of level-triggers (i.e. time windows). It also results in possibly weird windows definitions like `yesterday` and `weekly`.

Another idea is to add low-level strategy (e.g. the filesystem-based one at https://github.com/coreos/zincati/issues/245) so that high-level components could plug into it. In this case a systemd timer+service unit would handle the filesystem entry based on local time.

jlebon:
Heh, was going to file a similar issue. I think for the "single pet FCOS server" case, it's really annoying to have to define things in UTC time. And it also means that the time window shifts wrt your local time when going from e.g. EST to EDT or vice versa.

lucab:
On this topic, I still think it's worthy exploring a path where a timer unit does most of the time-related work (i.e. expose a flexible configuration), and Zincati only checks its status in order to get a go/no-go signal.

lucab:
Expanding a bit more on the timer units, from an high level perspective the optimal UX would be something like:
 * via Ignition, the user defines one or more [systemd timer units](https://www.freedesktop.org/software/systemd/man/systemd.timer.html#) covering all the allowed reboot timespans (e.g. `wed-night.timer` and `sunday-morning.timer`).
[...]

PR #524, opened 2021-04-09
strategy/periodic: support configuring a time zone for reboot windows

kelvinfan001:
Support configuring reboot windows defined by a user-specified time zone, instead of only UTC.

Include a single `time_zone` field (that defaults to UTC if not
specified) in the `periodic` strategy's configuration. Whenever we
need to check whether we're currently in an allowed reboot window,
we convert the current naive time to the naive time under the
specified time zone, and check whether we are currently in a reboot
window.

Under this implementation, we keep invariant clock time for reboots,
but do NOT keep invariant reboot window length; thus, in some cases,
reboot windows can be lengthened, shortened, or skipped entirely.

### DIFF
--- a/docs/usage/updates-strategy.md
+++ b/docs/usage/updates-strategy.md
@@ -71,7 +71,7 @@ The `periodic` strategy allows Zincati to only reboot for updates during certain
 Outside of those maintenance windows, reboots are not automatically performed and auto-updates are staged and held until the next available window.
 
 Reboot windows recur on a weekly basis, and can be defined in any arbitrary order and length. Their individual length must be greater than zero.
-To avoid timezone-related skews in a fleet of machines, all maintenance windows are defined in UTC dates and times.
+By default, all maintenance windows are defined in UTC dates and times. This is meant to avoid timezone-related skews in a fleet of machines, as well as possible side-effects of Daylight Savings Time (DST) policies.
 
 Periodic reboot windows can be configured and enabled in the following way:
 
@@ -101,3 +101,71 @@ Reboot windows can be separately configured in multiple snippets, as long as eac
  * `length_minutes`: non-zero window duration, in minutes
 
 For convenience, multiple entries can be defined with overlapping times, and each window definition is allowed to cross day and week boundaries (wrapping to the next day).
+
+## Time zone configuration
+
+To configure a non-UTC time zone for all the reboot windows, specify the `time_zone` field in a `updates.periodic` entry. The specified time zone must be either `"localtime"` or a time zone name from the [IANA Time Zone Database][IANA_tz_db] (you can find an unofficial list of time zone names [here][wikipedia_tz_names]).
+
+If using `"localtime"`, the system's [local time zone configuration file][localtime], `/etc/localtime`, is used. As such, `/etc/localtime` must either be a symlink to a valid `tzfile` entry in your system's local time zone database (under `/usr/share/zoneinfo/`), or not exist, in which case `UTC` is used.
+
+Note that you can only specify a single time zone for _all_ reboot windows.
+
+A time zone can be specified in the following way:
+
+```toml
+[updates]
+strategy = "periodic"
+
+[updates.periodic]
+time_zone = "America/Panama"
+
+[[updates.periodic.window]]
+days = [ "Sat", "Sun" ]
+start_time = "23:30"
+length_minutes = 60
+
+[[updates.periodic.window]]
+days = [ "Mon" ]
+start_time = "00:00"
+length_minutes = 60
+```
+
+Since Panama does not have Daylight Savings Time and follows Eastern Standard Time (which has a fixed offset of UTC -5) all year, the above configuration would result in two maintenance windows during which Zincati is allowed to reboot the machine for updates:
+ * 60 minutes starting at 23:30 EST on Saturday night, and ending at 00:30 EST on Sunday morning
+ * 90 minutes starting at 23:30 EST on Sunday night, and ending at 01:00 EST on Monday morning
+
+### Time zone caveats
+
+:warning: **Reboot window lengths may vary.**
+
+Because reboot window clock times are always obeyed, reboot windows may be lengthened or shortened due to shifts in clock time. For example, with the `US/Eastern` time zone which shifts between Eastern Standard Time and Eastern Daylight Time, on "fall back" day, a specified reboot window may be lengthened by up to one hour; on "spring forward" day, a specified reboot window may be shortened by up to one hour, or skipped entirely.
+
+Example of varying length reboot windows using the `US/Eastern` time zone:
+
+```toml
+[updates]
+strategy = "periodic"
+
+[updates.periodic]
+time_zone = "US/Eastern"
+
+[[updates.periodic.window]]
+days = [ "Sun" ]
+start_time = "01:30"
+length_minutes = 60
+```
+
+The above configuration will result in reboots being allowed at 1:30 AM to 2:30 AM on _every_ Sunday. This includes days when a Daylight Savings Shift occurs.
+
+On the `US/Eastern` time zone's "fall back" day, where clocks are shifted back by one hour on a Sunday in Fall just before 3:00 AM, the thirty minutes between 2:00 AM and 2:30 AM will occur twice. As such, the reboot window will be lengthened by thirty minutes each year on "fall back" day.
+
+On "spring forward" day, where clocks are shifted forward by one hour on a Sunday in Spring just before 2:00 AM, the thirty minutes between 2:00 AM and 2:30 AM will not occur. As such, the reboot window will be shortened by thirty minutes each year on "spring forward" day. Effectively, the reboot window on "spring forward" day will only be between 1:30 AM and 2:00 AM.
+
+:warning: **Incorrect reboot times due to stale time zone database.** 
+
+Time zone data is read from the system's time zone database at `/usr/share/zoneinfo`. This directory and its contents are part of the `tzdata` RPM package; in the latest release of Fedora CoreOS, `tzdata` should be kept fairly up-to-date with the latest official release from the IANA.
+However, if your system does not have the latest IANA time zone database, or there is a sudden policy change in the jurisdiction associated with your configured time zone, then reboots may happen at unexpected and incorrect times.
+
+[IANA_tz_db]: https://www.iana.org/time-zones
+[wikipedia_tz_names]: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+[localtime]: https://www.freedesktop.org/software/systemd/man/localtime.html
--- a/src/config/fragments.rs
+++ b/src/config/fragments.rs
@@ -77,6 +77,11 @@ pub(crate) struct UpdateFleetLock {
 pub(crate) struct UpdatePeriodic {
     /// A weekly window.
     pub(crate) window: Option<Vec<UpdatePeriodicWindow>>,
+    /// A time zone in the IANA Time Zone Database (https://www.iana.org/time-zones)
+    /// or "localtime". If unset, UTC is used.
+    ///
+    /// Examples: `America/Toronto`, `Europe/Rome`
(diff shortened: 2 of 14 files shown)


kelvinfan001:
The current implementation uses Option 3 from https://github.com/coreos/zincati/issues/301#issuecomment-816728794.
This is still a work in progress as we need to properly address the third bullet point in https://github.com/coreos/zincati/issues/301#issuecomment-816685924.

src/strategy/mod.rs
```diff
@@ -92,7 +92,13 @@ impl UpdateStrategy {
                 "{}, total schedule length {} minutes (next window {})",
                 self.configuration_label(),
                 p.schedule_length_minutes(),
-                p.human_remaining()
+                // If not using UTC, we cannot accurately calculate remaining time;
+                // display next window week day and time, instead.
+                if p.use_system_time {
+                    p.human_next_window()
```
kelvinfan001, line 98: `human_next_window()` attempts to return the week day and time of the next nearest reboot window. However, the current implementation _may_ lead to an incorrect next nearest reboot window if the function is called in the week of a "DST shift" (e.g. in Eastern Time, the week of "spring forward" or "fall back" day).

src/config/fragments.rs
```diff
@@ -77,6 +77,9 @@ pub(crate) struct UpdateFleetLock {
 pub(crate) struct UpdatePeriodic {
     /// A weekly window.
     pub(crate) window: Option<Vec<UpdatePeriodicWindow>>,
+    /// Whether to use system time's time zone (i.e. `/etc/localtime`)
+    /// as the time zone for reboot windows.
+    pub(crate) use_system_time: Option<bool>,
```
lucab, line 82: Bikeshedding: I think this may work better has a `use_utc: Option<bool>` which is on by default, and users have to explicitly opt-out.

src/strategy/periodic.rs
```diff
@@ -2,19 +2,35 @@
...
+use tzfile::Tz;
 
 /// Strategy for periodic (weekly) updates.
-#[derive(Clone, Debug, Default, Serialize)]
+#[derive(Clone, Debug, Serialize)]
 pub(crate) struct StrategyPeriodic {
     /// Whitelisted time windows during which updates are allowed.
     schedule: WeeklyCalendar,
+    /// Time zone in which time windows are defined in.
+    #[serde(skip_serializing)]
```
kelvinfan001, line 23: `tzfile::Tz` currently does not implement `Serialize`.

src/strategy/periodic.rs
```diff
@@ -41,15 +58,57 @@ impl StrategyPeriodic {
...
+    fn get_time_zone(cfg: &inputs::PeriodicInput) -> Fallible<Tz> {
+        let tz;
+        if &cfg.time_zone == "localtime" {
+            // Use `read_link` instead of `exists()` because we only want to check for
+            // the existence of the `/etc/localtime` symlink, not whether it points to
+            // a valid location. (`read_link` returns an error if symlink doesn't exist)
+            if let Err(_) = read_link(Path::new("/etc/localtime")) {
+                tz = Tz::named("UTC");
+            } else {
+                tz = Tz::local();
```
kelvinfan001, line 78: After switching to `anyhow`, we should probably add some context to the error that `Tz::local()` might return if `/etc/localtime` is a symlink to a bad/non-existent `tzfile`.
Currently, if `/etc/localtime` points to bogus, all we see is the following in the journal:
```
[ERROR] critical error: failed to assemble configuration settings
[ERROR]  -> failed to validate update-strategy configuration
[ERROR]  -> No such file or directory (os error 2)
```
It may be difficult to identify what went wrong in the update-strategy config.

src/strategy/periodic.rs
```diff
@@ -41,15 +58,57 @@ impl StrategyPeriodic {
...
         Ok(strategy)
     }
 
+    /// Get the time zone from `periodic` strategy config.
+    fn get_time_zone(cfg: &inputs::PeriodicInput) -> Fallible<Tz> {
+        let tz;
+        if &cfg.time_zone == "localtime" {
+            // Use `read_link` instead of `exists()` because we only want to check for
+            // the existence of the `/etc/localtime` symlink, not whether it points to
+            // a valid location. (`read_link` returns an error if symlink doesn't exist)
```
kelvinfan001, line 74: To add some context as to why I used `read_link()` here, we only want to default to `UTC` if `/etc/localtime` doesn't exist, and not if the location is invalid, in which case `Tz::local()` will return an Error. This is to prevent surprising users with a UTC time zone if they _think_ they're using `localtime`, but accidentally symlinked an invalid time zone when trying to configure their system's local time.

src/strategy/periodic.rs
```diff
@@ -41,15 +58,57 @@ impl StrategyPeriodic {
...
+        let naive_utc_dt = Utc::now().naive_utc();
+        let dt = (&self.time_zone).from_utc_datetime(&naive_utc_dt);
+        let next_window_minute_in_week = self.schedule.next_window_minute_in_week(&dt);
+
+        match next_window_minute_in_week {
+            Some(minute_in_week) => {
+                let (weekday, hour, minute) = utils::weekly_minute_as_weekday_time(minute_in_week);
+                // `tmp_dt` is created just to use the `%Z` specifier of its `format()` method.
+                let tmp_dt = (&self.time_zone).from_utc_datetime(&Utc::now().naive_utc());
+                let tz_abbrev = tmp_dt.format("%Z").to_string();
```
kelvinfan001, line 102: This is a hacky way to get an abbreviation or otherwise some sort of name of the configured time zone. Ideally, `tzfile::Tz` should implement `Display`.

kelvinfan001 (commented):
I added some comments in places I feel like needs some extra attention :)

docs/usage/updates-strategy.md
```diff
@@ -101,3 +101,35 @@ Reboot windows can be separately configured in multiple snippets, as long as eac
  * `length_minutes`: non-zero window duration, in minutes
 
 For convenience, multiple entries can be defined with overlapping times, and each window definition is allowed to cross day and week boundaries (wrapping to the next day).
+
```
lucab, line 104: Let's add a `## Timezone configuration` header so that's it's easier to link to.

docs/usage/updates-strategy.md
```diff
@@ -101,3 +101,35 @@ Reboot windows can be separately configured in multiple snippets, as long as eac
...
+To configure a non-UTC time zone for reboot windows, specify the `time_zone` field in a `updates.periodic` entry. The specified time zone must be either `"localtime"` or a time zone name from the [IANA Time Zone Database][IANA_tz_db] (you can find an unofficial list of time zone names [here][wikipedia_tz_names]). 
+If using `"localtime"`, `/etc/localtime` must either be a symlink to a valid `tzfile` entry in your system's local time zone database (under `/usr/share/zoneinfo/`), or not exist, in which case `UTC` is used.
+
+Note that you can only specify a single time zone for _all_ reboot windows.
+
+:warning: **When using non UTC time zones:** reboot window clock times are always obeyed, therefore **reboot window lengths may vary**!
+Reboot windows may be lengthened or shortened due to shifts in clock time. For example, with the `US/Eastern` time zone, on "fall back" day, a specified reboot window may be lengthened by up to one hour; on "spring forward" day, a specified reboot window may be shortened by up to one hour, or skipped entirely.
+
+Example of periodic reboot windows using the `US/Eastern` time zone.
```
lucab, line 113: It may better to start with a simpler non-DST example first, and move this longer `US/Eastern` example and explanation to its own `## Timezone caveats` section.
`America/Panama` should be a good enough pick as it never introduced DST (so far) and is fairly recognizable by the average user.

src/config/fragments.rs
```diff
@@ -77,6 +77,11 @@ pub(crate) struct UpdateFleetLock {
 pub(crate) struct UpdatePeriodic {
     /// A weekly window.
     pub(crate) window: Option<Vec<UpdatePeriodicWindow>>,
+    /// A time zone in the IANA Time Zone Database (https://www.iana.org/time-zones)
+    /// or "localtime".
```
lucab, line 81: `If unset, UTC datetimes are used.`

src/strategy/mod.rs
```diff
@@ -89,10 +89,16 @@ impl UpdateStrategy {
             UpdateStrategy::FleetLock(_) => self.configuration_label().to_string(),
             UpdateStrategy::Immediate(_) => self.configuration_label().to_string(),
             UpdateStrategy::Periodic(p) => format!(
-                "{}, total schedule length {} minutes (next window {})",
```
lucab, line 92: This `format!` logic became large enough that it can be moved to its own `p.calendar_summary()` or similar.

src/strategy/mod.rs
```diff
@@ -89,10 +89,16 @@ impl UpdateStrategy {
...
+                "{}, total schedule length {} minutes; next window {}",
                 self.configuration_label(),
                 p.schedule_length_minutes(),
-                p.human_remaining()
+                // If not using UTC, we cannot accurately calculate remaining time;
+                // display next window week day and time, instead.
+                if p.time_zone != tzfile::Tz::named("UTC").unwrap() {
```
lucab, line 97: I think it makes sense to place this into a `p.is_calendar_utc() -> bool`, caching the boolean upfront instead of re-computing it each time.

src/strategy/mod.rs
lucab: IMHO This is seem good enough already as you did. Do you think we should maybe consider doing `human_next_window()` unconditionally instead?

src/strategy/periodic.rs
lucab: From an out-of-band chat: we should consider adding `serde` support directly to that crate, possibly behind a feature gate.

src/strategy/periodic.rs
lucab: This is possibly something to bring to upstream, however it may be hard to sell as timezones don't have a canonical/preferred name (see for example `UTC` vs `Etc/UTC`).

src/strategy/periodic.rs
```diff
@@ -41,15 +58,57 @@ impl StrategyPeriodic {
...
+                format!(
+                    "at {}:{} on {} ({}) if no shift in clock.",
```
lucab, line 104: Up for your consideration, but you may turn the last part into a `(subject to timezone caveats)` if not-UTC.

src/strategy/mod.rs
kelvinfan001: I think `human_next_window()` is easier to understand if the reader is in the same time zone as the configured time zone (likely the case when using a non-UTC time zone). But if the user is in a different time zone than the configured time zone (likely the case when using UTC), `human_next_window()` might require some careful mental gymnastics, especially if the reboot window is near.
TL;DR, I think `human_remaining()` is still useful if we're using UTC.

src/strategy/periodic.rs
kelvinfan001: Good point. Perhaps don't implement `Display`, but maybe have a `name()` method that returns the last bit of the path of the `tzfile(5)` that the `tzfile::Tz` struct reads its data from, e.g. two different `Tz` instances may return `America/Toronto` or `Canada/Eastern` even though the data should be the same.

src/strategy/periodic.rs
kelvinfan001: That's a good idea, since the next reboot window information isn't _always_ wrong even in the week of a shift in clock.

src/strategy/mod.rs
kelvinfan001: We now have a `p.tz_name()` that gets the "cached" name of the time zone. Using this over a boolean because we need to use this name in `human_next_reboot_window()` too.

lucab (approved):

jlebon:
Thanks, @kelvinfan001! Looking forward to using this on my FCOS server. :)

travier:
Same here! Thanks!

PR #529, opened 2021-04-15
ci: Use fcos buildroot

cgwalters:
I think we should try to have a common workflow across CoreOS
projects, so use the new upstream buildroot.

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -8,18 +8,12 @@ on:
     branches: [master]
 
 jobs:
-  container-dev:
-    name: "Dev container"
+  build-fcos:
+    name: "Build in FCOS buildroot"
     runs-on: ubuntu-latest
+    container: quay.io/coreos-assembler/fcos-buildroot:testing-devel
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
-      - name: Assemble dev-container
-        run: docker build -f contrib/Dockerfile.dev -t coreos/zincati:dev .
-      - name: Run dev-container
-        run: docker run --rm -v "$(pwd):/source" -v "/tmp/assembler:/assembler" coreos/zincati:dev
-      - name: Archive artifacts (overrides/rootfs)
-        uses: actions/upload-artifact@v2
-        with:
-          name: zincati-overrides-rootfs
-          path: /tmp/assembler/overrides/rootfs/
+      - name: Build and test
+        run: make


cgwalters:
(This leaves open the question of whether to keep the `contrib/Dockerfile.dev` at all)

lucab:
I think it's fine to drop `contrib/Dockerfile.dev`, we just need to adjust https://github.com/coreos/zincati/blob/master/docs/development/quickstart.md#build-and-test for that. No need to do that here though.

kelvinfan001 (approved):
This makes sense to me :)

.github/workflows/containers.yml
```diff
@@ -8,18 +8,15 @@ on:
...
         uses: actions/checkout@v2
-      - name: Assemble dev-container
-        run: docker build -f contrib/Dockerfile.dev -t coreos/zincati:dev .
-      - name: Run dev-container
-        run: docker run --rm -v "$(pwd):/source" -v "/tmp/assembler:/assembler" coreos/zincati:dev
-      - name: Archive artifacts (overrides/rootfs)
-        uses: actions/upload-artifact@v2
-        with:
-          name: zincati-overrides-rootfs
-          path: /tmp/assembler/overrides/rootfs/
```
kelvinfan001, line 25: I actually have a question about this part in the original code though, what is the purpose of this?

.github/workflows/containers.yml
```diff
@@ -8,18 +8,15 @@ on:
...
+      - name: Build
+        run: make
+      - name: Build
```
kelvinfan001, line 20: Do we want to use a different name for this step?

kelvinfan001: Also seems like `make` already runs the cargo tests, so is `make check` still necessary?

.github/workflows/containers.yml
lucab: It does persist built artifacts into an archive. You can see it under "Artifacts" on other PRs, like here: https://github.com/coreos/zincati/pull/524/checks.

I never really used so I'm not too worried to see it go, but longer term it would be nice to have auto-built binaries ready to download (in the form of an RPM, or ostree content, or a tarball, or whatever).

kelvinfan001: ah, I see. Thanks!

kelvinfan001:
> I think it's fine to drop contrib/Dockerfile.dev, we just need to adjust https://github.com/coreos/zincati/blob/master/docs/development/quickstart.md#build-and-test for that.

Submitted https://github.com/coreos/zincati/pull/532 to do that.

.github/workflows/containers.yml
cgwalters: Fixed

kelvinfan001 (approved):
LGTM

PR #532, opened 2021-04-15
doc/quickstart: use FCOS buildroot image

kelvinfan001:
We can just use for development the same image that CI in CoreOS projects use.


kelvinfan001:
Depends on https://github.com/coreos/zincati/pull/529

cgwalters (approved):

PR #533, opened 2021-04-16
Clean up error handling

kelvinfan001:
### Move from `failure` to `anyhow`

The `failure` library is now deprecated, use `anyhow` instead.

### Add and use `fn-error-context` for more reliable error traces:

Same motivation as https://github.com/coreos/bootupd/pull/163.

Effectively what we're doing here is creating a human-readable subset
of the stack trace. This is nicer than having the calling functions
add `with_context()` because it's more verbose (gets duplicative at
each call site), easy to forget, etc.

While we're at it, add some `fn-error-context`-style contexts to
functions whose callers forgot to or didn't add context.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,13 +11,14 @@ edition = "2018"
 
 [dependencies]
 actix = "0.10"
+anyhow = "1.0"
 cfg-if = "1.0"
 chrono = { version = "0.4", features = ["serde"] }
 env_logger = "0.8"
 envsubst = "0.2"
 fail = "0.4"
-failure = "0.1"
 filetime = "0.2"
+fn-error-context = "0.1"
 futures = "0.3"
 glob = "0.3"
 intervaltree = "0.2.6"
@@ -35,8 +36,9 @@ regex = "1.4"
 reqwest = { version = "0.10", features = ["json"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-tempfile = "^3.2"
 structopt = "0.3"
+tempfile = "^3.2"
+thiserror = "1.0"
 tokio = "0.2"
 toml = "0.5"
 url = { version = "2.2", features = ["serde"] }
--- a/src/identity/platform.rs
+++ b/src/identity/platform.rs
@@ -6,7 +6,7 @@
 //!  flags. Logic is taken from Afterburn, please backport any bugfix there too:
 //!  https://github.com/coreos/afterburn/blob/v4.1.0/src/util/cmdline.rs
 
-use failure::{bail, format_err, Fallible, ResultExt};
+use anyhow::{Context, Result};
 use std::io::Read;
 use std::{fs, io};
 
@@ -14,21 +14,21 @@ use std::{fs, io};
 static CMDLINE_PLATFORM_FLAG: &str = "ignition.platform.id";
 
 /// Read platform value from cmdline file.
-pub(crate) fn read_id<T>(cmdline_path: T) -> Fallible<String>
+pub(crate) fn read_id<T>(cmdline_path: T) -> Result<String>
 where
     T: AsRef<str>,
 {
     // open the cmdline file
     let fpath = cmdline_path.as_ref();
-    let file = fs::File::open(fpath)
-        .with_context(|e| format_err!("failed to open cmdline file {}: {}", fpath, e))?;
+    let file =
+        fs::File::open(fpath).with_context(|| format!("failed to open cmdline file {}", fpath))?;
 
     // read content
     let mut bufrd = io::BufReader::new(file);
     let mut contents = String::new();
     bufrd
         .read_to_string(&mut contents)
-        .with_context(|e| format_err!("failed to read cmdline file {}: {}", fpath, e))?;
+        .with_context(|| format!("failed to read cmdline file {}", fpath))?;
 
     // lookup flag by key name
     match find_flag_value(CMDLINE_PLATFORM_FLAG, &contents) {
@@ -36,7 +36,7 @@ where
             log::trace!("found platform id: {}", platform);
             Ok(platform)
         }
-        None => bail!(
+        None => anyhow::bail!(
             "could not find flag '{}' in {}",
             CMDLINE_PLATFORM_FLAG,
             fpath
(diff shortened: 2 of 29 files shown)


Cargo.toml
```diff
@@ -11,13 +11,14 @@ edition = "2018"
 
 [dependencies]
 actix = "0.10"
+anyhow = "1.0.40"
```
lucab, line 14: Here and everywhere else, it should be enough to keep dependencies labels as "x.y" to avoid a lot z-updating activity (unless we know we strictly depend on some bugfix in a given a z-version).
The `Cargo.lock` in the this repo is the one meant to keep track of specific "x.y.z" pinnings.

src/identity/platform.rs
```diff
@@ -6,37 +6,37 @@
...
+    let file =
+        fs::File::open(fpath).with_context(|| anyhow!("failed to open cmdline file {}", fpath))?;
 
     // read content
     let mut bufrd = io::BufReader::new(file);
     let mut contents = String::new();
     bufrd
         .read_to_string(&mut contents)
-        .with_context(|e| format_err!("failed to read cmdline file {}: {}", fpath, e))?;
+        .with_context(|| anyhow!("failed to read cmdline file {}", fpath))?;
```
lucab, line 31: I think this may be a plain `format!()`.

src/strategy/fleet_lock.rs
```diff
@@ -3,7 +3,8 @@
 use crate::config::inputs;
 use crate::fleet_lock::{Client, ClientBuilder};
 use crate::identity::Identity;
-use failure::{format_err, Error, Fallible};
+use anyhow::{anyhow, Error, Result};
+// use failure::{format_err, Error, Fallible};
```
lucab, line 7: Leftover.

src/weekly/utils.rs
```diff
@@ -74,27 +75,28 @@ pub(crate) fn weekday_from_string(input: &str) -> Fallible<Weekday> {
 /// assert_eq!(morning.0, 14);
 /// assert_eq!(morning.0, 5);
 /// ```
-pub(crate) fn time_from_string(input: &str) -> Fallible<(u8, u8)> {
+#[context("failed to parse time string {}", input)]
```
lucab, line 78: I guess this will result in duplicated `input` in the error message, as the inner errors are also printing it. It is likely enough to only print it once, either here or at the bottom of the chain (my vote goes to the latter).

src/strategy/periodic.rs
```diff
@@ -22,7 +23,8 @@ impl StrategyPeriodic {
     pub const LABEL: &'static str = "periodic";
 
     /// Build a new periodic strategy.
-    pub fn new(cfg: inputs::UpdateInput) -> Fallible<Self> {
+    #[context("failed to build new periodic strategy")]
```
lucab, line 26: I'd suggest something like: `parsing periodic updates strategy`.

src/strategy/mod.rs
```diff
@@ -39,14 +40,15 @@ pub(crate) enum UpdateStrategy {
 
 impl UpdateStrategy {
     /// Try to parse config inputs into a valid strategy.
-    pub(crate) fn with_config(cfg: inputs::UpdateInput, identity: &Identity) -> Fallible<Self> {
+    #[context("failed to validate update-strategy configuration")]
```
lucab, line 43: `updates strategy`

lucab (commented):
Thanks, great cleanup! I've left a bunch of minor comments, and a general remark on dependencies.

src/strategy/periodic.rs
cgwalters: One data point here: the bootupd/rpm-ostree wording in error contexts is semi-predicated on the idea that we'll flatten the error message to a single line.  IOW an error would look like:

```
error: Updating: Post-transaction cleanup: Removing /var/lib/somedata: Input/output error
```

The default anyhow formatter though really wants to preserve the full cause chain - but the way it prints looks weird unless each string stands on its own:

```
error: Failed to update
Caused by:
  0: Failed post-transaction cleanup
  1: Couldn't remove /var/lib/somedata
  2: Input/output error
```

In this model the formatting hence IMO looks weird unless *every* context message includes the string "failed" or otherwise makes clear it's an error.

I think this can be kind of mitigated by ensuring that the final error has `error: ` as a prefix.  In this case if it was just "parsing periodic updates strategy" then you'd see:

`error: parsing periodic updates strategy`
which isn't bad at all, but IMO reads better as `error: Failed to parse periodic updates strategy`.

cgwalters (approved):
LGTM overall!  Nice work!

src/identity/platform.rs
kelvinfan001: ah thanks! just noticed there are a couple instances of this. I wonder why it was using `format_err!` before though?

Cargo.toml
kelvinfan001: ohh that makes sense. thanks!

src/strategy/periodic.rs
kelvinfan001: Admittedly, I hadn't thought about each string standing on its own vs flattening the error message to a single line, I just mimicked the style in the `anyhow` docs. But IIUC, Zincati logs the errors to the journal by chaining them like this https://github.com/coreos/zincati/blob/3a88c05fae179cc467f6968362c0f00077261b16/src/main.rs#L65, which I think is closer to how the defauly `anyhow` formatter prints errors right? It looks something like
```
[ERROR] critical error: failed to assemble configuration settings
[ERROR]  -> failed to validate update-strategy configuration
[ERROR]  -> No such file or directory (os error 2)
```
To me, actually, both styles seem to make sense to me, so probably the most important thing is to be consistent. Based on what I've seen, it looks like in majority of other places we are doing `.context("failed to run 'rpm-ostree' binary")?;`.

lucab: Uhm yeah, we should probably do a project-wide sweep. Right now Zincati critical errors are logged like this:
```
[INFO ] starting update agent (zincati 0.0.19)
[ERROR] critical error: failed to assemble configuration settings
[ERROR]  -> failed to validate agent identity configuration
[ERROR]  -> unexpected negative rollout wariness: -0.002
```

But transient errors (e.g. from remote HTTP requests or from OS calls) are instead logged in the single-line format, like this:
```
[ERROR] failed to check Cincinnati for updates: client-side error: error sending request for url (https://updates.coreos.fedoraproject.org/...): error trying to connect: Connection reset by peer (os error 104)
```

I guess that contrary to my initial comment, sticking to the format of `failed to perform xyz` is the best way to get a uniform result.

lucab (approved):

Issue #518, opened 2021-04-01
cargo: update to latest actix + tokio + reqwest


lucab:
[Tokio 1.0](https://tokio.rs/blog/2020-12-tokio-1-0) has been released, and both actix and reqwest have been now ported to it. We can finally move Zincati to it.

This involves updating all the dependencies at once and possibly some API massaging, so it can't be performed by @dependabot. It will need a manual PR instead.

PR #536, opened 2021-04-22
Bump tokio, actix, reqwest

kelvinfan001:
Bump tokio to 1.5, actix to 0.11, reqwest to 0.11 and massage some APIs.

Closes https://github.com/coreos/zincati/issues/518

### DIFF
--- a/src/cli/agent.rs
+++ b/src/cli/agent.rs
@@ -37,23 +37,24 @@ pub(crate) fn run_agent() -> Result<()> {
     PROCESS_START_TIME.set(start_time.timestamp());
 
     trace!("creating actor system");
-    let sys = actix::System::builder()
-        .name(crate_name!())
-        .stop_on_panic(true)
-        .build();
+    let sys = actix::System::new();
 
-    trace!("creating metrics service");
-    let _metrics_addr = metrics::MetricsService::bind_socket()?.start();
+    sys.block_on(async {
+        trace!("creating metrics service");
+        let _metrics_addr = metrics::MetricsService::bind_socket()?.start();
 
-    trace!("creating rpm-ostree client");
-    let rpm_ostree_addr = rpm_ostree::RpmOstreeClient::start(1);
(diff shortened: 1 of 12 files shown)


src/cli/agent.rs
```diff
@@ -37,23 +37,24 @@ pub(crate) fn run_agent() -> Result<()> {
...
     trace!("creating actor system");
-    let sys = actix::System::builder()
-        .name(crate_name!())
-        .stop_on_panic(true)
-        .build();
+    let sys = actix::System::new();
 
-    trace!("creating metrics service");
-    let _metrics_addr = metrics::MetricsService::bind_socket()?.start();
+    sys.block_on(async {
```
kelvinfan001, line 42: Self-note: API massaging here mostly caused by breaking changes caused by https://github.com/actix/actix-net/pull/207 in `actix-rt`.

kelvinfan001:
Looks like the metrics service isn't started properly on the Unix-domain socket...

src/cli/agent.rs
lucab: I'm not sure if that is the cause of further issues, but all the address variables that were previously scoped to `main` are now only alive inside this `async` block.

kelvinfan001: That should be fine right, since they're moved into structs later.

src/metrics/mod.rs
```diff
@@ -51,12 +51,15 @@ impl Actor for MetricsService {
             .listener
             .try_clone()
             .expect("failed to clone metrics listener");
+        listener
+            .set_nonblocking(true)
```
kelvinfan001, line 55: This addresses https://github.com/coreos/zincati/pull/536#issuecomment-825335567.
In `tokio 0.3.2`, the `tokio::net::UnixListener::from_std()` method no longer helps us set Unix listeners to nonblocking, so we need to do it ourselves now.

kelvinfan001:
@lucab This is ready for another look now :)

PR #540, opened 2021-04-28
strategy: add new filesystem strategy

kelvinfan001:
Add a simple, flexible strategy that checks `/var/lib/zincati` for
a JSON file (optionally containing an expiry timestamp key) named
`allowfinalize.json` and allows update finalization if such file
exists and has the proper permissions.


src/main.rs
```diff
@@ -7,6 +7,9 @@
 extern crate fail;
 #[macro_use]
 extern crate prometheus;
+#[cfg(test)]
+#[macro_use]
+extern crate serde_json;
```
cgwalters, line 12: Note `extern crate` isn't necessary anymore in Rust 2018, and also macros can be imported just like other items (functions/structs/etc).

src/strategy/mod.rs
```diff
@@ -148,6 +157,12 @@ impl UpdateStrategy {
         let periodic = StrategyPeriodic::new(cfg)?;
         Ok(UpdateStrategy::Periodic(periodic))
     }
+
+    /// Build a new "filesystem" strategy.
+    fn new_filesystem() -> Self {
+        let filesystem = StrategyFilesystem::default();
+        UpdateStrategy::Filesystem(filesystem)
```
cgwalters, line 164:
```suggestion
        UpdateStrategy::Filesystem(Default::default())
```
would probably work too.

src/strategy/filesystem.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    #[test]
+    fn test_parse_finalization_marker() {
+        let tmp_dir = tempdir().unwrap();
+        let test_marker_file_path = tmp_dir.path().join("allowfinalize.json");
+        let test_marker_file_path = test_marker_file_path.to_str().unwrap();
+        // 1619640863 is Apr 28 2021 20:14:23 UTC.
+        // Expect this to be expired.
+        let json = "
```
cgwalters, line 148: See https://docs.serde.rs/serde_json/value/index.html#constructing-json

src/strategy/filesystem.rs
```diff
@@ -0,0 +1,180 @@
...
+            \"notAfter\": 1619640863
+        }";
+        fs::write(test_marker_file_path, json).expect("could not write test finalization file");
+        let expired = is_expired(test_marker_file_path);
+        let runtime = rt::Runtime::new().unwrap();
+        let expired = runtime.block_on(expired).unwrap();
+        assert_eq!(expired, true);
+
+        // Expect timepstamp with value `u64::MAX` to not be expired.
+        let json = serde_json::to_string(&json!({ "notAfter": u64::MAX })).unwrap();
```
cgwalters, line 159: Ah I see you are using `json!` here, why not above?

src/strategy/filesystem.rs
```diff
@@ -0,0 +1,180 @@
...
+        fs::write(test_marker_file_path, json).expect("could not write test finalization file");
```
cgwalters, line 160: Not that it matters in a test, but it's more efficient to serialize to a file directly.

src/strategy/filesystem.rs
```diff
@@ -0,0 +1,180 @@
...
+    let attr = tokio_fs::metadata(path).await;
+    // If `path` doesn't exist, return false early.
+    if attr.is_err() {
+        return Ok(false);
+    }
+    let attr = attr.unwrap();
+    let mode = attr.permissions().mode();
+
+    // Octal 100644 represents regular file, rw-r--r--.
+    if mode != 0o100644 {
```
cgwalters, line 75: I may have missed it, why are we strictly requiring 0644?

src/strategy/filesystem.rs
```diff
@@ -0,0 +1,180 @@
...
+        }
+        None => Ok(false),
+    }
+}
+
+#[context("failed to parse expiry timestamp from marker file")]
+async fn parse_expiry_timestamp(path: &str) -> Result<Option<u64>> {
+    let data = tokio_fs::read_to_string(path).await?;
+    let marker_json: FinalizationMarker =
+        serde_json::from_str(&data).context("failed to parse JSON content")?;
```
cgwalters, line 110: Again this doesn't matter here but with nontrivial data it's better to deserialize streaming; see https://github.com/serde-rs/json/issues/316 which can be worked easily enough in most cases (like this one) deserializing in a `spawn_blocking(||)`.

Makefile
```diff
@@ -18,6 +18,7 @@ build:
 
 .PHONY: install
 install: build
+	mkdir -p ${DESTDIR}/var/lib/zincati
```
cgwalters, line 21: (Side note; rpm-ostree will convert this to a `tmpfiles.d` snippet)

cgwalters (commented):
Seems sane to me!

lucab:
I'd like to flush out the current pile of changes into a release before getting into reviewing this.

Additionally, this should come with some docs for users.

kelvinfan001:
@lucab yep, still working on the docs and possibly some kola tests. (sorry forgot to apply `on-hold` label).

src/strategy/filesystem.rs
```diff
@@ -0,0 +1,180 @@
...
+    not_after: Option<u64>,
+}
+
+/// Strategy for immediate updates.
+#[derive(Clone, Debug, Default, Serialize)]
+pub(crate) struct StrategyFilesystem {}
+
+impl StrategyFilesystem {
+    /// Strategy label/name.
+    pub const LABEL: &'static str = "filesystem";
```
lucab, line 28: I'd try to find something more descriptive. Ideally the name would convey at least some of following key concepts:
 * it's a local marker (i.e. it has no fleet-wide coordination)
 * it defines a time-window
 * it can persist across reboots
 * it's a level trigger (i.e. it can hold into a state, and can be queried over time)
 * it's baked by a regular file

src/strategy/filesystem.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// Local filesystem path to finalization marker file.
+    pub const FINALIZATION_MARKER_FILE_PATH: &'static str = "/var/lib/zincati/allowfinalize.json";
```
lucab, line 30: Pending the final strategy name, I'd nest this a bit deeper so that the path hierarchy can clearly split out that this is 1) userdata input, and 2) belonging to a specific strategy.

src/strategy/filesystem.rs
```diff
@@ -0,0 +1,180 @@
...
+/// Verify that finalization marker file exists, is a regular file,
+/// has the correct permissions, and is owned by the root user.
+#[context("failed to verify finalization marker file metadata")]
+async fn verify_file_metadata(path: &str) -> Result<bool> {
+    let attr = tokio_fs::metadata(path).await;
+    // If `path` doesn't exist, return false early.
+    if attr.is_err() {
+        return Ok(false);
+    }
+    let attr = attr.unwrap();
```
lucab, line 71: You can hoist this into the err-check above, using a `match` to avoid the unwrapping.

src/strategy/filesystem.rs
kelvinfan001: I thought this would be to prevent non-root users from being able to modify it, since this directly affects reboots.

src/strategy/filesystem.rs
kelvinfan001: How about `marker_file` strategy.  Conveys first point and last point and pretty simple to understand :)

src/strategy/filesystem.rs
kelvinfan001: `/var/lib/zincati/admin/marker_file_strategy/allowfinalization.json`?

src/strategy/mod.rs
kelvinfan001: hmm, I was trying to keep it consistent with how the other strategies are created.

src/strategy/filesystem.rs
kelvinfan001: I've just put a block of synchronous code in a block and put it inside a `spawn_blocking()`. Am I allowed to do this? 😅

src/strategy/filesystem.rs
kelvinfan001: I used it here because I wanted to use `u64::MAX` more easily. But yeah I think I should've used it above too! :)

src/strategy/filesystem.rs
kelvinfan001: Thanks for pointing these things out. Forced me learn about how reading files actually works.

kelvinfan001:
Notes:
After this basic functionality is merged, we should consider adding an inotify watch on the `allowfinalize.json` file to tick now so reboots happen immediately if an update is staged.
Also consider extending the marker file with an additional `persistAfterReboot` key where Zincati will remove the file before rebooting if false.

src/strategy/marker_file.rs
```diff
@@ -0,0 +1,227 @@
...
+        Ok(attr) => attr,
+        // If `path` doesn't exist, return false early.
+        Err(_) => return Ok(false),
+    };
+
+    if !attr.is_file() {
+        anyhow::bail!("file is not regular file");
+    }
+
+    ensure_owned_by_root(&attr)?;
```
kelvinfan001, line 82: Do we still need this check if `/var/lib/zincati` is owned by the `zincati` user already anyway?

src/strategy/filesystem.rs
lucab: `/var/lib/zincati/admin/strategy/<$NAME>/allowfinalization.json` I'd say.

docs/usage/updates-strategy.md
```diff
@@ -65,6 +65,53 @@ Such an approach is only recommended where nodes are already grouped into an orc
...
+    - named `allowfinalize.json`
+    - under `/var/lib/zincati/admin/strategy/marker_file`
+    - is a valid JSON file
+    - owned by root
+    - not writable by others
+
+If any of the above is not satisfied in your marker file, Zincati will not allow reboots.
+
+`allowfinalize.json` can optionally contain a `allowUntil` key with a Unix timestamp integer as its value to indicate the expiry date and time of this marker file. If the current time timestamp is _greater than or equal to_ this timestamp, then reboots will not be allowed.
+Otherwise, if the `allowUntil` key is not present, reboots will be allowed for as long as `allowfinalize.json` exists (in the right location), and it must be removed to disallow reboots.
```
dustymabe, line 84:
```
Otherwise, if the `allowUntil` key is not present, reboots will be allowed for as long as `allowfinalize.json` exists (in the right location), and it must be removed to disallow reboots.
```

Can we get an example of that? is it just `echo '{}' | sudo tee /var/lib/zincati/admin/strategy/marker_file/allowfinalize.json` ?

docs/usage/updates-strategy.md
```diff
@@ -65,6 +65,53 @@ Such an approach is only recommended where nodes are already grouped into an orc
...
+Note that `allowfinalize.json` must still be a valid JSON file, regardless of whether the `notallowUntilAfter` key is present.
```
dustymabe, line 85:
```
Note that `allowfinalize.json` must still be a valid JSON file, regardless of whether the `notallowUntilAfter` key is present.
```

Is there an example using `notallowUntilAfter`?

docs/usage/updates-strategy.md
```diff
@@ -65,6 +65,53 @@ Such an approach is only recommended where nodes are already grouped into an orc
...
 
+# File-based strategy
+The `marker_file` strategy is a simple, low-level strategy that only allows Zincati to reboot for updates when a specific marker file exists on the local filesystem.
+
+Similar to the `fleet_lock` strategy, the `marker_file` strategy provides a large amount of flexibility to admins, and can be used with a central controller.
+Unlike `fleet_lock`, where the central controller must be a lock-manager on the network, the central controller for the `marker_file` strategy can be a containerized agent, some central task manager able to manipulate files on machines (e.g. Ansible), or even a human via SSH.
+
+To indicate that a machine is allowed to finalize an update and reboot, a file with the following properties must be present on the machine's local filesystem:
+    - named `allowfinalize.json`
+    - under `/var/lib/zincati/admin/strategy/marker_file`
```
dustymabe, line 76:
```
    - named `allowfinalize.json`
    - under `/var/lib/zincati/admin/strategy/marker_file`
```

I wonder how this will work if you have multiple containers that might need to control reboot. If they are all trying to edit a single file then it gets problematic. Could or should we support multiple files under here?

docs/usage/updates-strategy.md
kelvinfan001: Sorry, that was a typo. So far there is only an `allowUntil` key.

docs/usage/updates-strategy.md
kelvinfan001: Yep, `echo '{}' | sudo tee /var/lib/zincati/admin/strategy/marker_file/allowfinalize.json` would work. I'll add an example :)

docs/usage/updates-strategy.md
lucab: That's a legit point but I'd say it's outside of the scope of this strategy. Specifically, here we have a bias towards a single controller who can be audited/blamed for finalization decisions.

Multiple un-coordinated agents are a likely interesting case but it possibly needs either 1) higher-level coordination, or 2) a different approach (i.e. some kind of rule to determine who makes the final go/no-go call).

docs/usage/updates-strategy.md
```diff
@@ -65,6 +65,53 @@ Such an approach is only recommended where nodes are already grouped into an orc
...
+    - is a valid JSON file
+    - owned by root
+    - not writable by others
+
+If any of the above is not satisfied in your marker file, Zincati will not allow reboots.
+
+`allowfinalize.json` can optionally contain an `allowUntil` key with a Unix timestamp integer as its value to indicate the expiry date and time of this marker file. If the current time timestamp is _greater than or equal to_ this timestamp, then reboots will not be allowed.
+Otherwise, if the `allowUntil` key is not present, reboots will be allowed for as long as `allowfinalize.json` exists (in the right location), and it must be removed to disallow reboots.
```
lucab, line 84: Thinking aloud: do we need this to be optional right now? I am not fully sold on this way to say "allow all finalizations forever". Should we maybe start with making this mandatory and relax this to an `Option` later on if there is a need to?

docs/usage/updates-strategy.md
```diff
@@ -65,6 +65,53 @@ Such an approach is only recommended where nodes are already grouped into an orc
...
+```
+
+The above JSON file can be created using `jq` by entering the following command:
+
+```bash
+echo '"2021-05-01T00:00:00Z"' | jq '{allowUntil: 'fromdate'}' \
+| sudo tee /var/lib/zincati/admin/strategy/marker_file/allowfinalize.json
+```
+
+Warning: In `jq` versions `1.6` and lower, `jq` [may output incorrect Unix timestamps][jq_bug] for certain datetimes on machines with certain `localtime`s.
```
lucab, line 102: This sounds scary enough, do we want `fromdateiso8601` maybe instead?

src/strategy/marker_file.rs
lucab: No, we trust systemd-tmpfiles for that (also, it's probably fine if the hierarchy is owned by `root` either).

lucab:
Dusty was commenting out-of-band that the name here is still somehow suboptimal, possibly because still too generic.
I do share the same feeling, but I didn't come up with better ideas so far.

/cc @dustymabe

lucab:
@kai-uwe-rommel we do envision this strategy (name still to be finalized) to provide the basic primitive for a central manager (like Ansible) pushing finalization markers to individual hosts.
I tried to put together a very simple playbook to see how the integration could look like, at https://github.com/lucab/ansible-zincati-marker-file (no fancy stuff, just basic file writing/deleting). 
Can you give a look at all of this and see if it would generally fit into your expected flow?

dustymabe:
> Dusty was commenting out-of-band that the name here is still somehow suboptimal, possibly because still too generic.
> I do share the same feeling, but I didn't come up with better ideas so far.

Yes. We were discussing this in the context of default allow versus default disallow (inhibit). This name is only sub-optimal if it doesn't cover both of those scenarios (which it currently doesn't) and we want to add another strategy later for the default to allow case in which case we might want to make it more obvious what the difference is between the two strategies.

Random ideas:

- `green_file` (green light on a stop light for "go") - `red_file` in the future if we want to support disallowing update via file strategy
- `allow_file`/`disallow_file`

I'm not really in love with any of them.

kelvinfan001:
> We were discussing this in the context of default allow versus default disallow (inhibit). This name is only sub-optimal if it doesn't cover both of those scenarios (which it currently doesn't) and we want to add another strategy later for the default to allow case in which case we might want to make it more obvious what the difference is between the two strategies.

Sorry if this had already been discussed/clarified before elsewhere, but I still feel like default allow and default disallow is not too important here for this strategy. In theory, if there were a reliable single controller that is in charge of writing and removing a marker file, then the "defaults" for this strategy is totally up to the controller, right? I do recall vaguely it being mentioned that it is easier to touch a file than remove a file, and vice versa, and so there might be some preference around whether we would like to default to Zincati being stuck in either constantly allowing reboots or constantly denying reboots. But I feel like this is a separate issue.

I think the key point for this strategy, in my opinion, is the emphasis that there should be a single controller that is the single decider on whether finalizations are allowed. If multiple processes/containers wanted to express their intent to allow/disallow reboots, that should be communicated to the central controller, rather than to Zincati's update strategy directly. Let me know if I've misunderstood anything :)

docs/usage/updates-strategy.md
kelvinfan001: I was thinking that the ability to leave a (valid) file in the directory forever would allow for the use case of "allow reboots for the most part unless I need to disallow it for a while". So for example, if theoretically Dusty's containers had some central controller, that controller could easily make use of this new strategy.

docs/usage/updates-strategy.md
kelvinfan001: I think `fromdateiso8601` is the same as `fromdate` right now, but right, I think `fromdateiso8601` would be more explicit.

kai-uwe-rommel:
> 
> 
> @kai-uwe-rommel we do envision this strategy (name still to be finalized) to provide the basic primitive for a central manager (like Ansible) pushing finalization markers to individual hosts.
> I tried to put together a very simple playbook to see how the integration could look like, at https://github.com/lucab/ansible-zincati-marker-file (no fancy stuff, just basic file writing/deleting).
> Can you give a look at all of this and see if it would generally fit into your expected flow?

Thanks. Yes, that looks good, i.e. practical. It would allow us to explicitly upgrade a bunch of machines via a central action (Ansible) within a defined maintenance window.
In our case, these machines are load balancers and NFS servers for OCP/OKD clusters.
So we need to strictly adhere to defined maintenance windows.
Currently, we maintain the upgrades of the machines manually ...

kelvinfan001:
@lucab I think this should be ready for another look as well.

dustymabe:
> @lucab I think this should be ready for another look as well.

bump

mysticaltech:
Will be very useful! Keep up the good work folks.

dustymabe:
I'd like to revive this at some point. The PR was closed but it doesn't say who or why it was closed.

jlebon:
> I'd like to revive this at some point. The PR was closed but it doesn't say who or why it was closed.

I think what happened there is that the fork was closed, so GitHub auto-closed all PRs from it.

PR #543, opened 2021-04-30
update_agent: log deploy attempt outcome

lucab:
This logs the result of a deploy attempt, which is currently visible
as service status only.


lucab:
/cc @kelvinfan001

This may hopefully help tracking down https://github.com/coreos/fedora-coreos-tracker/issues/756.

kelvinfan001 (approved):
LGTM

PR #547, opened 2021-05-07
update_agent: fix logic around reboot delays

kelvinfan001:
The previous logic incorrectly causes the final reboot postponement
to be postponed for `DEFAULT_REFRESH_PERIOD_SECS`; more importantly,
it causes refresh tick delays of length
`DEFAULT_POSTPONEMENT_TIME_SECS` if update finalization is denied
due to update strategy constraints, potentially leading to much
more frequent finalization attempts.

We actually want the opposite of this behaviour.

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -85,8 +85,7 @@ impl Handler<RefreshTick> for UpdateAgent {
         };
 
         let update_machine = state_action.then(move |_r, actor, ctx| {
-            if let Some(interval) = actor.refresh_delay(prev_state) {
-                let pause = Self::add_jitter(interval);
+            if let Some(pause) = actor.refresh_delay(prev_state) {
                 log::trace!(
                     "scheduling next agent refresh in {} seconds",
                     pause.as_secs()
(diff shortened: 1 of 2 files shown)


kelvinfan001:
Only discovered this when playing around with version 34.20210427.1.0 when it was upgrading to 34.20210503.1.0.
I think more e2e tests around this type of logic would've prevented something like this :)

src/update_agent/actor.rs
```diff
@@ -141,8 +142,10 @@ impl UpdateAgent {
                 self.steady_interval
             }
             UpdateAgentState::UpdateStaged((_, postponements)) => {
-                // Note: no jitter if postponing reboots.
```
lucab, line 144: I think this method outgrew its initial small logic and it's now in need of some unit tests.

I'd suggest to split out the discriminant and state check into its own function (only needing previous and current states). Then it can be tested by manually driving through states, like this other existing test: https://github.com/coreos/zincati/blob/v0.0.20/src/update_agent/mod.rs#L497

kelvinfan001: I factored them out as independent associated functions for ease of unit testing just those two functions within `update_agent/actor.rs`.

src/update_agent/actor.rs
```diff
@@ -124,33 +124,52 @@ impl UpdateAgent {
...
         use std::mem::discriminant;
 
         // State changes trigger immediate tick/action.
-        if discriminant(&prev_state) != discriminant(&self.state) {
+        if discriminant(prev_state) != discriminant(cur_state) {
             // Unless we're transitioning from ReportedSteady to NoNewUpdate.
-            if !(prev_state == UpdateAgentState::ReportedSteady
-                && self.state == UpdateAgentState::NoNewUpdate)
+            if *prev_state == UpdateAgentState::ReportedSteady
+                && *cur_state == UpdateAgentState::NoNewUpdate
```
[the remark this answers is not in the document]
kelvinfan001: Haha, I just reread this and I have no idea why I changed it to this complicated mess. I think I might've been trying to debug some of the logic here and confused myself.

src/update_agent/actor.rs
```diff
@@ -85,8 +85,7 @@ impl Handler<RefreshTick> for UpdateAgent {
         };
 
         let update_machine = state_action.then(move |_r, actor, ctx| {
-            if let Some(interval) = actor.refresh_delay(prev_state) {
-                let pause = Self::add_jitter(interval);
```
lucab, line 89: Just wondering, were we previously applying jittering twice due to this?

kelvinfan001: Sorry I missed this earlier, I believe we were applying jittering twice :(

lucab (approved):

PR #552, opened 2021-05-13
main: update systemd StatusText on SIGTERM

kelvinfan001:
Currently, when we stop zincati.service, the systemd StatusText
remains as whatever it was before it was stopped. This is
somewhat misleading when the StatusText is shown in e.g.
`rpm-ostree status`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ serde_json = "1.0"
 structopt = "0.3"
 tempfile = "^3.2"
 thiserror = "1.0"
-tokio = { version = "1.5", features = ["rt", "rt-multi-thread"] }
+tokio = { version = "1.5", features = ["signal", "rt", "rt-multi-thread"] }
 toml = "0.5"
 tzfile = "0.1.3"
 url = { version = "2.2", features = ["serde"] }
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,6 +27,8 @@ mod rpm_ostree;
 mod strategy;
 /// Update agent.
 mod update_agent;
+/// Utility functions.
+mod utils;
 /// Logic for weekly maintenance windows.
 mod weekly;
 
--- a/src/cli/agent.rs
+++ b/src/cli/agent.rs
@@ -1,12 +1,13 @@
 //! Logic for the `agent` subcommand.
 
 use super::ensure_user;
-use crate::{config, dbus, metrics, rpm_ostree, update_agent};
+use crate::{config, dbus, metrics, rpm_ostree, update_agent, utils};
 use actix::Actor;
 use anyhow::{Context, Result};
 use log::{info, trace};
 use prometheus::IntGauge;
 use structopt::clap::{crate_name, crate_version};
+use tokio::runtime::Runtime;
 
 lazy_static::lazy_static! {
     static ref PROCESS_START_TIME: IntGauge = register_int_gauge!(opts!(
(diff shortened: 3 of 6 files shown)


Cargo.toml
```diff
@@ -14,6 +14,7 @@ actix = "0.11"
 anyhow = "1.0"
 cfg-if = "1.0"
 chrono = { version = "0.4", features = ["serde"] }
+ctrlc = { version = "3.0", features = ["termination"] }
```
cgwalters, line 17: I think it'd be more natural to use https://docs.rs/tokio/1.5.0/tokio/signal/index.html

cgwalters (commented):
One minor comment, but seems good as is too.
Nice work on this and the tests!

It's a bit surprising though that systemd retains the unit status literally after a unit is stopped.  Hmm actually, it seems that `systemctl show` on a stopped unit doesn't display its last unit status by default.

So...is the bug just that e.g. rpm-ostree is showing the status of a stopped unit?

kelvinfan001:
@cgwalters hmm, I just tried again with `systemctl show` on a stopped Zincati and it still shows the StatusText.
```
$ systemctl show -p StatusText zincati.service
StatusText=periodically polling for updates (last checked Thu 2021-05-13 22:00:08 UTC)
$ sudo systemctl stop zincati.service 
$ systemctl show -p StatusText zincati.service
StatusText=periodically polling for updates (last checked Thu 2021-05-13 22:00:08 UTC)
$ systemctl status zincati.service 
○ zincati.service - Zincati Update Agent
     Loaded: loaded (/usr/lib/systemd/system/zincati.service; enabled; vendor preset: enabled)
     Active: inactive (dead) since Thu 2021-05-13 22:00:27 UTC; 20s ago
       Docs: https://github.com/coreos/zincati
    Process: 1219 ExecStart=/usr/libexec/zincati agent ${ZINCATI_VERBOSITY} (code=killed, signal=TERM)
   Main PID: 1219 (code=killed, signal=TERM)
     Status: "periodically polling for updates (last checked Thu 2021-05-13 22:00:08 UTC)"
```

cgwalters:
Ah heh, I tested with rpm-ostree which indeed today has a `SIGTERM` handler.  I bet it's this bit that's clearing the status: `      sd_notify (FALSE, "STOPPING=1");` which would make sense instead of resetting the status text if true (because that cleanly tells systemd we're shutting down).

kelvinfan001:
ooh, right, thanks for pointing out `sd_notify (FALSE, "STOPPING=1")`, I think we should do this here, as well. 
However, the StatusText is still showing after notifying systemd that we're stopping. According to the man page, "This is useful to allow the service manager to track the service's internal state, and present it to the user."
With `sd_notfy(FALSE, "STOPPING=1")`, I get:
```
$ systemctl status zincati.service 
● zincati.service - Zincati Update Agent
     Loaded: loaded (/usr/lib/systemd/system/zincati.service; enabled; vendor preset: enabled)
     Active: inactive (dead) since Thu 2021-05-13 22:25:28 UTC; 4s ago
       Docs: https://github.com/coreos/zincati
    Process: 1204 ExecStart=/usr/libexec/zincati agent ${ZINCATI_VERBOSITY} (code=exited, status=0/SUCCESS)
   Main PID: 1204 (code=exited, status=0/SUCCESS)
     Status: "periodically polling for updates (last checked Thu 2021-05-13 22:25:16 UTC)"
```
The status is still there, but at least it shows that we exited successfully. Will investigate this a bit more.

src/main.rs
```diff
@@ -35,6 +35,11 @@ use structopt::StructOpt;
 
 /// Binary entrypoint, for all CLI subcommands.
 fn main() {
+    ctrlc::set_handler(move || {
```
lucab, line 38: This seems to be placed in the wrong place. It should belong somewhere under `run_agent` I guess?

kelvinfan001: ah, that makes sense, since this is specifically dealing with users attempting to stop the daemon.
I put it in main originally because I just thought we wanted to set up a handler as early as possible.

cgwalters (approved):

PR #556, opened 2021-05-17
update_agent: reword reboot delay broadcast mesage

kelvinfan001:
Reword message broadcasted to users when reboots are delayed due
to active interactive user sessions to emphasize that logging out
active sessions will allow Zincat to continue the auto-update
process.

### DIFF
--- a/src/update_agent/mod.rs
+++ b/src/update_agent/mod.rs
@@ -473,7 +473,10 @@ fn format_reboot_warning(seconds: u64, release_ver: &str) -> String {
     let time_till_reboot = format_seconds(seconds);
 
     format!(
-        "New update {} deployed.\nRebooting into this update in around {} (if permitted by update strategy).",
+        "New update {} is available and has been deployed.\n\
+        If permitted by the update strategy, Zincati will reboot into this update when all\n\
+        interactive users have logged out, or in {}, whichever comes earlier.\n\
+        Please log out of all active sessions in order to let the auto-update process continue.",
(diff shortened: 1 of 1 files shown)


src/update_agent/mod.rs
```diff
@@ -473,7 +473,9 @@ fn format_reboot_warning(seconds: u64, release_ver: &str) -> String {
     let time_till_reboot = format_seconds(seconds);
 
     format!(
-        "New update {} deployed.\nRebooting into this update in around {} (if permitted by update strategy).",
+        "New update {} is available and will be automatically applied.\n\
+        Finalization delayed until all interactive users have logged out (or for up to {}), update strategy permitting.\n\
+        Please log out all active sessions in order to let the auto-update process continue.",
```
dustymabe, line 478: That second sentence might be a little long for a single line. Suggestion:
```suggestion
        "New update {} is available and has been deployed. Finalization has been delayed\n\
        for up to {} (update strategy permitting) or until all interactive users have\n\
        logged out. Please log out all active sessions in order to let the auto-update\n\
        process continue.",
```

kelvinfan001: Thanks, I've updated the PR to make the lines roughly the same length (also roughly accounting for what goes inside the `{}`s). I've also made it a bit more verbose for extra clarity.
Let me know if the lines are still too long or if it's getting too verbose now.

dustymabe (approved):
LGTM

PR #557, opened 2021-05-18
cargo/release: set sign-tag property

lucab:
This sets the new `sign-tag` metadata field which has been recently
by cargo-release.


kelvinfan001 (approved):
LGTM

PR #559, opened 2021-05-19
rpm_ostree/actor: add traces when CLI returns

kelvinfan001:
This should help us determine when and whether calls to the
rpm-ostree CLI returns; this, together with
https://github.com/coreos/zincati/pull/543, should help us narrow
down whether the `update_agent` actor's apparent freeze during a
https://github.com/coreos/fedora-coreos-tracker/issues/756 test
flake is due to a non-returning rpm-ostree CLI or communication
issues between the `update_agent` actor and the `rpm_ostree` actor.

### DIFF
--- a/src/rpm_ostree/actor.rs
+++ b/src/rpm_ostree/actor.rs
@@ -53,7 +53,9 @@ impl Handler<StageDeployment> for RpmOstreeClient {
 
     fn handle(&mut self, msg: StageDeployment, _ctx: &mut Self::Context) -> Self::Result {
         trace!("request to stage release: {:?}", msg.release);
-        super::cli_deploy::deploy_locked(msg.release, msg.allow_downgrade)
+        let release = super::cli_deploy::deploy_locked(msg.release, msg.allow_downgrade);
+        trace!("rpm-ostree CLI returned: {:?}", release);
+        release
     }
 }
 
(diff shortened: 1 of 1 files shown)


lucab (approved):

src/rpm_ostree/actor.rs
```diff
@@ -53,7 +53,9 @@ impl Handler<StageDeployment> for RpmOstreeClient {
 
     fn handle(&mut self, msg: StageDeployment, _ctx: &mut Self::Context) -> Self::Result {
         trace!("request to stage release: {:?}", msg.release);
-        super::cli_deploy::deploy_locked(msg.release, msg.allow_downgrade)
+        let release = super::cli_deploy::deploy_locked(msg.release, msg.allow_downgrade);
+        trace!("rpm-ostree CLI returned: {:?}", release);
```
cgwalters, line 57: It may be good at some point to investigate using the [tracing crate](https://crates.io/crates/tracing) for zincati.  It's much better for async in particular.

Hmm wait, isn't something logging this `Result` if it's not `Ok` somewhere already?

kelvinfan001: There is currently logging at the `update_agent` actor side. I was mainly trying to figure out whether rpm-ostree CLI returned at all, or whether messages aren't being sent properly across actors, since I wasn't able to tell from the logs.

PR #568, opened 2021-05-31
dist: fix dbus configuration

lucab:
This tweaks dbus configuration, fixing some mistakes in the ACLs.
It gets rid of the following dbus-broker warning:
> Conflicting attributes in /usr/share/dbus-1/system.d/org.coreos.zincati.conf +14


kelvinfan001 (approved):
Thanks for noticing this. LGTM!

Issue #544, opened 2021-05-03
strategy: add metrics around finalization requests/results


lucab:
As Zincati is growing more strategies and as they get more complex, chances of buggy or mis-configured strategies also get larger.
We should add some metrics around [`can_finalize`](https://github.com/coreos/zincati/blob/v0.0.20/src/strategy/mod.rs) so that we can graph:
 * total number of allowed finalization requests
 * total number of errors
 * amount of "false" vs "true" results

PR #571, opened 2021-05-31
Add more metrics around finalization attempts

kelvinfan001:
This adds metrics to measure the total number of finalization
attempts, `can_finalize` responses from  update strategy, and
attempts denied due to factors unrelated to update strategy,
such as active user sessions, or (in the future) systemd inhibitor
locks present.

Note: this is in addition to and distinct from the
`zincati_rpm_ostree_finalize_attempts_total`
`zincati_rpm_ostree_finalize_failures_total` metrics, as those two
metrics are solely related to the `rpm_ostree` actor and rpm-ostree
binary.

Closes https://github.com/coreos/zincati/issues/544.

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -7,14 +7,30 @@ use actix::prelude::*;
 use anyhow::Error;
 use futures::prelude::*;
 use log::trace;
-use prometheus::IntGauge;
+use prometheus::{IntCounter, IntCounterVec, IntGauge};
 use std::collections::BTreeSet;
 use std::time::Duration;
+
+/// Label for finalization attempts blocked due to active interactive user sessions.
+pub static ACTIVE_USERSESSIONS_LABEL: &str = "active_usersessions";
 
 lazy_static::lazy_static! {
     static ref LAST_REFRESH: IntGauge = register_int_gauge!(opts!(
         "zincati_update_agent_last_refresh_timestamp",
         "UTC timestamp of update-agent last refresh tick."
+    )).unwrap();
+    static ref FINALIZATION_ATTEMPTS: IntCounter = register_int_counter!(opts!(
+        "zincati_update_agent_finalization_attempts",
+        "Total number of attempts to finalize a staged deployment by the update agent."
+    )).unwrap();
+    static ref FINALIZATION_BLOCKED: IntCounterVec = register_int_counter_vec!(
+        "zincati_update_agent_finalization_blocked_count",
+        "Total number of finalization attempts blocked due to reasons unrelated to update strategy.",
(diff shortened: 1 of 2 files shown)


src/update_agent/actor.rs
```diff
@@ -16,6 +16,26 @@ lazy_static::lazy_static! {
         "zincati_update_agent_last_refresh_timestamp",
         "UTC timestamp of update-agent last refresh tick."
     )).unwrap();
+    static ref FINALIZATION_ATTEMPTS: IntCounter = register_int_counter!(opts!(
+        "zincati_update_agent_finalization_attempts",
+        "Total number of attempts to finalize a staged deployment by the update agent."
+    )).unwrap();
+    static ref FINALIZATION_ALLOWED_BY_STRATEGY: IntCounter = register_int_counter!(opts!(
```
lucab, line 23: I think this allowed/denied can be folded into a single metrics with a `outcome=allowed|denied` label and moved into the `strategy` module. I'm not sure whether to fold the error outcome into this label as well.

src/update_agent/actor.rs
```diff
@@ -16,6 +16,26 @@ lazy_static::lazy_static! {
...
+        "zincati_update_agent_finalization_allowed_by_strategy",
+        "Total number of finalization attempts by the update agent allowed by its update strategy."
+    )).unwrap();
+    static ref FINALIZATION_DENIED_BY_STRATEGY: IntCounter = register_int_counter!(opts!(
```
lucab, line 27: These `denied` cases can also likely be folded into labels.

How do you plan to fit inhibitors into the flow (and thus metrics counting)?

src/update_agent/actor.rs
kelvinfan001: By "folding", you mean using e.g. `IntCounterVec` right?

src/update_agent/actor.rs
kelvinfan001: I was thinking of making a separate metric, similar to `zincati_update_agent_finalization_denied_by_usersessions` to count denials due to systemd inhibitors. Should we combine all non strategy-related gates into one?

src/update_agent/actor.rs
lucab: Yes. There are some examples in this project in the `fleet_lock` module.

src/update_agent/actor.rs
lucab: Yes, I'd combine them together via labels.

kelvinfan001:
@lucab I think this is good for another look if you have time :)

lucab (approved):

Issue #569, opened 2021-05-31
Allow concurrent handling of messages to `UpdateAgent` actor


kelvinfan001:
Currently, the `UpdateAgent` actor handles all messages synchronously in the order in which the messages are received. This works well for the most part currently as the only messages it handles are `RefreshTick` and an experimental `LastRefresh` (mainly for testing purposes).
However, in the future where we have more meaningful D-Bus methods that would like to access the state of the `UpdateAgent` actor, D-Bus methods should be allowed to be handled concurrently with the `UpdateAgent` actor's regular ticking. Therefore, we should allow messages sent to the `UpdateAgent` actor to be handled concurrently, e.g. calling `spawn`, instead of `wait` on its Actix `context`. However, we would still like to have `RefreshTick` messages handled in sequence. One possible way to achieve this is to use a [future-aware lock](https://docs.rs/tokio/1.6.1/tokio/sync/struct.RwLock.html) to lock `actor.state`.

xref: https://github.com/coreos/zincati/pull/514#discussion_r619567763

PR #577, opened 2021-06-10
update_agent: put state behind RwLock

kelvinfan001:
We would like to put update agent state behind a RwLock so ticks
can be handled sequentially while allowing messages to the
`update_agent` actor to be handled concurrently, in general.

Closes: https://github.com/coreos/zincati/issues/569

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -1,6 +1,6 @@
 //! Update agent actor.
 
-use super::{UpdateAgent, UpdateAgentState};
+use super::{UpdateAgent, UpdateAgentInfo, UpdateAgentState};
 use crate::rpm_ostree::{self, Release};
 use crate::utils;
 use actix::prelude::*;
@@ -9,6 +9,7 @@ use futures::prelude::*;
 use log::trace;
 use prometheus::{IntCounter, IntCounterVec, IntGauge};
 use std::collections::BTreeSet;
+use std::sync::Arc;
 use std::time::Duration;
 
 /// Label for finalization attempts blocked due to active interactive user sessions.
@@ -40,7 +41,7 @@ impl Actor for UpdateAgent {
     fn started(&mut self, ctx: &mut Self::Context) {
         trace!("update agent started");
 
-        if self.allow_downgrade {
+        if self.info.allow_downgrade {
             log::warn!("client configuration allows (possibly vulnerable) downgrades via auto-updates logic");
         }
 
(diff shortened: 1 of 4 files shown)


src/update_agent/actor.rs
```diff
@@ -100,7 +114,7 @@ impl Handler<RefreshTick> for UpdateAgent {
         });
 
         // Process state machine refresh ticks sequentially.
-        ctx.wait(update_machine);
```
lucab, line 103: I am unsure whether this `wait` should be converted into a `spawn`, or directly dropped by returning the `update_machine` chain. But I have a feeling this may be ok this way.

src/update_agent/actor.rs
```diff
@@ -73,52 +74,92 @@ impl Message for RefreshTick {
...
+            if let Some(pause) = pause {
                 log::trace!(
                     "scheduling next agent refresh in {} seconds",
                     pause.as_secs()
                 );
                 Self::tick_later(ctx, pause);
             } else {
                 let update_timestamp = chrono::Utc::now();
-                actor.state_changed = update_timestamp;
+                actor.info.state_changed = update_timestamp;
```
kelvinfan001, line 156: Not exactly related to this PR, but I noticed that the logic around updating this `state_changed` field is highly coupled with the logic around `refresh_delay()` which is in turn dependent on whether the `state` was actually changed. I think this made sense back in simpler times when `tick_now` if and only if change in state; however, there are now cases where a change in state does not necessarily `tick_now`.
Anyways, IIUC, I think the logic around updating this `state_changed` field should be refactored to a more intuitive location.

src/update_agent/actor.rs
```diff
@@ -73,52 +74,92 @@ impl Message for RefreshTick {
...
+        let state_action = async move {
+            let mut agent_state_guard = lock.write_owned().await;
+            // Consider `LAST_REFRESH` time to be when lock is acquired.
+            let tick_timestamp = chrono::Utc::now();
+            LAST_REFRESH.set(tick_timestamp.timestamp());
+
+            trace!("update agent tick, current state: {:?}", *agent_state_guard);
+            let prev_state = (*agent_state_guard).clone();
+
+            let modify_state_outcome = match prev_state.clone() {
```
kelvinfan001, line 92: The `clone()` here is because the left side of the match arms require a partial move; but we need to borrow `prev_state` later in this function

kelvinfan001:
Some groundwork that happened in this PR includes:
1. use async/await within the tick functions and their helpers
2. convert the ticks and some of their helpers to methods of the new `UpdateAgentInfo` instead of `UpdateAgent`

`1.` solved the issue of it being super cumbersome to change all the parameters and return types of helper functions to always take ownership of state and return ownership of state once they're done with it, and also makes the code cleaner to read, IMO
`2.` was necessary because the `async` block does not have access to (ownership of) an entire `UpdateAgent` instance; it now only has access to a clone of the "read-only" side of `UpdateAgent` and a lock guard to the `UpdateAgent`'s state.

src/update_agent/actor.rs
kelvinfan001: I've experimented with directly returning the `update_machine` chain in my toy example, and it behaves as expected, so I think dropping it is fine (no need for `ctx.spawn`).

src/update_agent/mod.rs
```diff
@@ -370,7 +372,7 @@ pub(crate) struct UpdateAgent {
     /// Update strategy.
     strategy: UpdateStrategy,
     /// Current status for agent state machine.
-    state: UpdateAgentState,
+    state: Arc<RwLock<UpdateAgentState>>,
```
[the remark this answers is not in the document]
kelvinfan001: (Also added as a comment in the code) I believe we need to use an `Arc` because consumers of this field will likely need to own it (e.g. consumers in futures).

src/update_agent/actor.rs
lucab: Just wondering, would this `state_changed` field fit more naturally into the state itself?

src/update_agent/actor.rs
```diff
@@ -73,52 +74,90 @@ impl Message for RefreshTick {
...
+                UpdateAgentState::UpdateFinalized(release) => {
+                    let update = release.clone();
+                    update_agent_info
+                        .tick_end(&mut *agent_state_guard, update)
+                        .await
+                }
+                UpdateAgentState::EndState => Ok(()),
+            };
+
+            if modify_state_outcome.is_err() {
```
lucab, line 134: It looks like most of the `tick_*` functions are internally handling error logging, so the error type here is not really used (in fact, it's a `()`). Do you think it would be possible to drop the `Result` at all? That would probably allow factoring our the `.await` too, I think.

src/update_agent/actor.rs
kelvinfan001: I thought about that, but `UpdateAgentState` is an enum. Would it be a good idea to attach the data to the enum the same way we attach e.g. `Release` to some of the variants?

lucab:
This is amazing! The side-effect of switching to plain `async` is possibly even nicer than the improved locking granularity :tada:
I've left a couple of comments related to possible low hanging fruits / cleanups, but the PR itself already looks good.

src/update_agent/actor.rs
kelvinfan001: Ah right, there are actually only a couple `tick`-functions that even make use of this `Result`. I was originally thinking to make minimal changes to the areas that didn't _have_ to be changed. Taking another look to see if not returning `Result` is doable.

src/update_agent/actor.rs
lucab: I was more thinking about the opposite: turning `state` into its own struct and making the enum just one of its field, next to the timestamp.
On a second thought, I'm not seeing right now what we are using this `state_changed` information for. Possibly it's unused at the moment, in which case you can either decide to drop it or to clarify its location and semantics (e.g. so that it could be sanely queried maybe via dbus).

lucab (approved):

src/update_agent/actor.rs
kelvinfan001: Follow up to this here https://github.com/coreos/zincati/pull/610

PR #599, opened 2021-07-09
rpm_ostree/deploy: block on driver registration

kelvinfan001:
Do not send a reply to the `update_agent` actor until the
`rpm_ostree` actor has successfully registered as driver. This
effectively blocks Zincati initialization on driver registration.
We consider failures to register as driver as transient failures,
so the `rpm_ostree` actor will keep retrying forever.

### DIFF
--- a/src/rpm_ostree/cli_deploy.rs
+++ b/src/rpm_ostree/cli_deploy.rs
@@ -4,6 +4,7 @@
 use super::Release;
 use anyhow::{bail, Context, Result};
 use prometheus::IntCounter;
+use std::time::Duration;
 
 const DRIVER_NAME: &str = "Zincati";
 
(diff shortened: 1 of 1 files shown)


src/rpm_ostree/cli_deploy.rs
```diff
@@ -31,20 +36,54 @@ pub fn deploy_locked(release: Release, allow_downgrade: bool) -> Result<Release>
...
-    invoke_cli_register()?;
+    let mut register_attempt = invoke_cli_register();
+    let mut retry_secs = 1;
+    while let Err(attempt) = register_attempt {
+        REGISTER_DRIVER_FAILURES.inc();
+        log::error!(
+            "{}\nretrying in {} second{}",
+            attempt,
+            retry_secs,
+            if retry_secs > 1 { "s" } else { "" }
```
cgwalters, line 50: I usually try to avoid doing English-specific tweaks like this; the simplest here is probably to do: `"retrying in {:?}", Duration::from_secs(retry_secs)` or so (which will print e.g. `5s` and `1s` etc.).

But once you do that it probably makes sense to have `retry_secs` just be an instance of `Duration` (note that it implements the `Add` and `Mul` traits).

(Eventually we might consider actually translating and using something like https://lib.rs/crates/gettext-rs but...that's a *big* topic)

cgwalters (approved):
Looks sane to me!

src/rpm_ostree/cli_deploy.rs
kelvinfan001: Ah, makes sense. Thanks for pointing that out!

PR #610, opened 2021-07-27
update_agent: remove unused `Result`s

kelvinfan001:
### update_agent: remove unused `Result`s

Previously, the `tick_` functions returned a `Result`, but only two
of these functions would ever return the `Err` variant. Even in the
case where the returned `Result` was of the `Err` variant, the caller
does not really do anything with it.

Since we already do a good amount of error logging in the inner
functions, refactor the `tick_` functions to return `()` instead.

### update_agent: misc cleanup
For helpers that return a `Result`, move error logging to the caller
functions.

Refactor `register_driver` to return `()`, because errors shouldn't
be returned since https://github.com/coreos/zincati/pull/599.

Use `anyhow::Result`.

Refactor `tick_finalize_update` to return early to avoid deeply nested
if-else statements.

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -4,7 +4,7 @@ use super::{UpdateAgent, UpdateAgentInfo, UpdateAgentState};
 use crate::rpm_ostree::{self, Release};
 use crate::utils;
 use actix::prelude::*;
-use anyhow::Error;
+use anyhow::{Error, Result};
 use futures::prelude::*;
 use log::trace;
 use prometheus::{IntCounter, IntCounterVec, IntGauge};
@@ -260,15 +260,12 @@ impl UpdateAgentInfo {
     async fn tick_initialize(&self, state: &mut UpdateAgentState) {
         trace!("update agent in start state");
         if self.enabled {
-            let register_outcome = self.register_as_driver().await;
-            if register_outcome.is_err() {
-                // In practice, we should not reach here: https://github.com/coreos/zincati/pull/599.
-                return;
-            };
+            self.register_as_driver().await;
         }
         let local_depls = self.local_deployments().await;
-        if let Ok(depls) = local_depls {
-            UpdateAgent::log_excluded_depls(&depls, &self);
+        match local_depls {
+            Ok(depls) => UpdateAgent::log_excluded_depls(&depls, &self),
+            Err(e) => log::error!("failed to query local deployments: {}", e),
         }
         let status;
         if self.enabled {
@@ -316,7 +313,10 @@ impl UpdateAgentInfo {
                     .fetch_update_hint(&self.identity, depls, allow_downgrade)
                     .await
             }
-            _ => None,
+            Err(e) => {
+                log::error!("failed to query local deployments: {}", e);
+                None
+            }
         };
 
         match release {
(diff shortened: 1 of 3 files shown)


src/update_agent/actor.rs
```diff
@@ -261,10 +257,14 @@ impl UpdateAgent {
 
 impl UpdateAgentInfo {
     /// Initialize the update agent.
-    async fn tick_initialize(&self, state: &mut UpdateAgentState) -> Result<(), ()> {
+    async fn tick_initialize(&self, state: &mut UpdateAgentState) {
         trace!("update agent in start state");
         if self.enabled {
-            self.register_as_driver().await?;
+            let register_outcome = self.register_as_driver().await;
```
lucab, line 263: It looks like this method as well may be moved to `async` and stop returning a `Result`, perhaps?

src/update_agent/actor.rs
```diff
@@ -128,13 +128,9 @@ impl Handler<RefreshTick> for UpdateAgent {
                         .tick_end(&mut *agent_state_guard, update)
                         .await
                 }
-                UpdateAgentState::EndState => Ok(()),
+                UpdateAgentState::EndState => (),
```
lucab, line 131: I think it could be possible to unify all the `.await` points above into a single one, perhaps after converting this one to an async closure.

src/update_agent/actor.rs
```diff
@@ -367,29 +357,22 @@ impl UpdateAgentInfo {
...
-        let state_change = if !strategy_can_finalize {
+        if !strategy_can_finalize {
             utils::update_unit_status(&format!(
                 "update staged: {}; reboot pending due to update strategy",
                 &release.version
             ));
             // Reset number of postponements to `MAX_FINALIZE_POSTPONEMENTS`
             // if strategy does not allow finalization.
             state.update_staged(release);
-            Err(())
```
lucab, line 392: As we are not throwing back errors from all branches anymore, it could be worth adding a direct early return here and in the `else` branch below, so that the bottom part of this function/logic is not anymore deeply nested into multiple if-else. I suspect the resulting code layout would be easier to read.

src/update_agent/actor.rs
```diff
@@ -402,28 +385,20 @@ impl UpdateAgentInfo {
                 ));
                 // Record postponement and postpone finalization.
                 state.record_postponement();
-                Err(())
-            } else {
-                self.finalize_deployment(release).await
+            } else if let Ok(release) = self.finalize_deployment(release).await {
```
lucab, line 388: This one is a strange mix where half of the result handing is done in the function itself (error logging) and half is done here at the caller (update on success).
It may be worth trying to move this one too to plain `async` too, and then decide whether to do the whole `Result` handling either fully in the function or fully here.

lucab (commented):
Code looks good, I think there are couple of directly related spots where the same cleanup approach can be applied.

src/update_agent/actor.rs
kelvinfan001: From https://users.rust-lang.org/t/expected-opaque-type-found-a-different-opaque-type-when-trying-futures-join-all/40596/4
Hmm so apparently "every `async` function generates a different, unique type, so we'll need to wrap the returned the future with `Pin<Box<dyn Future<Output = ()>>>`.
I think `.await` might still be the better option here.

src/update_agent/actor.rs
kelvinfan001: I just moved `Result` handling fully here, but noticed that this made the `self.finalize_deployment` function slightly different from the other similar helpers. In the other helpers, such as `attempt_deploy`,  it does the error logging in the function itself, but the caller _also_ does some additional handling.
For the case of `finalize_deployment` here, there is nothing additional to do for the caller if the result is `Err`. So would it actually make more sense to leave this as is, or perhaps we should move the error logging to the caller function in the other helpers as well?

kelvinfan001: On second thought, yeah I think it would make sense to just move all the error logging to the callers.

src/update_agent/actor.rs
kelvinfan001: I've made it stop returning `Result`s, but I thought this is already an `async` function, or am I missing something...

lucab: You are totally right, sorry for the confusion. I think I was locally reading by mistake a previous revision where `register_as_driver` was still using actix future types.

lucab (approved):

Issue #609, opened 2021-07-27
Split out `state_changed` field from `UpdateAgentInfo`


kelvinfan001:
Originally from discussion at https://github.com/coreos/zincati/pull/577#discussion_r677515055, and paraphrasing some of @lucab's comments from out-of-band chat.
In that PR, we split out the `UpdateAgentState` and put it behind RwLock for more granular locking, and so allowing for the update_agent actor to be able to handle messages concurrently in general.

However, the `state_changed` field that was originally part of `UpdateAgent` (and now part of `UpdateAgentInfo`) is now in an awkward position. Conceptually, all the data in `UpdateAgentInfo` should be read-only, and so it does not need to be behind any locks. But clearly `state_changed` is a read-write field.

One straightforward approach to solve this issue would be to move `state_changed` to be embedded within the state, so it gets locked with it. The downside to this is that the current RwLock that we are using cannot be read while the state is write-locked for updates.

Since this field is just a timestamp, we can actually have a weaker guarantee. For example, it could be fine to read a stale timestamp while the update_agent's FSM is being updated.

PR #614, opened 2021-07-29
update_agent: split `state_changed` out of `UpdateAgentInfo`

kelvinfan001:
We would like for everything under `UpdateAgentInfo` to be read-only.
Conceptually, it would make sense to put the `state_changed` field
behind the same lock as `UpdateAgentState`, but `UpdateAgentState` is
behind a `RwLock`, which doesn't allow for reads when it is acquired for
writing.
Get around this by manually ensuring that we only write to
`state_changed` when the `RwLock` that `UpdateAgentState` is behind is
acquired for writing. This allows us to ensure that the `RefCell` that
`state_changed` is behind will not panic, and also allows for reads even
when the `UpdateAgentState`'s `RwLock` is acquired for writing.

Addresses https://github.com/coreos/zincati/pull/577#discussion_r677072104 and closes https://github.com/coreos/zincati/issues/609.

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -9,6 +9,7 @@ use futures::prelude::*;
 use log::trace;
 use prometheus::{IntCounter, IntCounterVec, IntGauge};
 use std::collections::BTreeSet;
+use std::mem::discriminant;
 use std::rc::Rc;
 use std::time::Duration;
 
@@ -79,6 +80,7 @@ impl Handler<RefreshTick> for UpdateAgent {
         // long enough lifetime.
         let update_agent_info = self.info.clone();
         let lock = Rc::clone(&self.state);
+        let last_changed = Rc::clone(&self.state_changed);
         let state_action = async move {
             // Acquire RwLock to access state.
             let mut agent_state_guard = lock.write().await;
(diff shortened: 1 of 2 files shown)


src/update_agent/actor.rs
```diff
@@ -131,23 +133,29 @@ impl Handler<RefreshTick> for UpdateAgent {
                 UpdateAgentState::EndState => (),
             };
 
+            // Update state_changed timestamp if necessary.
+            if discriminant(&prev_state) != discriminant(&*agent_state_guard) {
+                let cur_timestamp = chrono::Utc::now();
+                // This mutable borrow will not panic because we are still holding
```
lucab, line 139: Self-note: we may need to revisit this statement as soon as we add a consumer for this field, which may or may not result in other consumers having an immutable borrow when we reach here. We'll likely need to replace the `RefCell` usage at that point, but we can cross that bridge once we get there.

lucab (approved):

src/update_agent/actor.rs
kelvinfan001: Ah right, I don't think my comment there is true once there are consumers. I completely forgot that consumers need to borrow the field for reading as well.

kelvinfan001: Ah right, I don't think my comment there is true once there are consumers. I completely forgot that consumers need to borrow the field for reading as well.

PR #619, opened 2021-08-04
update_agent: use `Cell` for `state_changed` field

kelvinfan001:
The previous incorrect usage of `RefCell` allowed for panics at runtime
if there were consumers.
`Cell` should be sufficient here, since the data that we are storing
is a simple timestamp (easily `Copy`-able). We do not need to use an atomic type because the
update_agent actor runs on a single thread on the SystemArbiter.

Ref: https://github.com/coreos/zincati/pull/614#discussion_r679747046

### DIFF
--- a/src/update_agent/actor.rs
+++ b/src/update_agent/actor.rs
@@ -135,10 +135,9 @@ impl Handler<RefreshTick> for UpdateAgent {
 
             // Update state_changed timestamp if necessary.
             if discriminant(&prev_state) != discriminant(&*agent_state_guard) {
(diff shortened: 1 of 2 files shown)


kelvinfan001:
Note to self:
The reason why we do not need to use an atomic type here (even though `update_agent` is concurrent) is because we achieved concurrency via green threads (using Tokio), so we control exactly when and where suspension points occur. As such, we can be certain that modifications of the `state_changed` field won't be interrupted halfway.

lucab (approved):

Issue #558, opened 2021-05-19
Use new --skip-branch-check and verify fedora-coreos.stream commit metadata before unlocking


jlebon:
This is the Zincati part of https://github.com/coreos/fedora-coreos-tracker/issues/749. We need to have it start using the new `--skip-branch-check` when calling `rpm-ostree deploy` and also check in the commit metadata that the stream matches the expected output.

See: https://github.com/coreos/fedora-coreos-tracker/issues/749#issuecomment-834801971
See: https://github.com/coreos/fedora-coreos-tracker/issues/749#issuecomment-839947093

kelvinfan001:
Re: https://github.com/coreos/fedora-coreos-tracker/issues/749#issuecomment-839947093, wouldn't it make more sense to check the commit metadata before trying to stage it (i.e. check it at the check updates stage, that way we avoid fully deploying an invalid release)? 
I think checking metadata makes more sense at the check updates stage, since a release with bad metadata isn't a transient error, i.e. we should never deploy and finalize into an update with bad metadata and it won't fix itself over time; there's no point retrying again later, or ever.
So I think https://github.com/coreos/fedora-coreos-tracker/issues/749#issuecomment-834801971 makes more sense to me. Am I overlooking anything here?

jlebon:
I think the rationale there was that the likelihood this happens was so low that it wasn't worth the added complexity of doing a separate `pull --commit-metadata-only` to check the stream. It's more like a final sanity-check before lifting the lock using information we already have on-hand. Though definitely cool with me if we go the `pull --commit-metadata-only` path too!

kelvinfan001:
Ah okay, that also makes sense. I guess I don't fully know how dangerous it is to deploy a commit potentially from some unknown stream, so it seemed a bit weird to me to check _after_ already creating a deployment and writing everything on disk.

kelvinfan001:
Got it, the sanity check is not for any security purposes (since the the super old node should still have the right keys to verify the nearest barrier release that it's trying to update to). The sanity check is in place to address
```
17:01:14 <jlebon> imagine if we somehow screw up the graph, and the node goes from testing to stable. now, from that point on it's extremely difficult to rectify that situation because the node will now be checking the wrong update graph
17:01:29 <jlebon> it's super unlikely, but super bad
```
which is extremely unlikely.

PR #622, opened 2021-08-09
Use new --skip-branch-check and verify fedora-coreos.stream commit metadata before unlocking

kelvinfan001:
### update_agent: skip branch check when deploying
Use the new `--skip-branch-check` option when deploying new updates with
rpm-ostree to address https://github.com/coreos/zincati/issues/558.
After staging a new update, sanity check that the newly deployed release is on
the same stream as the current booted deployment and abandon and denylist
the update if not.

### tests: add test for checking stream


kelvinfan001:
This is the groundwork for https://github.com/coreos/zincati/issues/558.

kelvinfan001:
Requires https://github.com/coreos/coreos-assembler/pull/2351 to pass CI.

kelvinfan001:
Depends on https://github.com/coreos/zincati/pull/624

src/rpm_ostree/cli_status.rs
```diff
@@ -98,12 +98,23 @@ pub fn parse_booted(status: &StatusJson) -> Result<Release> {
...
-pub fn parse_updates_stream(status: &StatusJson) -> Result<String> {
+pub fn parse_booted_updates_stream(status: &StatusJson) -> Result<String> {
     let json = booted_json(status)?;
     ensure!(!json.base_metadata.stream.is_empty(), "empty stream value");
     Ok(json.base_metadata.stream)
 }
 
+/// Parse updates stream for pending deployment from status object.
+pub fn parse_pending_updates_stream(status: &StatusJson) -> Result<String> {
+    let pending_json = status.deployments[0].clone();
```
lucab, line 109: Does this implicitly expect the first deployment to be the pending one? It does seem a bit fragile (and possibly wrong).

src/update_agent/actor.rs
```diff
@@ -464,6 +463,16 @@ impl UpdateAgentInfo {
         depls
     }
 
+    /// Return whether the update stream of the pending deployment, and cleanup the pending
```
lucab, line 466: I'm failing to parse the "whether" here.

src/rpm_ostree/cli_status.rs
kelvinfan001: Yes, this does expect the first deployment to be the pending one.

src/update_agent/actor.rs
kelvinfan001: ah, that was a typo. There shouldn't be a "whether" there.

src/rpm_ostree/cli_status.rs
lucab: It would be good not to hardcode that assumption, we can pick up the staged/pending deployment through properties.

src/update_agent/actor.rs
```diff
@@ -491,6 +499,52 @@ impl UpdateAgentInfo {
...
+                    log::info!("{}", msg);
+                    state.machine_state.update_staged(release);
+                    // The release we just deployed is on the correct stream so we
+                    // return early.
+                    return;
+                }
+            }
+            Err(e) => {
+                log::error!(
+                    "failed to check pending deployment's update stream: {}, abandoning update {}",
```
lucab, line 538: I think this is a bit too aggressive, on transient errors (e.g. dbus communication failures) we may end up blocking a release through the `denylist` below.

lucab (approved):
I think this is good already to merge. There are two minor corner-cases which could be handled a bit more gracefully, but I'll follow up on those in a separate PR.

lucab:
Followup at https://github.com/coreos/zincati/pull/647.

PR #647, opened 2021-09-20
actor: handle staging failures in a more graceful way

lucab:
This tweaks the staging and introspection logic in order to be more
resilient to failures, and to handle temporary errors in a graceful way.
In particular, the update agent is now more conservative and only
marks a target update as blocked if it really observed a stream mismatch.

Ref: https://github.com/coreos/zincati/pull/622#pullrequestreview-758708050

### DIFF
--- a/src/rpm_ostree/cli_status.rs
+++ b/src/rpm_ostree/cli_status.rs
@@ -104,15 +104,21 @@ pub fn parse_booted_updates_stream(status: &StatusJson) -> Result<String> {
     Ok(json.base_metadata.stream)
 }
 
-/// Parse updates stream for pending deployment from status object.
-pub fn parse_pending_updates_stream(status: &StatusJson) -> Result<String> {
-    let pending_json = status.deployments[0].clone();
-    ensure!(!pending_json.booted, "no pending deployment found");
-    ensure!(
-        !pending_json.base_metadata.stream.is_empty(),
-        "empty stream value"
-    );
-    Ok(pending_json.base_metadata.stream)
+/// Parse pending deployment from status object.
+pub fn parse_pending_deployment(status: &StatusJson) -> Result<Option<(Release, String)>> {
+    // There can be at most one staged/pending rpm-ostree deployment,
+    // thus we only consider the first matching entry (if any).
+    let staged = status.deployments.iter().find(|d| d.staged).cloned();
+
+    match staged {
+        None => Ok(None),
+        Some(json) => {
+            let stream = json.base_metadata.stream.clone();
+            ensure!(!stream.is_empty(), "empty stream value");
+            let release = json.into_release();
+            Ok(Some((release, stream)))
+        }
+    }
 }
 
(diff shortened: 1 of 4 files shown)


src/rpm_ostree/cli_status.rs
```diff
@@ -104,15 +104,30 @@ pub fn parse_booted_updates_stream(status: &StatusJson) -> Result<String> {
...
-        !pending_json.base_metadata.stream.is_empty(),
-        "empty stream value"
-    );
-    Ok(pending_json.base_metadata.stream)
+/// Parse pending deployment from status object.
+pub fn parse_pending_deployment(
+    status: &StatusJson,
+    preferred: &Release,
+) -> Result<(Release, String)> {
+    // Pick a deployment among all staged ones, if any. In case of
```
cgwalters, line 112: There can only be one staged, FWIW.

src/update_agent/actor.rs
```diff
@@ -463,13 +460,49 @@ impl UpdateAgentInfo {
...
+            log::error!(
+                "pending deployment {} expected to be on stream '{}', but found '{}' instead",
+                pending.version,
+                self.identity.stream,
+                stream
+            );
+            Ok(false)
+        } else {
+            log::trace!(
+                "detected valid stream '{}' for pending deployment {}",
```
cgwalters, line 500: "detected" has a slight negative connotation, and is used as such above in an error message.  Maybe:
```suggestion
                "validated stream '{}' for pending deployment {}",
```
?

src/update_agent/actor.rs
```diff
@@ -339,27 +340,23 @@ impl UpdateAgentInfo {
...
-            }
+        if let Err(e) = deploy_outcome {
+            log::error!("failed to stage deployment: {}", e);
+            let fail_count =
+                UpdateAgentInfo::deploy_attempt_failed(&release, &mut state.machine_state);
+            let msg = format!(
+                "trying to stage {} ({} failed deployment attempt{})",
+                release.version,
+                fail_count,
+                if fail_count > 1 { "s" } else { "" }
```
cgwalters, line 356: I know this existed before, but I try to avoid this by doing e.g.:
`trying to stage {} (failed attempts: {})` - it's OK to have "failed attempts: 1" for example.

cgwalters (approved):
Fine as is, just some optional issues for your followup to some optional issues.

src/rpm_ostree/cli_status.rs
lucab: Thanks for highlighting that. I'll note it down in a comment, and simplify the code accordingly.

src/update_agent/actor.rs
lucab: Yes, it'd be good to tweak it here, I'll bundle that in the next rebase.

lucab:
Amended following review suggestions, PTAL.

cgwalters (approved):
Nice, LGTM!

PR #747, opened 2022-03-17
github/release-checklist: Remove Windows binaries from vendored sources

travier:
Workaround to remove Windows specific binaries that we don't need from
the vendored sources.

See: https://github.com/rust-lang/cargo/issues/7058

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-checklist.md
+++ b/.github/ISSUE_TEMPLATE/release-checklist.md
@@ -52,6 +52,7 @@ Push access to the upstream repository is required in order to publish the new t
 
 - assemble vendor archive:
   - [ ] `cargo vendor target/vendor`
+  - [ ] `rm -r target/vendor/winapi*gnu*/lib/*.a`
   - [ ] `tar -czf target/zincati-${RELEASE_VER}-vendor.tar.gz -C target vendor`
 
 - publish the release:


.github/ISSUE_TEMPLATE/release-checklist.md
```diff
@@ -52,6 +52,7 @@ Push access to the upstream repository is required in order to publish the new t
 
 - assemble vendor archive:
   - [ ] `cargo vendor target/vendor`
+  - [ ] `rm -rf vendor/winapi*gnu*/lib/*.a`
```
cgwalters, line 55: I am totally fine with this as is, but...I think we should aim to use a shared wrapper script for this.  I am unsure of a good place.  Perhaps just randomly pick one project and have a shell script that is literally:
```
#!/bin/bash
set -euo pipefail
cargo vendor target/vendor
rm -rf vendor/winapi*gnu*/lib/*.a
```
to start.

(But there's plenty to improve on there, by having it in a shared place even just for coreos we can start doing so)

jlebon: This is related to https://github.com/coreos/fedora-coreos-tracker/issues/1035.

jlebon: This needs to be 

```suggestion
  - [ ] `rm -rf target/vendor/winapi*gnu*/lib/*.a`
```

Did that for coreos-installer at least in https://github.com/coreos/coreos-installer/pull/845.

(And of course, a perfect example why these checklists should be centralized better. :) )

PR #876, opened 2022-10-28
Remove requirement for `coreos-assembler.basearch`

cgwalters:
This is aiming to increase compatibility of zincati with systems that aren't derived from Fedora CoreOS (e.g. IoT, desktop systems, etc.)

Additionally, even in the FCOS case, when one derives a container image from FCOS, the final build isn't generated by coreos-assembler and hence the ostree commit won't have this metadata.

In the end - I can't think of a use case for even trying to have zincati managing an image for a different architecture than the one it's compiled for.

So, copy the code I wrote for oci-spec-rs which maps from the Rust architecture to the GNU one.  xref
https://github.com/containers/oci-spec-rs/commit/52e450e631a212769df4dcd5b7904d942d246d24

### DIFF
--- a/src/identity/mod.rs
+++ b/src/identity/mod.rs
@@ -93,8 +93,7 @@ impl Identity {
     pub fn try_default() -> Result<Self> {
         // Invoke rpm-ostree to get the status of the currently booted deployment.
         let status = rpm_ostree::invoke_cli_status(true)?;
-        let basearch = rpm_ostree::parse_basearch(&status)
-            .context("failed to introspect OS base architecture")?;
+        let basearch = crate::identity::current_architeture().to_string();
         let current_os =
             rpm_ostree::parse_booted(&status).context("failed to introspect booted OS image")?;
         let node_uuid = {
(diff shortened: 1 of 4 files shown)


lucab:
This is effectively a revert of https://github.com/coreos/zincati/pull/58, and I'm fine with merging this.
But if you read through the reference there the idea wasn't to support cross-arch updates, but to make sure the underlying environment is sane (i.e. the underlying basearch and platform ID can be introspected through some stable API).
The fact that the commit metadata get lost and Zincati is the first thing to break is a canary in the mine.

/cc @jlebon

cgwalters:
OK, if you want to look at it we can - here's the neat thing, with this move we can use the *standard container image metadata*.  Specifically I can change things (in the container case) to call https://docs.rs/ostree-ext/latest/ostree_ext/container/store/fn.query_image_ref.html and we can dig from there into https://docs.rs/oci-spec/0.5.8/oci_spec/image/struct.ImageConfiguration.html#method.architecture

(However, we would then need to re-map the Go/OCI architecture back to GNU, we have Go code for that in https://github.com/coreos/stream-metadata-go/blob/main/arch/arch.go that we'd need to put somewhere in Rust)

src/identity/mod.rs
```diff
@@ -190,6 +189,17 @@ fn compute_node_uuid(app_id: &id128::Id128) -> Result<id128::Id128> {
     Ok(id)
 }
 
+// Translate the architecture we were compiled for to the
+// output from the `arch` command that is used for RPM and
+// coreos-assembler builds.
+pub(crate) fn current_architeture() -> &'static str {
```
cgwalters, line 195: https://github.com/coreos/stream-metadata-rust/pull/34

cgwalters:
I've been thinking about this more, and I think 

> make sure the underlying environment is sane (i.e. the underlying basearch and platform ID can be introspected through some stable API).

is covered by https://github.com/coreos/stream-metadata-rust/pull/34

We could *also* instead add rpm-ostree (really libdnf)'s idea of the architecture name into `rpm-ostree status --json` - not as part of the booted deployment.

But, I'm still not sure that the original change was really strongly motivated; we don't have any case for cross-architecture updates, so again the architecture that we built for Rust must be the same as the RPM/GNU/Linux one - after a suitable mapping.  So it's more about maintaining that suitable mapping in a sustainable way, and I do think the stream-metadata-rust PR is that.

lucab (approved):

PR #878, opened 2022-11-02
Initial support for container image updates

cgwalters:
This is part of https://github.com/coreos/fedora-coreos-tracker/issues/1263

If we're booted into a container image, then instead of looking for the special `fedora-coreos.stream` ostree commit metadata, we can do the much more obvious and natural thing of looking at the container image tag.

### DIFF
--- a/src/rpm_ostree/cli_status.rs
+++ b/src/rpm_ostree/cli_status.rs
@@ -11,6 +11,9 @@ use std::collections::BTreeSet;
 use std::fs;
 use std::rc::Rc;
 
+/// The well-known Fedora CoreOS base image.
+const FEDORA_COREOS_CONTAINER: &str = "quay.io/fedora/fedora-coreos";
+
 /// Path to local OSTree deployments. We use its mtime to check for modifications (e.g. new deployments)
 /// to local deployments that might warrant querying `rpm-ostree status` again to update our knowledge
 /// of the current state of deployments.
@@ -48,6 +51,7 @@ pub struct StatusJson {
 #[serde(rename_all = "kebab-case")]
 pub struct DeploymentJson {
     booted: bool,
+    container_image_reference: Option<String>,
     base_checksum: Option<String>,
     #[serde(rename = "base-commit-meta")]
     base_metadata: BaseCommitMetaJson,
@@ -62,7 +66,7 @@ pub struct DeploymentJson {
 #[derive(Clone, Debug, Deserialize)]
 struct BaseCommitMetaJson {
     #[serde(rename = "fedora-coreos.stream")]
-    stream: String,
+    stream: Option<String>,
 }
 
 impl DeploymentJson {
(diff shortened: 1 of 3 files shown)


cgwalters:
This requires
https://github.com/coreos/rpm-ostree/pull/4123
And it also needs us to add version tags to the container image, e.g. we should make `quay.io/fedora/fedora-coreos:37.20221021.1.0` real.
And we also need support in rpm-ostree for `rpm-ostree deploy tag=` to work when tracking a container image reference.

cgwalters:
(It's kind of heavyweight here to add the `ostree-ext` dependency *solely* to parse the container image reference string.  A bit tempting to explicitly separate out the parts from that in the status JSON...or I guess we could split out a tiny crate for that)

cgwalters:
Or, of course...ultimately, perhaps zincati's functionality should be folded into rpm-ostree.

cgwalters:
Split the prep patches to https://github.com/coreos/zincati/pull/879

cgwalters:
OK rebased :surfer: and lifting draft - I think we can ship this.

cgwalters:
What will happen here now is zincati will switch the node back from container → ostree, which is obviously not what we want.   Fixing that however requires a lot more work - we need to plumb through corresponding container image digests into the cincinnati metadata *or*/in-addition we add tags to the registry corresponding to the desired version.

But...I think this is arguably better than having zincati just fail today.

lucab:
> we can do the much more obvious and natural thing of looking at the container image tag

I think this may scratch the immediate itch but it's likely not a good foundation to build upon.

Overall we have been quite successful uncoupling each single point releases from the moving tip of the stream.
I'm unsure why we aren't publishing dotted tags right now (e.g. `quay.io/fedora/fedora-coreos:36.20221030.3.0`), but I can see FCOS going in that direction in the future which will clash with this.

Also, the main usecase for image containerization is that people will end up pushing their own images, in which case we can't really rely on tags nor match our upstream refspec.

Maybe I'm misunderstanding something, but I think even right the ostree commit we ship in the container on quay does carry the `fedora-coreos.stream` metadata key. Is it being lost on the host after rebase, and you are trying to fix that here? Or are you trying to tackle the case of derived customized images which may not carry the metadata?

jlebon:
Would be good to flesh out the design for https://github.com/coreos/fedora-coreos-tracker/issues/1263 before doing implementation changes. I'm also hesitant to bind us to certain conventions before tackling the larger issue of "layering + automatic updates".

jlebon:
> Fixing that however requires a lot more work - we need to plumb through corresponding container image digests into the cincinnati metadata or/in-addition we add tags to the registry corresponding to the desired version.

Right, I think that'd be a good avenue. The graph schema is generic over the payload type and has a `org.fedoraproject.coreos.scheme` for this purpose. It's currently set to `checksum`, but in a container world, it could be set to e.g. `sha256digest` where the payload is then the container digest to feed to rpm-ostree. ISTM like the existing interface of `rpm-ostree deploy revision=$X` could be extended to support container digests to fit into this model.

That way Zincati doesn't really need to know in details about container things, which AFAIK matches the current status for OSTree things.

But again, the "layering" aspect of this changes things and how all this intersects with that use case needs some careful thought.

jlebon:
Though one thing we could do I guess is to gate this PR behind some sufficiently scary flag/env var?

But as an aside, I do think the `fedora-coreos.stream` metadata is important enough that it should be kept both in the layered commit and probably also promoted to an image label, even if we don't end up using it for Zincati's purposes.

cgwalters:
> But as an aside, I do think the fedora-coreos.stream metadata is important enough that it should be kept both in the layered commit and probably also promoted to an image label, even if we don't end up using it for Zincati's purposes.

Well...part of the idea of all this is to get away from "ostree stuff".  We could try to set up a system where metadata on the original commit does propagate to the local merge commit, but...in the end I think what we want is to just have it in the container image metadata, so: https://github.com/coreos/coreos-assembler/pull/3214

Then zincati can parse the container metadata using standard APIs.

All of this helps lead to a world where a zincati-like tool can also be used for application container (e.g. podman) updates.  And it should help us align with how e.g. OCP is already using Cincinnati to target a container image (not an ostree commit).

kaplan-michael:
Hello,  Is this already "usable" in some way(custom build rpm-ostree, zinacati, etc.) 
I don't mind if it's still unstable, mostly want to try it out early.

Thanks a lot

jbtrystram:
Closing this as we got this functionality in https://github.com/coreos/zincati/pull/1241

PR #881, opened 2022-11-07
Start a port to once_cell

cgwalters:
It's what's likely to be in `std` eventually.  Just a drive-by change.


lucab (approved):
Definitely a good idea, LGTM!

PR #896, opened 2022-11-19
service: exit and prevent restarts if booted into a container image

cgwalters:
This is part of https://github.com/coreos/fedora-coreos-tracker/issues/1263

We don't yet have an official stance on how zincati and custom container images interact.  Today, zincati just crash loops. This changes things so that we gracefully exit if we detect the booted system is using a container image origin.

(The code here isn't quite as clean as it could be; calling
 `std::process::exit()` in the middle of the call chain isn't
  elegant but doing better would require plumbing through an
  `Option<T>` through many layers)

### DIFF
--- a/src/rpm_ostree/cli_status.rs
+++ b/src/rpm_ostree/cli_status.rs
@@ -37,6 +37,18 @@ lazy_static::lazy_static! {
     )).unwrap();
 }
 
+/// An error which should not result in a retry/restart.
+#[derive(Debug, Clone)]
+pub struct FatalError(String);
+
+impl std::fmt::Display for FatalError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl std::error::Error for FatalError {}
+
 /// JSON output from `rpm-ostree status --json`
 #[derive(Clone, Debug, Deserialize)]
 pub struct StatusJson {
@@ -48,6 +60,7 @@ pub struct StatusJson {
 #[serde(rename_all = "kebab-case")]
 pub struct DeploymentJson {
     booted: bool,
+    container_image_reference: Option<String>,
     base_checksum: Option<String>,
     #[serde(rename = "base-commit-meta")]
     base_metadata: BaseCommitMetaJson,
@@ -62,7 +75,7 @@ pub struct DeploymentJson {
 #[derive(Clone, Debug, Deserialize)]
 struct BaseCommitMetaJson {
     #[serde(rename = "fedora-coreos.stream")]
-    stream: String,
+    stream: Option<String>,
 }
 
 impl DeploymentJson {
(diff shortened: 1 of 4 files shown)


lucab:
Thanks for assembling this patch, I like this direction. It sounds like we are now generally trying to address the case of Zincati somehow running on unexpected OS/state, which is a good idea from the point of view of UX.

One of the goals here seem to be avoiding an infinite restart loop. I think systemd has a dedicated [`RestartPreventExitStatus=`](https://www.freedesktop.org/software/systemd/man/systemd.service.html#RestartPreventExitStatus=) for that exact purpose, but I've personally never used it.
If it allows us to both prevent the restart-loop but also record a failed state and a non-zero exit code, I think it is worth to switch to that.

Aside, I personally think it would be a good time to try to plumb an "error object with optional exit code" till here:
https://github.com/coreos/zincati/blob/ebb241c631b822cec4ebe617bc12a1337dec461b/src/main.rs#L58-L61

If you don't want to waste too much time on that, I can give it a try using `anyhow` primitives/downcasting, then add a commit on top here for that.

cgwalters:
So, automatic updates on by default are core to the CoreOS philosophy.  Is your rationale behind having zincati explicitly fail (and hence that having that failure at least minimally appear via CLHM and other automation tools) to represent this state in a very visible way?

It might be of course that admins have set up something to upgrade the system not via zincati; it's easy to have an external orchestration system effectively just run `rpm-ostree upgrade` targeting a tag or even an explicit digest.

In that scenario, the user can mask zincati and hence mask the spurious failure.

On the other hand, what I was thinking with this patch is that their choice to use an image also kind of signals such an intent.

cgwalters:
OK, I reworked this a bit to use a custom error type and still exit with an error.

cgwalters:
> On the other hand, what I was thinking with this patch is that their choice to use an image also kind of signals such an intent.

lucab:
I think so. Overall I don't think that "booted from a containerized commit" and "user wants to opt-out auto-update" are two things that we should conflate at this time. Instead I'd scope this PR to "Zincati inspected this booted commit and doesn't know how to auto-update it".

If Zincati manages to start it can surface different kind of failures (both transient and permanent) through metrics to the monitoring infra. But if it fails to complete startup for whatever reason, it needs some other kind of system to expose this failure state.

Auto-updates are critical enough that "keep restarting" is often a good solution to eventually bypass any transient error. But in this case we know we can't get out of this situation without manual intervention, so I think it's fair to stay in a perma-failed state.

src/rpm_ostree/cli_status.rs
```diff
@@ -85,12 +98,21 @@ impl DeploymentJson {
 
 /// Parse the booted deployment from status object.
 pub fn parse_booted(status: &StatusJson) -> Result<Release> {
-    let json = booted_json(status)?;
-    Ok(json.into_release())
+    let status = booted_json(status)?;
+    if let Some(img) = status.container_image_reference.as_ref() {
+        let msg = format!("Automatic updates disabled; booted into container image {img}");
+        crate::utils::update_unit_status(&msg);
+        return Err(anyhow::Error::new(FatalError(msg)));
```
lucab, line 105: I was more expecting this additional step to be performed in the error case in `fedora_coreos_stream_from_deployment()` (so that both cases would prevent restarts), but I guess just here it could work as well.

lucab (approved):

cgwalters:
> Overall I don't think that "booted from a containerized commit" and "user wants to opt-out auto-updates" are two things that we should conflate at this time. Instead I'd scope this PR to "Zincati inspected this booted commit and doesn't know how to auto-update it".

Agree, this PR is just having a better error in the latter case.

PR #1059, opened 2023-08-15
update_agent: handle new output format for loginctl

dustymabe:
In systemd 254 the `loginctl list-sessions --json` output was modified slightly:

Before
```
[{"session":"4","uid":1000,"user":"core","seat":"","tty":""}]
```

After
```
[{"session":"17","uid":1000,"user":"core","seat":"","tty":"n/a","state":"active","idle":false,"since":null}]
```

Notice the change from `""` to `"n/a"` for the `tty` field.

This change probably happened in one of:

- https://github.com/systemd/systemd/pull/27769
- https://github.com/systemd/systemd/pull/27740
- https://github.com/systemd/systemd/pull/27606

Let's update the code to handle the new case.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1547


prestist (approved):
LGTM! great job.

src/update_agent/mod.rs
```diff
@@ -79,14 +79,16 @@ pub struct InteractiveSession {
...
 }
 
-/// Function to deserialize field to `Option<String>`, where empty strings are
-/// deserialized into `None`.
-fn empty_string_as_none<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
+/// Function to deserialize field to `Option<String>`, where empty strings or
+/// `n/a` (not available) strings are deserialized into `None`. In systemd v254+
+/// loginctl list-sessions --json started outputting `n/a` instead of an empty
+/// string for tty.
+fn empty_or_na_string_as_none<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
```
cgwalters, line 86: I'd call this "deserialize_systemd_tty_canonicalized" perhaps.  It's only usable for the systemd tty I suspect; I can't imagine many other things use "n/a".  (Or perhaps I'm wrong and systemd started emitting n/a for other things...)

src/update_agent/mod.rs
```diff
@@ -79,14 +79,16 @@ pub struct InteractiveSession {
     tty_dev: String,
 }
 
-/// Function to deserialize field to `Option<String>`, where empty strings are
-/// deserialized into `None`.
-fn empty_string_as_none<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
+/// Function to deserialize field to `Option<String>`, where empty strings or
+/// `n/a` (not available) strings are deserialized into `None`. In systemd v254+
```
cgwalters, line 83: I think the "a" here usually expands to applicable, not available.

src/update_agent/mod.rs
```diff
@@ -79,14 +79,16 @@ pub struct InteractiveSession {
...
+/// loginctl list-sessions --json started outputting `n/a` instead of an empty
+/// string for tty.
+fn empty_or_na_string_as_none<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
 where
     D: Deserializer<'de>,
 {
     let s = String::deserialize(deserializer)?;
-    if s.is_empty() {
+    if s.is_empty() || s.eq(&String::from("n/a")) {
```
cgwalters, line 91:
```suggestion
    if s.is_empty() || s == "n/a" {
```

cgwalters: (Not only is this shorter, `String::from` is doing an avoidable heap allocation)

src/update_agent/mod.rs
dustymabe: updated!

src/update_agent/mod.rs
dustymabe: I don't know why I came up with available. Of course, it's applicable.

src/update_agent/mod.rs
dustymabe: I've been exposed! I need to learn more rust!

dustymabe: Fixed!

cgwalters (approved):

PR #1088, opened 2023-09-07
README.md: Note likely deprecation

cgwalters:
Since this project isn't container-native.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,15 @@ Zincati is an auto-update agent for Fedora CoreOS hosts.
 
 It works as a client for [Cincinnati] and [rpm-ostree], taking care of automatically updating/rebooting machines.
 
+## Note: future directions
+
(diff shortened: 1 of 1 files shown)


dustymabe:
I'd rather make an update like this when we migrate Fedora CoreOS to something else.

On the other hand this is just a comment and I don't have the brainpower to resist it.

cgwalters:
> I'd rather make an update like this when we migrate Fedora CoreOS to something else.

OK.  I will be working on fleshing out
https://gitlab.com/cgwalters-playground/bootc-demo-base-images/
And building it in Fedora; this will be clearly container-native from the start, and will definitely not include zincati.  Then my thoughts are that we add this as a git submodule and use it in fedora-coreos-config (and for that matter the ostree desktops too), so things like "Fedora iptables hacks" end up there and still shared/reused by FCOS which we will continue to maintain.

Issue #245, opened 2020-03-10
updates: new strategy based on local filesystem


lucab:
One interesting idea that came out of https://github.com/coreos/zincati/issues/204#issuecomment-577212372 is:

> I'd still try to come up with a flow which does not completely bypass Zincati finalization (for example, giving permission to reboot only if a specific filepath exists) and which does not require SSHing to each node.

The idea is to have some kind of logic on each node to touch a file when finalization is allowed, and remove it when it is not allowed. 
The controller can be a containerized agent, or some central task manager able to manipulate files on machines, or even a human via SSH (not recommended).

We won't provide the file-creation, only the updates-strategy in Zincati. Strategy name still to be decided.

cgwalters:
This case feels like it overlaps at least somewhat with systemd inhibitors - any process that doesn't want the system to reboot can use those today.

lucab:
Self-note: removing a file to signal the end of the allowed finalization window is a critical step which may not be feasible at all times (e.g. because network/SSH/whatever is temporarily down). For this reason, there should be a way to encode an optional "not-after" timestamp so that windows can safely auto-expire.

kelvinfan001:
### Rationale

After some discussion with @lucab, we agreed that there is demand for having more manual control over Zincati updates. This could be achieved by giving users more fine-grained control over when update finalizations (reboots) are performed, and would thus be a compromise between full manual control over updates (https://github.com/coreos/zincati/issues/498) and fully automatic updates. Giving users the ability to more directly control when finalizations (reboots) are allowed should ideally divert most of the demand for fully manual updates (in most scenarios, if there is an update available, it should already be staged pretty quickly by Zincati).

A low level approach, such as checking for a file on the filesystem, should be flexible enough to address the need for manually controlling reboot windows. Advantages are that files can always be written even when Zincati is not running, and generally easier to manage from different environments since the only thing required is access to the specific filesystem directory (e.g. from a container bind-mount, scp, etc.) and is easily scriptable.

### Proposal (WIP)

#### Mechanism
- During a regular finalization attempts (`tick_finalize_update()`), Zincati will check the specified location and try to read the file at that location. If the file exists and the contents of the file matches the specified content (contains at least information about the expiry datetime of the file), then the finalization is allowed and a reboot will follow.
[...]

PR #1103, opened 2023-10-09
strategy: add new `marker_file` strategy

jokajak:
Re-opening of https://github.com/coreos/zincati/pull/540 since the original fork was removed.

Original summary:

Add a simple, flexible strategy that checks /var/lib/zincati for
a JSON file (optionally containing an expiry timestamp key) named
allowfinalize.json and allows update finalization if such file
exists and has the proper permissions.

See https://github.com/coreos/zincati/issues/245#issuecomment-823705074 for the rationale for this new strategy.


bitfisher:
Any news here?

This strategy would be very useful in combination with kured!
https://github.com/kubereboot/kured/issues/216

PR #1119, opened 2023-11-08
cargo: zincati release 0.0.27

jlebon:
Changes:
 * cargo: update all dependencies to latest versions
 * cargo: bump libsystemd to 0.7.0 to fix fd leak

Issue #498, opened 2021-03-10
RFE: Allow manual update checks and reboots


kelvinfan001:
# Feature Request #

## Desired Feature ##
Allow for manual, safe, Zincati-driven update checks and reboots.
Currently, in rpm-ostree (v2021.2 and later), if a user tries to `rpm-ostree upgrade` when an update driver (e.g. Zincati) "owns" updates on that machine, rpm-ostree will correctly refuse by default and instruct the user to refer to the updates driver's (Zincati's) documentation, implying that the user should perform an upgrade via the updates driver, instead. However, there is actually no convenient/user-friendly way to perform an upgrade immediately through Zincati, either. 
Stemming from the conversation in https://github.com/coreos/rpm-ostree/pull/2566#issuecomment-778394141, it would be nice if an admin could manually get Zincati to immediately check for an update and possibly reboot into it.
A possible use case for this feature would be when admins know that there is a new update available that contains a bug fix/feature and they want it immediately, but Zincati has not automatically updated into that release due to not checking for updates frequently enough, reboot strategy restraints, or phased rollouts (and wariness). Current ways admins could get around this:
1. Reconfigure Zincati to e.g. have lower wariness, change to `immediate` reboot strategy. Then restart Zincati.
2. Use the more direct `rpm-ostree upgrade --bypass-driver`.

`1.` does not seem very user-friendly, and `2.` is potentially unsafe as rpm-ostree has no knowledge of update graphs, barrier releases, reboot scheduling windows, etc.

### `check-update` ###
Include a command for telling Zincati to check for updates immediately. This should probably temporarily set Zincati's rollout wariness to `0.0` in order to hint Cincinnati to respond with the latest possible release.
```
$ zincatictl check-update
No new updates.

$ zincatictl check-update
Release ... found and deployed. Use `zincatictl finalize` to unlock staged deployment and reboot into it.
```

### `finalize-update` ###
Also include a `finalize-update` command to override the reboot strategy, unlock the staged deployment, and reboot immediately.

If strategy allows reboot now, machine will reboot:
```
$ zincatictl finalize-update
```
If strategy does not allow reboot:
```
$ zincatictl finalize-update
Update strategy does not allow for reboot. Use `--force` to force an update finalization.
```
Force a reboot, overriding the reboot strategy:
```
$ zincatictl finalize-update --force
```

[...]

PR #1121, opened 2023-11-08
zincati.service: periodically restart zincati daemon

dustymabe:
We've seen some issues in the past where a simple restart of the zincati daemon would have allowed systems to continue updating. Let's periodically restart the zincati daemon to handle cases like this in the future, which we can't always foresee.

The most recent example being: https://github.com/coreos/fedora-coreos-tracker/issues/1608

### DIFF
--- a/dist/systemd/system/zincati.service
+++ b/dist/systemd/system/zincati.service
@@ -22,6 +22,7 @@ Type=notify
 ExecStart=/usr/libexec/zincati agent ${ZINCATI_VERBOSITY}
 Restart=on-failure
 RestartSec=10s
+RuntimeMaxSec=3w
 
 [Install]
 WantedBy=multi-user.target


jlebon (approved):
Can you sanity-check that this meshes as expected with `Restart=on-failure`? I.e. that the service correctly gets restarted.

dist/systemd/system/zincati.service
```diff
@@ -22,6 +22,7 @@ Type=notify
 ExecStart=/usr/libexec/zincati agent ${ZINCATI_VERBOSITY}
 Restart=on-failure
 RestartSec=10s
+RuntimeMaxSec=3w
```
cgwalters, line 25: This seems to implicitly rely on the default FCOS release schedule meaning we'll have restarted before we hit this timeout by default.  Now, zincati is pretty FCOS specific but still, this seems...slightly unclean at best.

I think the most obvious bigger fix here is to not run as a persistent daemon at all, but to run as a timer.

cgwalters: (That said, I'm OK with this...but it does seem probably better done via a drop-in in fedora-coreos-config instead of hardcoded in source here)

jlebon: > This seems to implicitly rely on the default FCOS release schedule meaning we'll have restarted before we hit this timeout by default. Now, zincati is pretty FCOS specific but still, this seems...slightly unclean at best.

The chosen period is FCOS-informed, but isn't tightly coupled at all. Whether for a given environment updates come in faster or slower than that, it _shouldn't_ hurt to restart the service. I suggested making it longer so that at least in the context of FCOS (by far the most prevalent, if not only, use of Zincati), it's more of a last resort than a primary line of defense. But yes, also fine with having it be a dropin override in the FCOS overlays.

> I think the most obvious bigger fix here is to not run as a persistent daemon at all, but to run as a timer.

Yes, @travier suggested this as well, but it requires a larger rework.

dustymabe: I had [originally suggested 2w](https://github.com/coreos/fedora-coreos-tracker/issues/1608#issuecomment-1802038890) but @jlebon suggested 3w. TBH it could be 1 day, though I'm not sure how well that would work with our periodic update strategy that people can set.

I'd be OK setting this in FCOS configs if we think it would be better placed there. Theoretically if zincati had other consumers this would be useful to them too, though those users might prefer a different cadence for restart.

dustymabe:
> Can you sanity-check that this meshes as expected with `Restart=on-failure`? I.e. that the service correctly gets restarted.

according to the docs it should, but I can run some local tests.

dist/systemd/system/zincati.service
cgwalters: > it shouldn't hurt to restart the service.

Well, it will be racy with DBus today.

But my larger (yet still minor) concern is this restart will appear as an error, when it's not.

jlebon: > > it shouldn't hurt to restart the service.
> 
> Well, it will be racy with DBus today.

That's fair. I actually was unaware that Zincati _had_ a D-Bus interface until debugging https://github.com/coreos/zincati/pull/1119#issuecomment-1802343646 today. It still seems pretty hidden currently at least, but indeed if we were to develop it more down the line as part of https://github.com/coreos/zincati/issues/498, having it be seemingly randomly force restarted would be odd and we'd probably want to rework/drop this directive.

jmarrero:
closing per https://github.com/coreos/zincati/pull/1121#discussion_r1387261378 let's re-open and discuss further if this is still needed.

PR #1172, opened 2024-04-08
packit: add initial support

HuijingHei:


.packit.yaml
```diff
@@ -0,0 +1,33 @@
...
+files_to_sync:
+    - rust-zincati.spec
+    - .packit.yaml
+
+jobs:
+
+- job: propose_downstream
+  trigger: release
+  dist_git_branches:
+    - fedora-all
```
jlebon, line 23: Minor: I think this'll make it open against e.g. f38 too, which is totally unnecessary. Is there a way to tell Packit "development releases + only the newest stable release"?

jlebon (approved):

jlebon:
Thanks for doing that!

.packit.yaml
HuijingHei: Thanks for the review, refer to https://packit.dev/docs/configuration#aliases, `fedora-development` includes `fedora-40` and `fedora-rawhide`, `fedora-latest-stable` includes `fedora-39`, should exclude f38.

Also exclude `f38` for jobs `koji_build` and `bodhi_update`.

PR #1175, opened 2024-04-17
cargo: update dependencies

HuijingHei:


HuijingHei:
CI failed 
- error: package `clap_builder v4.5.2` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.71.0
- error: package `clap_derive v4.5.4` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.71.0

Should I update `rust-version = "1.71.0"` to `rust-version = "1.74.0"` in Cargo.toml?

jlebon:
I think we can bump to 1.75.0. See https://github.com/coreos/coreos-installer/pull/1453.

HuijingHei:
The failed log seems wired, run successfully locally, will look more about it.

jlebon:
Possibly related to the zbus bump from 3.6 to 3.15. Odd that you can't reproduce it locally though. Are you sure that's with the same `Cargo.lock`?

HuijingHei:
> Possibly related to the zbus bump from 3.6 to 3.15. Odd that you can't reproduce it locally though. Are you sure that's with the same `Cargo.lock`?

I rebuild with override/rootfs, can reproduce the failed issue locally, thanks!
```
# busctl call org.coreos.zincati /org/coreos/zincati org.coreos.zincati.Experimental LastRefreshTime
Call failed: The name is not activatable
```

HuijingHei:
Thanks @jlebon for the pointer, see https://github.com/coreos/zincati/pull/1119#issuecomment-1802558737 and https://github.com/coreos/zincati/pull/966, will skip `zbus` update
```
qemu0 kola-runext-test-experimental.sh[2136]: # Call failed: The name is not activatable

qemu0 kola-runext-test-experimental.sh[2129]: + fatal 'File '\''err.txt'\'' doesn'\''t match regexp '\''Access denied'\'''

qemu0 kola-runext-test-experimental.sh[2129]: + echo error: File ''\''err.txt'\''' 'doesn'\''t' match regexp ''\''Access' 'denied'\'''

qemu0 kola-runext-test-experimental.sh[2129]: error: File 'err.txt' doesn't match regexp 'Access denied'

qemu0 kola-runext-test-experimental.sh[2129]: + exit 1
```

jlebon (approved):
LGTM. Though we should try to get to the bottom of that zbus issue at some point. Would be nice if we could stop bundling in Fedora, and I think this will likely be an issue.

c4rt0 (approved):
/lgtm

PR #1178, opened 2024-04-26
update_agent: add hidden knob to override interactive session check

jlebon:
An issue as old as ~time~Fedora CoreOS is that when you're testing something related to updates, you want Zincati to reboot but because `cosa run` gives you an interactive session and you can't *not* have a session without losing the VM, you're stuck waiting for the 10 minutes grace period to expire.

Just add a hidden knob for this to make that scenario less painful.

The API is pretty much: when you're ready to have Zincati reboot, just `touch /run/zincati/override-interactive-check`.

### DIFF
--- a/src/update_agent/mod.rs
+++ b/src/update_agent/mod.rs
@@ -16,6 +16,7 @@ use serde::{Deserialize, Deserializer};
 use std::cell::Cell;
 use std::collections::BTreeSet;
 use std::fs;
+use std::path::Path;
 use std::rc::Rc;
 use std::time::Duration;
 use tokio::sync::RwLock;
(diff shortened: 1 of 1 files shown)


dustymabe:
I think this touches on a paper cut I thought would be handled by https://github.com/coreos/zincati/issues/498 when we finally implemented that.

As is it improves the state, but I think users should have a documented way to do this too.

dustymabe (approved):
LGTM

dustymabe:
> I think this touches on a paper cut I thought would be handled by #498 when we finally implemented that.

for example, #498 could handle the "I'm in an interactive session, an update is available, and I want to ignore update strategy (i.e. if your node isn't configured to update until the weekend)" case.

src/update_agent/mod.rs
```diff
@@ -37,6 +38,10 @@ const DEFAULT_POSTPONEMENT_TIME_SECS: u64 = 60; // 1 minute.
 /// before abandoning a target update.
 const MAX_DEPLOY_ATTEMPTS: u8 = 12;
 
+/// This is undocumented; it's for testing Zincati in e.g. cosa where you have
+/// an interactive session but you don't want to wait for the full timeout.
+const INTERACTIVE_SESSION_OVERRIDE: &str = "/run/zincati/override-interactive-check";
```
cgwalters, line 43: Seems OK as is, but I personally would have probably leaned towards making this an environment variable and told people to `systemctl edit --runtime zincati` instead.

cgwalters (approved):

src/update_agent/mod.rs
jlebon: Yeah, I debated on this a bit. It's just... so much easier to touch a file than to type out a dropin. Though we could also have added a `cosa run --add-ignition zincati-override-interactive-check` to help.

jlebon:
> > I think this touches on a paper cut I thought would be handled by #498 when we finally implemented that.

Yeah, thanks for calling out #498. I agree it'd mostly solve that case, though OTOH it's also a lot more work. I found this useful for debugging https://github.com/coreos/fedora-coreos-tracker/issues/1691 and it took a few minutes to type it out. We can nuke it when #498 happens.

Issue #1176, opened 2024-04-19
`zbus` bump 3.6.1 to newer version makes busctl failed with `Call failed: The name is not activatable`


HuijingHei:
When update `zbus` and related dependencies from 3.6.1 to newer version, test `kola-runext-test-experimental.sh` failed with error:
```
qemu0 kola-runext-test-experimental.sh[2136]: # Call failed: The name is not activatable

qemu0 kola-runext-test-experimental.sh[2129]: + fatal 'File '\''err.txt'\'' doesn'\''t match regexp '\''Access denied'\'''

qemu0 kola-runext-test-experimental.sh[2129]: + echo error: File ''\''err.txt'\''' 'doesn'\''t' match regexp ''\''Access' 'denied'\'''

qemu0 kola-runext-test-experimental.sh[2129]: error: File 'err.txt' doesn't match regexp 'Access denied'

qemu0 kola-runext-test-experimental.sh[2129]: + exit 1
```

Refer to https://github.com/coreos/zincati/pull/966#issuecomment-1555254514, https://github.com/coreos/zincati/pull/1119#issuecomment-1802558737, and https://github.com/coreos/zincati/pull/1175#issuecomment-2065897666.

jlebon:
I think the actual regression comes from a bump of zvariant and not zbus. Specifically bumping from 3.13.0 to 3.14.0 causes the regression. But it probably comes from how zbus is using zvariant. It prevents us from bumping zbus since it wants to pull in newer zvariant.

Diff: https://github.com/dbus2/zbus/compare/zvariant-3.13.0...zvariant-3.14.0

PR #1241, opened 2024-12-04
update_agent: support rebase to OCI pullspec

jbtrystram:
Add a configuration knob in `update` to optionnaly use OCI pullspec instead of ostree checksums.

This will default to false.
Requires https://github.com/coreos/fedora-coreos-cincinnati/pull/99 and https://github.com/coreos/rpm-ostree/pull/5120


travier (commented):
Very quick look but looks good 👍🏻

src/rpm_ostree/cli_deploy.rs
```diff
@@ -94,16 +94,22 @@ fn invoke_cli_register() -> Result<()> {
...
     let mut cmd = std::process::Command::new("rpm-ostree");
-    cmd.arg("deploy")
-        .arg("--lock-finalization")
-        .arg("--skip-branch-check")
-        .arg(format!("revision={}", release.checksum))
-        .env("RPMOSTREE_CLIENT_ID", "zincati");
+    if oci {
+        // TODO use --custom-origin-url and --custom-origin-description
+        cmd.arg("rebase")
+            .arg(format!("ostree-unverified-registry:{}", release.checksum));
```
cgwalters, line 105: Hmm, we'll need the finalization locking here too.

I guess one thing also related to this is using bootc instead of rpm-ostree here, but not a blocker.

tests/fixtures/00-config-sample.toml
```diff
@@ -13,6 +13,7 @@ base_url = "http://cincinnati.example.com:80/"
 allow_downgrade = true
 enabled = false
 strategy = "fleet_lock"
+use_oci = false
```
dustymabe, line 16: I think in our discussions we came to the conclusion that we didn't need this knob. i.e. inspecting the system state to see it was deployed via container or not would suffice and then we'd change behavior based on that.

jbtrystram: So you would have to manually rebase, then zincati follow ? 
I thought you'd use the knob to migrate

src/rpm_ostree/cli_deploy.rs
travier: We can not switch to bootc here until we have a solution for local layering.

src/rpm_ostree/cli_status.rs
```diff
@@ -85,37 +85,62 @@ impl Deployment {
...
+            .parse()
+            .map_err(|_| anyhow!("invalid Coreos version ID"))?;
+
+        match stream_id {
+            1 => Ok("next".to_string()),
+            2 => Ok("testing".to_string()),
+            3 => Ok("stable".to_string()),
+            _ => Err(anyhow!("This stream is not supported by Zincati")),
+        }
+    }
```
jbtrystram, line 143: This is a workaround for https://github.com/coreos/rpm-ostree/issues/5196
I tried to fix the parsing to properly handle the nested JSON but this proved difficult, so i went this route. 
This is a stopgap solution until we use `custom-origin-url` to fetch the update graph from an OCI artifact wanyway.

src/rpm_ostree/cli_deploy.rs
```diff
@@ -99,11 +99,21 @@ fn invoke_cli_deploy(release: Release, allow_downgrade: bool) -> Result<Release>
...
-    cmd.arg("deploy")
-        .arg("--lock-finalization")
-        .arg("--skip-branch-check")
-        .arg(format!("revision={}", release.checksum))
-        .env("RPMOSTREE_CLIENT_ID", "zincati");
+    if release.is_oci {
+        // TODO use --custom-origin-url and --custom-origin-description
+        cmd.arg("rebase")
+            .arg(format!(
+                "ostree-remote-image:fedora:docker://{}",
```
jbtrystram, line 106: Here I hard-coded the fedora ostree remote reference to get the ostree signature verification. 
In the future, we will sign the OCI image directly and switch to bootc, making this irrelevant. 
I suppose this will break Zincati for OCI images not created by coreos. I need to test with a layered build to see if the original ostree signature is still verified or if it breaks.

src/rpm_ostree/cli_status.rs
```diff
@@ -85,37 +85,62 @@ impl Deployment {
...
-        self.base_checksum
+        self.container_image_reference
             .clone()
+            .map(|pullspec| {
+                pullspec
+                    .strip_prefix("ostree-remote-image:fedora:docker://")
+                    .map(|s| s.to_string())
+            })
+            .flatten()
+            .or(self.base_checksum.clone())
```
jbtrystram, line 102: Maybe it would make more sense to add an `Option<String>` to the `Deployement` struct to be able to have both the pullspec and the base commit ? 
However the cincinnati graph does not contains checksum in the OCI case so we can't compare ostree commits when doing `rebase` later anyway.

jbtrystram (commented):
I updated this to use `ostree-remote-image` so we validate the ostree signatures, as suggested in https://github.com/coreos/fedora-coreos-tracker/issues/1823#issuecomment-2583058403

I think this is ready for review now. I will clean up and sqash commits once the design is agreed

jbtrystram:
OK, had some more discussion with Jlebon and travier today and there are some more things to flesh out

src/rpm_ostree/cli_status.rs
```diff
@@ -85,37 +86,58 @@ impl Deployment {
...
+    }
+}
+
+fn fedora_coreos_stream_from_custom_origin(origin: &String) -> Result<String> {
+    //ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos:testing
+    let origin_split: Vec<&str> = origin.split(':').collect();
+    if origin_split.len() != 5 {
+        Err(anyhow!("{origin} is not a valid custom URL"))
+    } else {
+        Ok(origin_split[4].to_string())
```
jbtrystram, line 139: Not sure if Zincati should support other ostree image references ? 

Other URLs schemes are supported by rpm-ostree : https://coreos.github.io/rpm-ostree/container/#url-format-for-ostree-native-containers

src/rpm_ostree/mod.rs
```diff
@@ -89,9 +92,20 @@ impl Release {
             version: node.version,
             checksum: node.payload,
             age_index: Some(age),
+            is_oci: { scheme == OCI_SCHEME },
         };
         Ok(rel)
     }
+    pub fn get_pullspec_hash(&self) -> Result<String> {
+        let split: Vec<&str> = self.checksum.split(":registry:").collect();
```
jbtrystram, line 100: This works but requires the use of that specific rpm-ostree URL format, see above comment

jbtrystram:
Allright, I think this is ready for review now. I highlighted a couple of things I am unsure of.

src/rpm_ostree/cli_deploy.rs
jlebon: > I can add the lock finalization to `rpm-ostree rebase`.

Just to follow-up on this, `--lock-finalization` already works with `rpm-ostree rebase`. It's [a hidden switch](https://github.com/coreos/rpm-ostree/blob/e529101f7444f9d4225e1a472ff587ec3d989bb0/src/app/rpmostree-builtin-rebase.cxx#L71), which is why it doesn't show up in `rpm-ostree --help`.

jbtrystram: Thanks jlebon ! I think you already mentioned it to me OOB because I went to update the PR and it's already there :)

src/rpm_ostree/mod.rs
jlebon, line 100: We can also add `ostree-ext` and use the parsing functions available there to parse the ref instead.

src/update_agent/actor.rs
```diff
@@ -478,18 +478,20 @@ impl UpdateAgentInfo {
...
+            return Ok(false);
         }
         if pending.checksum != release.checksum {
-            bail!(
+            log::error!(
                 "detected checksum mismatch for pending deployment '{}', got unexpected value '{}'",
                 release.version,
                 release.checksum,
             );
+            return Ok(false);
```
jlebon, line 494: Hmm, why do we need this?

jbtrystram: The rest of the function was built that way, and the bail error text was swallowed so I just aligned all the checks to use the same syntax

src/rpm_ostree/actor.rs
```diff
@@ -52,8 +52,52 @@ impl Handler<StageDeployment> for RpmOstreeClient {
...
+        let release = if msg.release.is_oci {
+            if let Some(custom_origin) = custom_origin {
+                // make sure the update comes from the same origin
+                let release_pullspec_split: Vec<&str> =
+                    msg.release.checksum.split("@sha256:").collect();
+                if !custom_origin[0].contains(release_pullspec_split[0]) {
+                    anyhow::bail!(
+                        "The update pullspec provided does not match local custom-origin-url",
+                    );
+                }
```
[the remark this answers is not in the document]
jlebon: > Then how do the users knows which stream they are on in rpm-ostree status if they see the pull spec ? through the description ?

Yeah, we can leave the custom description as it is.

src/update_agent/actor.rs
jlebon: This changes the semantics of the function. I think those differences might be on purpose.

In the OSTree case, if the checksum doesn't match, then something went really wrong because we expicitly deployed by checksum. Whereas the stream check below this line is done here because it can't be done before deploying; it's part of the commit metadata. (Though I guess we could pull it before deploying.)

We can't exactly transpose this to the OCI case, because there while we do deploy by digest, that digest doesn't actually mean much within the object store. That said, here I think `checksum` is actually the pullspec in the OCI path as this PR is written (right?), so we do also expect it to be always true. It's basically checking that the custom origin URL we've set has correctly been set. I think then we do still want to keep this a hard error?

For the version check, I think the change to not make it a hard error makes sense.

src/rpm_ostree/cli_status.rs
jbtrystram: haven't had the need for it in the Zincati context

jbtrystram:
Allright, I reworked this PR a bunch following the review comments, that ended up being pretty substantial changes. 

## Overview : 

- A local deployment is now unserialized with an enum containing either a `Checksum` or a `Pullspec`. Both are simply strings, but I use that to `match` my way around, rather than a `is_oci` boolean.
- custom-origin fields are now purely cosmetic : Zincati will cue from the `container-image-reference`.
- Now properly parses the OCI manifest output from `rpm-ostree --json` and get the `fedora-coreos.stream` key from that. This is working around https://github.com/coreos/rpm-ostree/issues/5196
-  imported a couple of new crates : `oci-spec` for the above issue and `ostree-rs` to properly parse ostree image references and containers pull specs. I removed the hacky `split(':')` and split(":sha256")` in favor of that.


## Testing 
Here are my notes to test this out it case someone wants to experiment: 

Start a VM with Zincati disabled : `cosa run --qemu-image fedora-coreos-41.20241215.2.0-qemu.x86_64.qcow2 --add-ignition noautoupdate`

Rebase to an OCI image in graph (recent enough to get `rpm-ostree-2024.9)` : 
```
    sudo rpm-ostree rebase ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos@sha256:74448159fae255797012a12eaf9089ea3c1018ffb0b3e76e03483cba59b50bb2 \
     --custom-origin-url=quay.io/fedora/fedora-coreos@sha256:74448159fae255797012a12eaf9089ea3c1018ffb0b3e76e03483cba59b50bb2 \
     --custom-origin-description="Fedora CoreOS testing stream"
     sudo reboot
 ``` 
Now import Zincati build with this PR then:     
```
    sudo rm /etc/zincati/config.d/90-disable-auto-updates.toml
    sudo systemctl restart zincati
```
 Zincati should now pick up an OCI update through the OCI graph

### Getting zincati custom build in the COSA VM

While we can make a custom FCOS build with an override to get our Zincati build inside, it will be erased by the content of the OCI image with a rebase, so here is how I did, after building zincati and copy the binary into `$COSA_DIR/tmp`

``` 
rpm-ostree usroverlay
cp /mnt/workdir-tmp/zincati /usr/libexec/zincati
cp /mnt/workdir-tmp/zincati.rules /usr/share/polkit-1/rules.d/zincati.rules

# restart polkit
systemctl restart polkit.service

# re enable updates
rm /etc/zincati/config.d/90-disable-auto-updates.toml

# restart zincati
systemctl restart zincati.service

# Show the update
journalctl --no-pager -o cat -u zincati --follow
```
In the meantime, I'll squash this and try to make commits that make sense :)

File: src/rpm_ostree/cli_finalize.rs [hunk 10]
```diff
@@ -18,18 +18,27 @@ lazy_static::lazy_static! {
...
+        .arg("finalize-deployment");
 
-    if !cmd.status.success() {
+    // XXX for OCI image, we don't know the checksum until we deployed it.
+    // Currently, rpm-ostree do not return the resulting ostree commit
+    // when rebasing to an OCI image. We could query the deployements and
+    // get the latest commit but that would be racy, so let's finalize the latest
+    // commit.
+    if release.is_oci {
+        cmd.arg("--allow-missing-checksum")
```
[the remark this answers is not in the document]
jbtrystram: Yes, definitely, but that requires work in rpm-ostree, let's not block on it ?

src/update_agent/actor.rs
jbtrystram: Reverted this as per discsussion

src/cincinnati/mod.rs
```diff
@@ -232,7 +232,13 @@ fn find_update(
         .find(|(_, node)| is_same_checksum(node, &booted_depl.checksum))
     {
         Some(current) => current,
-        None => return Ok(None),
+        None => {
+            log::warn!(
+                "booted deployment {} not found in the update graph",
+                &booted_depl.payload
```
dustymabe, line 242: At this point in time (this commit) I don't think `rpm_ostree::Release` has been updated from `pub checksum: String,` to `pub payload: Payload,`.

Maybe this commit should go last.

.github/workflows/rust.yml
dustymabe, line 1: would be good to add in the commit message of this commit why we are making this change

dustymabe (commented):
I've gone through the review and feel less good about my recent rust training, but at least I was able to attempt to understand most of it!

src/cincinnati/mod.rs
jbtrystram: good catch ! I'll reorder

.github/workflows/rust.yml
jbtrystram: done :)

dustymabe: it's better.. but you mention COSA container, but actually fcos-buildroot is the container image being used. Maybe you should mention that one instead?

src/cincinnati/mod.rs
```diff
@@ -236,7 +236,13 @@ fn find_update(
         .find(|(_, node)| is_same_checksum(node, &booted_depl))
     {
         Some(current) => current,
-        None => return Ok(None),
+        None => {
+            log::warn!(
```
jlebon, line 240: Fine here, though this is something we'll probably have to undo as part of https://github.com/coreos/fedora-coreos-tracker/issues/1872.

jlebon (commented):
Nice, I really like the `Payload` model and how it makes things more explicit throughout the codebase. Lots of comments but looks sane overall. I still need to test it.

src/rpm_ostree/cli_deploy.rs
```diff
@@ -94,16 +98,37 @@ fn invoke_cli_register() -> Result<()> {
...
+            if let Some(origin) = custom_origin {
+                cmd.arg("rebase")
+                    .arg(image)
+                    .arg("--lock-finalization")
+                    .arg("--custom-origin-url")
+                    .arg(origin.url.clone())
+                    .arg("--custom-origin-description")
+                    .arg(origin.description.clone());
+            } else {
+                bail!("Missing custom-origin information.")
```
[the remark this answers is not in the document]
jbtrystram: yeah, that's a left over from when i was cuing from the custom-origin rather than parsing the OCI manifest to find the stream. Good catch

dustymabe:
gave a test to this.. started with `41.20250215.1.0` qemu image.. then rebased to the containerized version:

```
sudo rpm-ostree rebase ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918 \                                                                                                                                                          
 --custom-origin-url=quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918 \                                  
 --custom-origin-description="Fedora CoreOS next stream"
```


Then got the `zincati` binary and the updated polkit rules into the system via `mount --bind` and started zincati:

```
[core@cosa-devsh ~]$ sudo systemctl start zincati 
Feb 25 21:57:26 cosa-devsh systemd[1]: Starting zincati.service - Zincati Update Agent...
                                                                                         Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::cli::agent] starting update agent (zincati 0.0.29)
                                                                                                                                                                                                     Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::cincinnati] Cincinnati service: https://updates.coreos.fedoraproject.org
                                                         Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::cli::agent] agent running on node '490dad56e52c4d118016f8868adab486', in update group 'default'
                                                                                                                                                                                                                  Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::update_agent::actor] registering as the update driver for rpm-ostree
                                                                  [core@cosa-devsh ~]$ Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::update_agent::actor] found 1 other finalized deployment
Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::update_agent::actor] deployment 41.20250130.1.0 (c1fed7e59bad26ced87cfad72681259a886579c35e41de8ad6ebfeb79c49297c) will be excluded from being a future update target
Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::update_agent::actor] initialization complete, auto-updates logic enabled
Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::strategy] update strategy: immediate
Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::update_agent::actor] reached steady state, periodically polling for updates
Feb 25 21:57:26 cosa-devsh systemd[1]: Started zincati.service - Zincati Update Agent.
Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::cincinnati] current release detected as not a dead-end
Feb 25 21:57:26 cosa-devsh zincati[1706]: [INFO  zincati::update_agent::actor] target release '41.20250215.1.0' selected, proceeding to stage it
```

During the transaction (while it was in progress) I saw:


```
[core@cosa-devsh ~]$ rpm-ostree status 
State: busy
AutomaticUpdatesDriver: Zincati
  DriverState: active; found update on remote: 41.20250215.1.0
Transaction: rebase 'ostree-remote-image:fedora:docker://quay.io/fedora/fedora-coreos@sha256:40c0e515ed93cb8a9d5a773fb5a206b13e2edda02d92c6011a84399ceb31ed03' --lock-finalization --custom-origin-url 'quay.io/fedora/fedora-coreos@sha256:40c0e515ed93cb8a9d5a773fb5a206b13e2edda02d92c6011a84399ceb31ed03' --custom-origin-description 'Fedora CoreOS next stream' --disallow-downgrade 
  Initiator: caller :1.43
Deployments:
● quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918
             CustomOrigin: Fedora CoreOS next stream
                  Version: 41.20250130.1.0 (2025-01-31T20:20:28Z)

  fedora:fedora/x86_64/coreos/next
                  Version: 41.20250130.1.0 (2025-01-31T20:20:28Z)
                   Commit: c1fed7e59bad26ced87cfad72681259a886579c35e41de8ad6ebfeb79c49297c
             GPGSignature: Valid signature by 466CF2D8B60BC3057AA9453ED0622462E99D6AD1
```


It seemed to update fully.. though I was surprised it went down for the reboot immediately (usually zincati will tell me I have interactive sessions and wait 10 minutes):

```
Feb 25 21:58:03 cosa-devsh zincati[1706]: [INFO  zincati::update_agent::actor] update staged: 41.20250215.1.0
Feb 25 21:58:03 cosa-devsh zincati[1706]: [ERROR zincati::update_agent] failed to check for interactive sessions: failed to deserialize output of `loginctl`
Feb 25 21:58:03 cosa-devsh zincati[1706]: [WARN  zincati::update_agent] assuming no active sessions and proceeding anyway
Feb 25 21:58:03 cosa-devsh zincati[1706]: [INFO  zincati::update_agent::actor] staged deployment '41.20250215.1.0' available, proceeding to finalize it

Broadcast message from root@localhost (Tue 2025-02-25 21:58:03 UTC):

The system will reboot now!
```

Clearly `failed to deserialize output of 'loginctl'` is something we should look into.

Upon log in it's updated:

```
The system will reboot now!

Fedora CoreOS 41.20250215.1.0
Tracker: https://github.com/coreos/fedora-coreos-tracker
Discuss: https://discussion.fedoraproject.org/tag/coreos

Last login: Tue Feb 25 21:58:16 2025
[core@cosa-devsh ~]$ 
[core@cosa-devsh ~]$ rpm-ostree status 
State: idle
Deployments:
● quay.io/fedora/fedora-coreos@sha256:40c0e515ed93cb8a9d5a773fb5a206b13e2edda02d92c6011a84399ceb31ed03
             CustomOrigin: Fedora CoreOS next stream
                  Version: 41.20250215.1.0 (2025-02-17T11:44:02Z)

  quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918
             CustomOrigin: Fedora CoreOS next stream
                  Version: 41.20250130.1.0 (2025-01-31T20:20:28Z)
```

dustymabe:
> Clearly failed to deserialize output of 'loginctl' is something we should look into.

https://github.com/coreos/fedora-coreos-tracker/issues/1880

src/rpm_ostree/cli_status.rs
```diff
@@ -61,21 +62,21 @@ pub struct Status {
...
     staged: bool,
     version: String,
 }
 
-/// Metadata from base commit (only fields relevant to zincati).
+/// Custom origin fields
 #[derive(Clone, Debug, Deserialize)]
-struct BaseCommitMeta {
-    #[serde(rename = "fedora-coreos.stream")]
-    stream: Option<String>,
```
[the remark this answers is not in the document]
jbtrystram: ohh yes that's way more elegant,  let serde populate the options accordingly. Thank you for the suggestion ! I somehow thought that serde could not deserialize that because both don't exist at the same time, but the options make it work.

src/rpm_ostree/cli_status.rs
```diff
@@ -90,30 +91,63 @@ impl Deployment {
...
-    let stream = deploy
-        .base_metadata
-        .stream
-        .as_ref()
-        .ok_or_else(|| anyhow!("Missing `fedora-coreos.stream` in commit metadata"))?;
+    let stream = if deploy.container_image_reference.is_some() {
+        // This is the edge case where we spoofed the origin file to make it looks like a container
+        // native deployment but it's still pure ostree.
+        // here, the base commit meta deserialize nicely and we don't need the "unstringify" operation.
+        if let Some(stream) = deploy.base_commit_meta.get("fedora-coreos.stream") {
```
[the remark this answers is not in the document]
jbtrystram: Unfortunately i can't use the const in serde macros : https://github.com/serde-rs/serde/issues/450
So that would be a bit weird to have it in a const but still hard-coded in the macro

src/rpm_ostree/cli_status.rs
```diff
@@ -90,30 +91,63 @@ impl Deployment {
...
     Ok(status.into_release())
 }
 
 fn fedora_coreos_stream_from_deployment(deploy: &Deployment) -> Result<String> {
-    let stream = deploy
-        .base_metadata
-        .stream
-        .as_ref()
-        .ok_or_else(|| anyhow!("Missing `fedora-coreos.stream` in commit metadata"))?;
+    let stream = if deploy.container_image_reference.is_some() {
```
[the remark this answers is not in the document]
jbtrystram: changed the flow of this

src/rpm_ostree/cli_finalize.rs
```diff
@@ -18,18 +18,26 @@ lazy_static::lazy_static! {
...
+    let mut cmd = std::process::Command::new("rpm-ostree");
+    cmd.env("RPMOSTREE_CLIENT_ID", "zincati")
+        .arg("finalize-deployment");
 
-    if !cmd.status.success() {
+    // XXX for OCI image, we don't know the checksum until we deployed it.
+    // Currently, rpm-ostree do not return the resulting ostree commit
+    // when rebasing to an OCI image. We could query the deployements and
+    // get the latest commit but that would be racy, so let's finalize the latest
+    // commit.
```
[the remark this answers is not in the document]
dustymabe: can we `s/deployements/deployments` in all cases where it exists?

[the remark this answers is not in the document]
jbtrystram: dammit I did it again ! I am sorry, I keep trying to sneak french in there :)

jlebon (commented):
To be clear again though, this will make https://github.com/coreos/zincati/issues/498 much more visible once we switch over to this. Ideally, we fix that wart before switching production nodes over.

OK, whipped up https://github.com/coreos/zincati/pull/1259 for now.

jbtrystram:
Ok I just re-did the testing, OCI and regular OSTree updates works as expected. 
I think we are good to merge this if there are no further comments on the code

jlebon (approved):
Nice work!

jbtrystram:
I just pushed because of an earlier unfortunate `cargo update` that ran into https://github.com/coreos/zincati/issues/1176
Now there should not be any changes in the lockfile other than what was added with `ostree` :)

jlebon:
> I just pushed because of an earlier unfortunate `cargo update` that ran into #1176

Opened https://github.com/coreos/zincati/pull/1260 for that one. Let's try to get the dependency bumps into this release.

PR #1259, opened 2025-02-26
Add new `zincati-update-now` command

jlebon:
This addresses the gap identified in
https://github.com/coreos/zincati/issues/498 in a blunt way by setting some runtime config knobs so Zincati in effect immediately downloads, deploys, and reboots the system.

The motivation to at least have this for now is
https://github.com/coreos/zincati/pull/1241, which will purposely break `rpm-ostree upgrade --bypass-driver`. But clearly the approach here is applicable regardless of whether the node has switched to OCI.

In the future, this may either get replaced by a proper `zincatictl` entrypoint, or by better integration into the system's update tooling (e.g. bootc), so that `bootc upgrade` does the right thing.

### DIFF
--- a/dist/bin/zincati-update-now
+++ b/dist/bin/zincati-update-now
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
+
+echo "WARN: This command is experimental and subject to change." >&2
+
+# this should exist already, but in case
+mkdir -p /run/zincati/config.d
+cat > /run/zincati/config.d/99-update-now.toml << EOF
+[identity]
+rollout_wariness = 0.0
+[updates]
+strategy = "immediate"
+EOF
+
+touch /run/zincati/override-interactive-check
+
+systemctl daemon-reload
+systemctl restart zincati
+
+echo "INFO: Streaming Zincati and RPM-OSTree logs..." >&2
+exec journalctl -f -u zincati -u rpm-ostreed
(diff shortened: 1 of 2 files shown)


dist/bin/zincati-update-now
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
+
+echo "WARN: This command is experimental and subject to change." >&2
+
+# this should exist already, but in case
+mkdir -p /run/zincati/config.d
```
prestist, line 7: So I am very much out of my depth here but I have to ask in what circumstance would this not exist but the services exist?  
would that be an indication of other issues that might want to halt the update attempt ?

prestist (approved):
From a code perspective LGTM!

dustymabe:
> In the future, this may either get replaced by a proper `zincatictl` entrypoint

and I'm guessing we didn't name this `zincatictl` with a subcommand (i.e. `zincatictl update-now`) because you want to explicitly get away from painting us into a UI corner for something that was whipped up quickly?

dist/bin/zincati-update-now
dustymabe: doubt it, I think this only means that somehow the [tmpfiles.d fragment](https://github.com/coreos/zincati/blob/f50707131c7f823068c6ef410817b2f064c8424a/dist/tmpfiles.d/zincati.conf#L5) didn't run

jlebon:
> > In the future, this may either get replaced by a proper `zincatictl` entrypoint
> 
> and I'm guessing we didn't name this `zincatictl` with a subcommand (i.e. `zincatictl update-now`) because you want to explicitly get away from painting us into a UI corner for something that was whipped up quickly?

I didn't think of it that way quite exactly, but yes I did intend to very clearly emphasize the stopgap nature of this.

jbtrystram (approved):
LGTM

dustymabe:
Tested this out and it seemed to work well!

PR #1260, opened 2025-02-27
tests/kola/experimental: fix LastRefreshTime method name

jlebon:
We were using the wrong method name here.

Likely fixes https://github.com/coreos/zincati/issues/1176.

(E.g. maybe what happened there is that previously zbus would return "Access denied" for unauthorized users even for non-existent methods and maybe the logic changed so the existence check was done first.)

### DIFF
--- a/tests/kola/dbus/test-experimental.sh
+++ b/tests/kola/dbus/test-experimental.sh
@@ -9,7 +9,7 @@ set -xeuo pipefail
 cd $(mktemp -d)
 
 # Ensure that methods in this interface can only be called by root.
-if sudo -u core busctl call org.coreos.zincati /org/coreos/zincati org.coreos.zincati.Experimental LastRefresh 2> err.txt; then
+if sudo -u core busctl call org.coreos.zincati /org/coreos/zincati org.coreos.zincati.Experimental LastRefreshTime 2> err.txt; then
   fatal "Non-root user calling Experimental interface unexpectedly succeeded"
 fi
 assert_file_has_content err.txt "Access denied"


HuijingHei:
I tried manually with the build and failed, it is wired that `busctl list` does not even include `org.coreos.zincati`
```
# busctl call org.coreos.zincati /org/coreos/zincati org.coreos.zincati.Experimental LastRefreshTime
Call failed: The name is not activatable
```

jlebon:
> I tried manually with the build and failed, it is wired that `busctl list` does not even include `org.coreos.zincati`
> 
> ```
> # busctl call org.coreos.zincati /org/coreos/zincati org.coreos.zincati.Experimental LastRefreshTime
> Call failed: The name is not activatable
> ```

Hmm, I couldn't reproduce it locally.

Oh! I bet the test is racing with Zincati actually setting up D-Bus. Can you try adding a e.g. a loop that waits until `zincati.service` is active before we do the `busctl` calls?

jbtrystram:
> > I tried manually with the build and failed, it is wired that `busctl list` does not even include `org.coreos.zincati`
> > ```
> > # busctl call org.coreos.zincati /org/coreos/zincati org.coreos.zincati.Experimental LastRefreshTime
> > Call failed: The name is not activatable
> > ```
> 
> Hmm, I couldn't reproduce it locally.
> 
> Oh! I bet the test is racing with Zincati actually setting up D-Bus. Can you try adding a e.g. a loop that waits until `zincati.service` is active before we do the `busctl` calls?

I am able to reproduce this issue, even with this fix. Did you update zbus ?

jlebon:
OK yes, I can reproduce it now. But actually, Zincati isn't exposing a D-Bus service at all (which explains the error message), so there's something more fundamental going wrong here. I think possibly related to the way we use zbus+actix.

HuijingHei:
> OK yes, I can reproduce it now. But actually, Zincati isn't exposing a D-Bus service at all (which explains the error message), so there's something more fundamental going wrong here. I think possibly related to the way we use zbus+actix.

Thanks! It is out of my knowledge now.

PR #1261, opened 2025-02-27
update_agent: adjust for recent changes in loginctl

dustymabe:
loginctl no longer accepts the `--output=json` argument [1]. Adjust
here to use `--json=short` instead. Also, loginctl now outputs
`null` when there is no tty [2], so we can drop the custom
deserialize_systemd_tty_canonicalized deserializer.

[1] https://github.com/systemd/systemd/issues/34901
[2] https://github.com/systemd/systemd/commit/febfec0849616cb8d7d63f66c6dc05ca48a6a1e7

This should fix https://github.com/coreos/fedora-coreos-tracker/issues/1780

### DIFF
--- a/src/update_agent/mod.rs
+++ b/src/update_agent/mod.rs
@@ -12,7 +12,7 @@ use actix::Addr;
 use anyhow::{Context, Result};
 use chrono::prelude::*;
 use prometheus::{IntCounter, IntGauge};
-use serde::{Deserialize, Deserializer};
+use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


cgwalters:
> Adjust here to use --json=short instead. 

If we were shipping this into RHEL I'd be concerned about handling skew (i.e. handling old systemd too). But that's not a concern here as we're not doing zincati there (and I don't think we should).

jbtrystram (approved):
lgtm

Issue #1272, opened 2025-03-03
leverage container-image-reference-digest from rpm-ostree for OCI updates


dustymabe:
Was discussing with @jlebon and found out that once we have switched and are using OCI for updates we'll be in a position where the analog of `sudo rpm-ostree rebase "fedora/${ARCH}/coreos/${STREAM}"` won't really work.

i.e. if you `sudo rpm-ostree rebase ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos:stable` then your zincati updates will break.

I think we need to still keep this working and *I think* we can. Enough information is in the `rpm-ostree status --json` output:

```
[core@cosa-devsh ~]$ rpm-ostree status --json | grep container-image
      "container-image-reference-digest" : "sha256:b5bdaa44a45084bf572c8c0f48d39c30b6c365941e86e60fcefb65aa9836ed9c",
      "container-image-reference" : "ostree-remote-image:fedora:docker://quay.io/fedora/fedora-coreos:rawhide",
```

So we have the container digest as well as the name we are following so we should be able to pull off the tag from what we are following and use the digest instead.

We could actually lean in to this strategy even more by relying on it in zincati by default (i.e. not just handling a corner case but making it part of the strategy), which would allow for more information to be shown to the user without having to leverage CustomOrigin:

i.e. this:

```
[core@cosa-devsh ~]$ rpm-ostree status 
State: idle
Deployments:
● ostree-remote-image:fedora:registry://quay.io/fedora/fedora-coreos:testing
                   Digest: sha256:b5bdaa44a45084bf572c8c0f48d39c30b6c365941e86e60fcefb65aa9836ed9c
                  Version: 43.20250303.dev.0 (2025-03-03T19:41:42Z)
```

versus:

```
[core@cosa-devsh ~]$ rpm-ostree status 
State: idle
Deployments:
● quay.io/fedora/fedora-coreos@sha256:b5bdaa44a45084bf572c8c0f48d39c30b6c365941e86e60fcefb65aa9836ed9c
             CustomOrigin: Fedora CoreOS testing stream
                  Version: 41.20250215.1.0 (2025-02-17T11:44:02Z)

```

It shows you what "stream" we are following along with the actual digest instead of relying on customOrigin.

cgwalters:
> --experimental

(tangent) You don't need this anymore for quite some time, where did you find it?

> then your zincati updates will break.

Can you be more specific? What breaks? Why?

jlebon:
[...]

PR #1273, opened 2025-03-05
rpm-ostree/status: Parse container-image-reference-digest

jbtrystram:
This pull the container digest from the rpm-ostree status, independently of what `container-image-reference` is.
This value is then used to find ourselves in the update graph.

This allow to rebase to a tag, (version or stream) while still allowing zincati to pick-up updates.

However, if no custom description is given, rpm-ostree status won't easily show what stream the node is following:

```
[root@cosa-devsh core]# rpm-ostree status
State: idle
AutomaticUpdatesDriver: Zincati
  DriverState: active; update staged: 41.20250215.2.0; reboot delayed due to active user sessions
Deployments:
  ostree-remote-image:fedora:docker://quay.io/fedora/fedora-coreos@sha256:e60a16be2dd7e6fdcb30e61a452ced46d0e8e473810d25f82a34eee78f6ce430
                   Digest: sha256:e60a16be2dd7e6fdcb30e61a452ced46d0e8e473810d25f82a34eee78f6ce430
                  Version: 41.20250215.2.0 (2025-02-17T11:43:37Z)
                     Diff: 125 upgraded

● ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos:41.20250105.2.0
                   Digest: sha256:c4a15145a232d882ccf2ed32d22c06c01a7cf62317eb966a98340ae4bd56dfa6
                  Version: 41.20250105.2.0 (2025-01-06T19:46:25Z)
```
See https://github.com/coreos/zincati/issues/1272 for more details.


src/rpm_ostree/cli_status.rs
```diff
@@ -121,11 +123,42 @@ impl Deployment {
...
             .as_ref()
             .and_then(|s| s.as_str().try_into().ok())
     }
+    /// Return the deployed container image as an oci image reference
+    /// but with the digest instead of the tag
+    /// e.g. quay.io/fedora/fedora-coreos@sha256:c4a15145a232d882ccf2ed32d22c06c01a7cf62317eb966a98340ae4bd56dfa6
+    pub fn container_image_reference_digest(&self) -> Option<Reference> {
+        match (
+            &self.container_image_reference(),
+            &self.container_image_reference_digest,
```
dustymabe, line 138: I found it slightly confusing that this is essentailly a variable name and also the name of this function..

i.e. `self.container_image_reference_digest` and `self.container_image_reference_digest()` are different things (or maybe I'm misunderstanding). 

Might help if we renamed this function `self.get_container_image_reference_digest()` ?

dustymabe (commented):
Some comments

src/rpm_ostree/cli_status.rs
jbtrystram: good suggestion !

src/update_agent/actor.rs
```diff
@@ -552,7 +551,7 @@ impl UpdateAgentInfo {
     ///
     /// If valid, advance the state-machine. If mistaken, clean up the pending deployment.
     /// If unsure, bubble up errors in order to retry later.
-    #[context("failed to confirm stream")]
+    //#[context("failed to confirm stream")]
```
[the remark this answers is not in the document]
jbtrystram: yeah sorry I should have push the PR as a draft there are still a couple of things to clean up

jbtrystram:
@jlebon using `deploy` the following flow does not work : 
rease to a version tag, e.g. `ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos:41.20241215.2.0`
Then zincati pick-up the stream from the commit meta and does : `rpm-ostree deploy sha256:7bb94d1516fc6e368eb9cf07a4fe23ca7378e2e1ddb76c6a1902f75cb5c9203a`. 
rpm-ostree fails with:
```
Mar 07 16:32:12 cosa-devsh zincati[2137]: [ERROR zincati::update_agent::actor] failed to stage deployment: rpm-ostree deploy failed:
Mar 07 16:32:12 cosa-devsh zincati[2137]:     error: Importing: Unencapsulating base: failed to invoke method GetBlob: failed to invoke method GetBlob: fetching blob: received unexpected HTTP status: 504 Gateway Time-out
```

src/rpm_ostree/cli_status.rs
```diff
@@ -90,8 +92,8 @@ struct BaseCommitMeta {
 impl Deployment {
     /// Convert into `Release`.
     pub fn into_release(self) -> Release {
-        let payload = if let Some(image) = self.container_image_reference {
-            Payload::Pullspec(image)
+        let payload = if let Some(reference) = self.container_image_reference_digest() {
+            Payload::Pullspec(reference)
```
dustymabe, line 96: This probably works fine outside of this one corner case, but in the case where we're trying to run the [migration script](https://github.com/coreos/fedora-coreos-config/pull/3355) `container_image_reference_digest()` will return `None` because the digest isn't actually in the JSON:




```
[core@cosa-devsh ~]$ sudo cat /ostree/deploy/fedora-coreos/deploy/c1fed7e59bad26ced87cfad72681259a886579c35e41de8ad6ebfeb79c49297c.0.origin
[origin]
container-image-reference=ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918
custom-url=quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918
custom-description=Fedora CoreOS testing stream
[core@cosa-devsh ~]$ 
[core@cosa-devsh ~]$ rpm-ostree status --json | grep container
        "ostree.container-cmd" : [
      "container-image-reference" : "ostree-remote-image:fedora:registry:quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918",
```

dustymabe:
I tested this on a system with no container deployments (i.e. making sure updates via OSTree continue to work) and it was happy.

I also tested this by manually rebasing a node first:

```
sudo rpm-ostree rebase ostree-remote-registry:fedora:quay.io/fedora/fedora-coreos@sha256:fcd6c0e85b1f80ba23b01d280db9f3e273ba9e4bfde9d00820d5141404ae0918 --reboot
```

And then mounting in new zincati and rpm-ostree and it seemed to work.

With the migration script it seems the "faking out" we are doing is no longer sufficient. https://github.com/coreos/zincati/pull/1273#discussion_r1985713244

src/rpm_ostree/cli_status.rs
jbtrystram: This is solved with the new override booted-status approach

jbtrystram:
Tested this today : 

- regression test : on system with no container deployments -> zincati is happy to continue with ostree updates.
- manual rebase to oci with stream tag ->  zincati calls rpm-ostree deploy with the new digest
- manual rebase to oci with version tag -> zincati picks up the stream and rebase (to show the stream in UX)
- manual rebase to oci pinned to digest -> zincati picks up the stream and rebase (to show the stream in UX)
- migration testing by setting the override status file -> zincati picks up the stream and trigger a rebase to OCI .

src/update_agent/actor.rs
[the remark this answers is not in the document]
jbtrystram: Note for future reviewers : this swallow the error messages  that `is_pending_deployment_on_correct_stream` returns, making it harder to debug. That's why I removed it

travier:
For reference, from Dusty:
> let's make sure we test rolling back and then forward again (i.e. making sure the state machine doesn't get out of whack)

src/rpm_ostree/actor.rs
```diff
@@ -56,51 +59,91 @@ impl Handler<StageDeployment> for RpmOstreeClient {
...
+                    };
+
+                    let options = StagingOptions {
+                        allow_downgrade: msg.allow_downgrade,
+                        custom_origin,
+                        rebase: rebase_target,
+                    };
+
+                    trace!("request to stage release: {:?}", &msg.release);
+                    super::cli_deploy::deploy_locked(msg.release, options)
```
jlebon, line 131: I guess we could factor this out of the match statement now and have the match be purely about constructing `options`.

```rust
let options = match &msg.release.payload { ... };
trace!("request to stage release: {:?}", &msg.release);
super::cli_deploy::deploy_locked(msg.release, options) // last statement
```

src/rpm_ostree/cli_deploy.rs
```diff
@@ -98,25 +94,32 @@ fn invoke_cli_register() -> Result<()> {
...
-                    .arg(origin.description.clone());
+        Payload::Pullspec(reference) => {
+            if let Some(rebase_target) = staging_options.rebase {
+                cmd.arg("rebase").arg("--lock-finalization");
+                match reference.digest() {
+                    Some(digest) => {
+                        cmd.arg(rebase_target.to_string()).arg(digest);
+                    }
+                    // zincati payload are digested oci pullspecs
+                    _ => unreachable!(),
```
jlebon, line 111: I think of `unreachable!()` as more something that can't happen given the internal logic of the code itself. This one depends more on input external to us (the graph). Though in practice possibly we would fail earlier somewhere else if the payload wasn't digested...

But anyway, I guess it doesn't hurt to just do e.g. `let digest = reference.digest().ok_or_else(|| anyhow!())?;`.

src/rpm_ostree/cli_finalize.rs
```diff
@@ -22,31 +22,20 @@ pub fn finalize_deployment(release: Release) -> Result<Release> {
...
-                let staged_imgref = staged_depl
-                    .container_image_reference()
-                    .map(|i| i.to_string());
-                if staged_imgref.as_ref() == Some(release_imgref) {
-                    cmd.arg(staged_depl.ostree_checksum())
-                } else {
-                    bail!("The staged deployment does not match the update reference. Won't finalize.");
-                }
+        super::Payload::Pullspec(reference) => {
+            let digest = reference.digest();
```
jlebon, line 27: Similarly here.

src/rpm_ostree/cli_status.rs
```diff
@@ -17,6 +18,14 @@ use std::rc::Rc;
 /// of the current state of deployments.
 const OSTREE_DEPLS_PATH: &str = "/ostree/deploy";
 
+/// Path to the fake deployment to use when migrating to OCI transport.
+/// Using this fake deployment instead of the booted one, zincati
+/// will get the next update from the OCI graph and rebase to the OCI image.
+/// See https://github.com/coreos/fedora-coreos-tracker/issues/1823
+static BOOTED_STATUS_OVERRIDE_FILE: &str = "/run/zincati/booted-status-override.json";
+/// Value we expect as the checksum in the overridden status to signal the rebase.
+pub static OCI_MIGRATION_CHECKSUM_SIGNAL: &str = "oci-rebase";
```
jlebon, line 27: As mentioned above, I _think_ we can drop this magic value.

src/rpm_ostree/cli_status.rs
```diff
@@ -62,6 +71,7 @@ pub struct Status {
 pub struct Deployment {
     booted: bool,
     container_image_reference: Option<String>,
+    container_image_reference_digest: Option<String>,
     custom_origin: Option<CustomOrigin>,
```
jlebon, line 75: Basically, I'd just nuke this field entirely and all the related bits and stop caring about it.

src/rpm_ostree/cli_status.rs
```diff
@@ -275,7 +306,22 @@ pub fn invoke_cli_status(booted_only: bool) -> Result<Status> {
...
+    let status_override_file = std::path::Path::new(BOOTED_STATUS_OVERRIDE_FILE);
+    if status_override_file.exists() {
+        let rdr = std::fs::File::open(status_override_file).map(std::io::BufReader::new)?;
+        let override_boot_depl: Deployment = serde_json::from_reader(rdr)?;
+
+        // Keep the actual status info and just replace the booted deployement.
+        // We need other deployements info to know if we rollbacked
+        // or if a deployment is staged.
+        status.deployments.retain(|s| !s.booted);
+        status.deployments.push(override_boot_depl);
```
jlebon, line 323: Hmm, this changes ordering. I think it'd be safer to do e.g. 

```rust
status.deployments = status.deployments.into_iter().map(|d| {
    if d.booted { override_boot_depl } else { d }
}).collect();
```

src/rpm_ostree/cli_status.rs
```diff
@@ -328,25 +374,66 @@ mod tests {
...
+        // assert!(custom_origin.is_some());
+        // let custom_origin = custom_origin.unwrap();
+        // assert_eq!(
+        //     custom_origin.url,
+        //     "quay.io/fedora/fedora-coreos@sha256:c4a15145a232d882ccf2ed32d22c06c01a7cf62317eb966a98340ae4bd56dfa6"
+        // );
+        // assert_eq!(
+        //     custom_origin.description,
+        //     "Fedora CoreOS Testing stream through OCI images"
+        // );
```
jlebon, line 403: Leftover?

src/rpm_ostree/mod.rs
```diff
@@ -36,15 +40,29 @@ pub struct Release {
 pub enum Payload {
     /// Represent a pure OSTree checksum
     Checksum(String),
-    /// an OCI image name
-    Pullspec(String),
+    /// an OCI image reference
+    Pullspec(Reference),
+}
+
+/// Optionnal details when staging a new deployement.
```
jlebon, line 47: Optional

src/rpm_ostree/mod.rs
```diff
@@ -36,15 +40,29 @@ pub struct Release {
...
+#[derive(Clone, Debug)]
+pub struct StagingOptions {
```
jlebon, line 49: This makes sense to me, though good also to go back to individual arg passing once we drop the custom origin bits.

src/update_agent/actor.rs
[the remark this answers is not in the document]
jlebon: Hmm weird. It should append to the error message, not override it. Is that what you're seeing?

jlebon (commented):
A lot of comments, but overall this looks really nice!

src/cincinnati/mod.rs
```diff
@@ -320,30 +319,11 @@ fn find_update(
...
-            Payload::Pullspec(imgref) => {
-                let ostree_imgref: Result<OstreeImageReference> = imgref.as_str().try_into();
-                // ostree_imgref here is `ostree-remote-image:fedora:docker://quay.io.....`
-                // while the cincinnati payload only contains `quay.io:///....`
-                Payload::Pullspec(
-                    ostree_imgref.map_or(imgref, |ostree_imgref| ostree_imgref.imgref.name),
-                )
-            }
-        })
-        .collect();
```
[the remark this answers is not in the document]
jbtrystram: I changed `Payload::Pullspec` to be an `oci_spec_distribution::Reference` instead of a string. 
So the ostree imgref bits are peeled of in `Deployment::into_release()` : https://github.com/coreos/zincati/pull/1273/files#diff-a3a9644aeb43678c5f1add5160544060e8bd8c6ea71fde4be20627be2ebe5336R103

This allow me tu use the nicer API with Reference such as `.tag()` , `.digest()` ... 
I switched to that because with your rpm-ostree changes, Zincati will do `deploy $digest` most of the time. I only need to know the ostree imgref bits when rebasing (which should be a rare operation)

src/rpm_ostree/actor.rs
```diff
@@ -56,51 +59,91 @@ impl Handler<StageDeployment> for RpmOstreeClient {
...
+                    // 2 - The image we are booted on does not match a stream tag, e.g. the node was manually
+                    //     rebased to a version tag. Here deploy would work but would lead to a weird UX:
+                    //     rpm-ostree status would show the version tag in the origin after we moved on to
+                    //     another version.
+                    // 3 - The image name we are following has changed (new registry, new name)
+                    //     In that case `deploy` won't work and we need to rebase to the new refspec.
+
+                    // case 1
+                    // Migrating from OSTree to OCI, signaled by the fake checksum
+                    let rebase_target = if local_deploy.get_checksum() == *OCI_MIGRATION_CHECKSUM_SIGNAL
```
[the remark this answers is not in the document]
jbtrystram: yes you are right ! I over-complicated this, if the migration script set a digest refspec it will trigger a rebase. Nice

src/rpm_ostree/actor.rs
```diff
@@ -56,51 +59,91 @@ impl Handler<StageDeployment> for RpmOstreeClient {
...
+                        // case 2: the booted tag does not match the stream
+                        || booted_oci_ref.tag() != local_deploy.get_fcos_update_stream().ok().as_deref()
+                        // case 3: new image location
+                        || booted_oci_ref.registry() != target_pullspec.registry()
+                        || booted_oci_ref.repository() != target_pullspec.repository()
```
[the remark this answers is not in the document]
jbtrystram: I feel that would be harder to read / understand

src/update_agent/actor.rs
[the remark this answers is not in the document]
jbtrystram: it's because it's not anyhow context, but https://docs.rs/fn-error-context/latest/fn_error_context/

I'll fix that

src/rpm_ostree/actor.rs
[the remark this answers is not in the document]
jbtrystram: Ok i refactored this, it's actually easier to follow (maybe, i stared at this code for way too long at this point)

src/rpm_ostree/actor.rs
jbtrystram: i ended up nuking the options structs since I removed the custom origin code. 
The match now is just about creating an `Option<OSTreeImageReference>` to rebase to

src/rpm_ostree/cli_deploy.rs
jbtrystram: Yeah if the cincinnati data is incorrect it should not go that far, but let's throw an error rather than panicking.

jbtrystram:
I addressed the review comments and split up into more individual commits where that make sense.
I'll do another round of testing and I think this is good to go!

dustymabe:
Found an issue this morning where the system gets in a weird state if Zincati somehow gets restarted after it has staged an OCI update, but before the system has rebooted into the new OCI deployment. 


So the reproducer here is to set everything up for migration and start zincati to stage the OCI update. Then reboot is blocked (i.e. interactive session detection). then stop and start zincati again and it gets into a state where it can't deploy anything.

See the logs from https://github.com/coreos/fedora-coreos-config/pull/3355#issuecomment-2721333793

jbtrystram:
> Found an issue this morning where the system gets in a weird state if Zincati somehow gets restarted after it has staged an OCI update, but before the system has rebooted into the new OCI deployment.
> 
> So the reproducer here is to set everything up for migration and start zincati to stage the OCI update. Then reboot is blocked (i.e. interactive session detection). then stop and start zincati again and it gets into a state where it can't deploy anything.
> 
> See the logs from [coreos/fedora-coreos-config#3355 (comment)](https://github.com/coreos/fedora-coreos-config/pull/3355#issuecomment-2721333793)

my initial idea was to make the override status pick whatever was staged if it exist, but could lead to skipping barrier code.
In `rpm_ostree/actor.rs`, zincati needs to learns to look at staged deployment as well.

jbtrystram:
I just tested this corner case with that last commit. 
What was happening is that zincati used the `booted` deployment to decide wether it should rebase or deploy. However, if a deployment is already staged but another update (or zincati restarts) comes in, rebasing again would yield an rpm-ostree error where it would not accept to rebase to the same refspec (as we use the $stream tagged refspec).
In that case we should deploy.

```
[root@cosa-devsh workdir-tmp]# rpm-ostree status
State: idle
AutomaticUpdatesDriver: Zincati
  DriverState: active; update staged: 41.20250302.2.0; reboot delayed due to active user sessions
Deployments:
  ostree-remote-image:fedora:docker://quay.io/fedora/fedora-coreos:testing
                   Digest: sha256:7bb94d1516fc6e368eb9cf07a4fe23ca7378e2e1ddb76c6a1902f75cb5c9203a
                  Version: 41.20250302.2.0 (2025-03-04T16:52:41Z)
                     Diff: 147 upgraded

● fedora:fedora/x86_64/coreos/testing
                  Version: 41.20241215.2.0 (2024-12-17T00:07:38Z)
                   Commit: 0e93f07d8d11856eb0773bde01757337320b26eb823cb5faff2c89d2849edb0a
             GPGSignature: Valid signature by 466CF2D8B60BC3057AA9453ED0622462E99D6AD1
                 Unlocked: development
[root@cosa-devsh workdir-tmp]# 
[root@cosa-devsh workdir-tmp]# systemctl restart zincati
[root@cosa-devsh workdir-tmp]# 
[root@cosa-devsh workdir-tmp]# journalctl --follow -u zincati.service 
....
Mar 13 15:40:19 cosa-devsh zincati[3662]: [TRACE zincati::update_agent::actor] update agent tick, current state: UpdateAvailable((Release { version: "41.20250302.2.0", payload: Pullspec(Reference { registry: "quay.io", mirror_registry: None, repository: "fedora/fedora-coreos", tag: None, digest: Some("sha256:7bb94d1516fc6e368eb9cf07a4fe23ca7378e2e1ddb76c6a1902f75cb5c9203a") }), age_index: Some(172) }, 0))
.....
Mar 13 15:40:19 cosa-devsh zincati[3662]: [TRACE zincati::rpm_ostree::cli_deploy] Requesting rpm ostree deploy with arguments: CommandArgs { inner: ["deploy", "--lock-finalization", "sha256:7bb94d1516fc6e368eb9cf07a4fe23ca7378e2e1ddb76c6a1902f75cb5c9203a", "--disallow-downgrade"] }
```

src/rpm_ostree/actor.rs
```diff
@@ -56,8 +56,16 @@ impl Handler<StageDeployment> for RpmOstreeClient {
     fn handle(&mut self, msg: StageDeployment, _ctx: &mut Self::Context) -> Self::Result {
         let rebase_target = match &msg.release.payload {
             Payload::Pullspec(target_pullspec) => {
-                let booted = super::cli_status::invoke_cli_status(true)?;
-                let local_deploy = super::cli_status::booted_status(&booted)?;
+                // If there is staged deployment we use that to determine if we should rebase or deploy
+                // Otherwise, fallback to booted.
+                // This is because if we already staged a rebase, rebasing again won't work
+                // as "Old and new refs are equal"
+                // see https://github.com/coreos/szincati/pull/1273#issuecomment-2721531804
```
jlebon, line 63: Hmm yes, tricky.

jlebon (approved):
LGTM! Comment not worth the respin.
Will defer to Dusty and you for final testing.
Feel free to merge when ready.

dustymabe (approved):
LGTM

If we find any other issues during further testing we can open new PRs for them.

PR #1276, opened 2025-03-11
Fixed minute format in periodic update log

dwysocki:
The log message reporting when the next update window is for the 'periodic' strategy does not format the minute nicely.  For instance, an update scheduled for `2:00` appears instead as `2:0` in the log, which may make new users think something is configured wrong.  This PR simply zero-pads the minute to ensure two digits are always displayed.

### DIFF
--- a/src/strategy/periodic.rs
+++ b/src/strategy/periodic.rs
@@ -132,7 +132,7 @@ impl StrategyPeriodic {
             Some(minute_in_week) => {
                 let (weekday, hour, minute) = utils::weekly_minute_as_weekday_time(minute_in_week);
                 format!(
-                    "at {}:{} on {} ({}), subject to time zone caveats.",
+                    "at {}:{:0>2} on {} ({}), subject to time zone caveats.",
                     hour, minute, weekday, self.tz_name
                 )
             }


src/strategy/periodic.rs
```diff
@@ -132,7 +132,7 @@ impl StrategyPeriodic {
             Some(minute_in_week) => {
                 let (weekday, hour, minute) = utils::weekly_minute_as_weekday_time(minute_in_week);
                 format!(
-                    "at {}:{} on {} ({}), subject to time zone caveats.",
+                    "at {}:{0>2} on {} ({}), subject to time zone caveats.",
```
dwysocki, line 135:
```suggestion
                    "at {}:{:0>2} on {} ({}), subject to time zone caveats.",
```
Of course I couldn't make a 3 character PR without a typo -- it should be a 4 character PR.

dwysocki (commented):
Fixed a typo

jbtrystram (changes requested):
Thanks for your contribution ! 
You can squash your two commits together to fix the typo ;)
Can you also update the commit message with the explanation for the change ? Your message in the PR is good.

dwysocki:
> Thanks for your contribution ! You can squash your two commits together to fix the typo ;) Can you also update the commit message with the explanation for the change ? Your message in the PR is good.

D'accord, c'est fini.

jbtrystram (approved):
LGTM, thanks !

PR #1277, opened 2025-03-13
quality of life improvements for zincati-update-now

dustymabe:
zincati-update-now: limit journal messages
    
    The `-f` will show 10 previous lines which include some lines from
    the previous instance of zincati. Let's just limit the messages to
    the ones from this point forward.

---

    zincati-update-now: don't block when restarting zincati
    
    This allows us to more quickly fall through to the `journalctl`
    following. I hit a case where it was taking some time for the
    restart operation to complete (zincati was crashlooping) and
    it would have helped to have already fallen through to the
    journal watching.

---

    zincati-update-now: give good error message when run without privs
    
    This is a bit better than what currently happens.

### DIFF
--- a/dist/bin/zincati-update-now
+++ b/dist/bin/zincati-update-now
@@ -23,4 +23,4 @@ systemctl daemon-reload
 systemctl restart zincati --no-block
 
 echo "INFO: Streaming Zincati and RPM-OSTree logs..." >&2
-exec journalctl -f -u zincati -u rpm-ostreed
+exec journalctl -f -u zincati -u rpm-ostreed --since now
--- a/dist/bin/zincati-update-now
+++ b/dist/bin/zincati-update-now
@@ -20,7 +20,7 @@ EOF
 touch /run/zincati/override-interactive-check
 
 systemctl daemon-reload
-systemctl restart zincati
+systemctl restart zincati --no-block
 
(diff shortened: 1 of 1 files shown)


dist/bin/zincati-update-now
```diff
@@ -23,4 +23,4 @@ systemctl daemon-reload
 systemctl restart zincati --no-block
 
 echo "INFO: Streaming Zincati and RPM-OSTree logs..." >&2
-exec journalctl -f -u zincati -u rpm-ostreed
+exec journalctl -f -u zincati -u rpm-ostreed --since now
```
jbtrystram, line 26: I think `--invocation=0` would be even nicer, as `--since-now` is leaving out on any messages between the restart and the execution of this. 
But that's being picky as it's an extremely short race :smile: 

And `-I` was just introduced in `v257` which isn't in F41. (it's in branched though so if this lands in the same release as the beta rebase that would be OK).

jbtrystram (approved):
One comment, not blocking but I thought I'd share as I just discovered it

dist/bin/zincati-update-now
dustymabe: nice - didn't know about `-I`. but yeah this is going to go into F41 too so let's not pick it up and use it yet

PR #1293, opened 2025-04-30
cincinnati: fix updates nodes comparison logic

jbtrystram:
In [1] i reworked how cincinnati nodes are turned into deployments for comparing them. I also simplified this logic to match already existing deployments with cincinnati nodes.

Basically the cincinnati node is turned into a deployment looking like:
```
Release {
    version: "42.20250410.3.1",
    payload: Checksum("e057a84658cd114c1fa8a944d8fe3de93f5413c0fbe3f1fabc737cc04bbd749b"),
    age_index: Some(144) }
```

And a local deployment looks like:
```
Release {
    version: "41.20250315.3.0",
    payload: Checksum("3c6418b40e2c7fe4605ff1fa744a845caadd50e10bd31aab14e059b529f49c98"),
    age_index: None }
```

Obviously, there is no age_index value for a local deployement as it's a cinncinati thing. Thus, a local deployment won't ever be equal to a cinncinati node.

Fix thi slogic by extracting the `Payload` enum and comparing those, as this is either a pullspec or an ostree checksum they will match. Which is was the original code was doing anyway.

[1] https://github.com/coreos/zincati/commit/ea83496bdaa8d438e83fb8d555299fdb4f46753d

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1938

### DIFF
--- a/src/cincinnati/mod.rs
+++ b/src/cincinnati/mod.rs
@@ -319,11 +319,14 @@ fn find_update(
 
 /// Try to match a set of (denylisted) deployments to their graph entries.
 fn find_denylisted_releases(graph: &client::Graph, depls: BTreeSet<Release>) -> BTreeSet<Release> {
+    use std::collections::HashSet;
+
     let mut local_releases = BTreeSet::new();
+    let local_payloads: HashSet<Payload> = depls.into_iter().map(|rel| rel.payload).collect();
 
     for entry in &graph.nodes {
         if let Ok(release) = Release::from_cincinnati(entry.clone()) {
-            if depls.contains(&release) {
+            if local_payloads.contains(&release.payload) {
                 local_releases.insert(release);
             }
         }


jbtrystram:
Looking at this code some more made me realize that there is a very specific edge case : if you migrate from ostree to OCI, then rollback AND opt out of OCI you will get the same version again but through the ostree. 

A fix for this would be to compare versions : 
```
diff --git a/src/cincinnati/mod.rs b/src/cincinnati/mod.rs
index fe4484a..e2baad0 100644
--- a/src/cincinnati/mod.rs
+++ b/src/cincinnati/mod.rs
@@ -322,11 +322,11 @@ fn find_denylisted_releases(graph: &client::Graph, depls: BTreeSet<Release>) ->
     use std::collections::HashSet;
 
     let mut local_releases = BTreeSet::new();
-    let local_payloads: HashSet<Payload> = depls.into_iter().map(|rel| rel.payload).collect();
+    let local_versions: HashSet<String> = depls.into_iter().map(|rel| rel.version).collect();
 
     for entry in &graph.nodes {
         if let Ok(release) = Release::from_cincinnati(entry.clone()) {
-            if local_payloads.contains(&release.payload) {
+            if local_versions.contains(&release.version) {
                 local_releases.insert(release);
             }
         }
```

But it's obviously not as strong as checksum / digests. Would there be a case when a version string could match multiple images ? (i don't think so but it's worth asking)

dustymabe:
> very specific edge case

I'm not too worried about it :)

dustymabe:
With this PR this is what I see now:

```
Starting zincati.service - Zincati Update Agent...
[INFO  zincati::cli::agent] starting update agent (zincati 0.0.30)
[INFO  zincati::cincinnati] Cincinnati service: https://updates.coreos.fedoraproject.org
[INFO  zincati::cli::agent] agent running on node '2b10f59f412d4543ad23c6f84e7624ad', in update group 'default'
[INFO  zincati::update_agent::actor] registering as the update driver for rpm-ostree
[INFO  zincati::update_agent::actor] found 1 other finalized deployment
[INFO  zincati::update_agent::actor] deployment 42.20250427.2.0 (quay.io/fedora/fedora-coreos@sha256:a8bc9e3ef585de1f4071bed845a8ac0b709b308ef0ef39ffb387df76002719eb) will be excluded from being a future update target
[INFO  zincati::update_agent::actor] initialization complete, auto-updates logic enabled
[INFO  zincati::strategy] update strategy: immediate
[INFO  zincati::update_agent::actor] reached steady state, periodically polling for updates
Started zincati.service - Zincati Update Agent.
[INFO  zincati::cincinnati] current release detected as not a dead-end
```

dustymabe (approved):

jlebon (approved):
Looks sane to me!

PR #1302, opened 2025-05-14
Sync repo templates ⚙

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/repo-templates/actions/workflows/sync.yml) ([source](https://github.com/coreos/repo-templates/blob/main/.github/workflows/sync.yml)).

Sync with coreos/repo-templates@49b82bf994af8b9e91606428dd5a89d628284924.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-checklist.md
+++ b/.github/ISSUE_TEMPLATE/release-checklist.md
@@ -95,15 +95,7 @@ Push access to the upstream repository is required in order to publish the new t
   - [ ] `git branch -d pre-release-${RELEASE_VER} release-${RELEASE_VER}`
 
 - Fedora packaging:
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

dustymabe:
The `ext.config.rpm-ostree.kernel-replace` test was failing because it was running out of space in the FCOS VM.

I adjusted the CI here to do a build using the RELEASE profile rather than DEBUG so the binary we build will be smaller.

dustymabe:
I moved the CI fix part of this to https://github.com/coreos/zincati/pull/1306 so please review that.

dustymabe (approved):
LGTM

.github/dependabot.yml
```diff
@@ -16,3 +16,11 @@ updates:
     open-pull-requests-limit: 10
     labels:
       - area/dependencies
+
```
dustymabe, line 19: I'm guessing this blank space isn't OK :(

PR #1309, opened 2025-07-17
bump zbus to 5.9.0

dustymabe:
This is a major update of zbus from 3 to 5. This also includes some updates to make it compile and a reworking of the connection creation to use a new (more modern) interface. See individual commit messages.


HuijingHei:
Thank you Dusty for fixing this.

jlebon:
Nice work Dusty!

